PR #1, opened 2014-09-24
Anti .DS_Store party

xbony2:
This is so mac users like me can send pull requests without the .DS_Store file that are auto-generatored by OS-X, since most people don't like those. See: http://en.wikipedia.org/wiki/.DS_Store


Blue64:
Oops, didn't mean to reference, sorry.

PR #2, opened 2014-09-25
Changed Inscriber to be upgradable with Cards

yueh:
Supports up to 3 cards
Currently max 3 Acceleration Cards, each reducing the crafting time by 25%


Blue64:
Oops, didn't mean to reference, sorry.

Issue #415, opened 2014-11-10
Suggestion: More info on invalid patterns


seebs:
Sometimes if mods have been updated, encoded patterns become invalid. It would be neat to have some way to examine an encoded pattern to see what it might have been, such as listing its inputs or outputs if any of them are still recognizable. Even something arcane like "Crafts One INVALID:4872 from 3 INVALID:2817" would be helpful, but if a couple of the parts are recognizable, you can probably figure it out.

leagris:
Some mod updates can have changed recipes, rendering existing patterns invalid.
It would be just fine it the pattern editor allowed placing the pattern back into the editor and even highlight wrong ingredients with some overlay texture.

PR #576, opened 2014-12-10
Add some detail for the 'Invalid Pattern' messages (issue #415)

riking:
Partially addresses issue #415.

Not the best solution - you can't see the valid data in the pattern, just a second line saying "BAD RECIPE". But I don't think a "valid=false" flag is very good, either. (Which the pattern terminal would attempt to ignore, and clear on re-encoding.) The problem with that is signaling the error condition to all the other consumers of `ItemEncodedPattern#getPatternForItem()` (such as the crafting CPU) which shouldn't have to deal with bad patterns.

See https://github.com/AppliedEnergistics/AppliedEnergistics-2-Localization/pull/84


Cisien:
@riking have you had a chance to work on the feedback that @thatsIch  provided?

src/main/java/appeng/helpers/InvalidPatternException.java
```diff
@@ -0,0 +1,94 @@
...
+		return getLocalizedMessage();
+	}
+
+	@Override
+	public String getLocalizedMessage()
+	{
+		String tmp = localization( reason ).getLocal();
+		if ( reason == REASON_BAD_ITEM )
+		{
+			tmp = tmp + ": " + nbtString;
```
riking: Not if the reason is REASON_BAD_ITEM; no race conditions here

halvors:
Maybe make the error messages lowercase? Not just uppercas letters?

Issue #66, opened 2014-09-27
[WAILA] explain states of MultiBlocks


AlgorithmX2:
If you look at a MB (for example a CPU) it tells you why it can not form or tells you whats wrong. Might need to pay attention information are not overloading the screen.

Originally Posted by thatsIch

This issue was migrated; For original details see http://ae-mod.info/Tracker/0726/

PR #621, opened 2014-12-22
Fixes #304 WAILA Integration updated and fixed

### LINKED ISSUE #304: AE2 channel usage and WAILA tooltip
Hello there,

For some reason, WAILA is displaying channel usage while in a singleplayer test world, but when I am on my private server it somehow doesn't display channel usage.

(Goes for Smart cables and Dense cables). Other information like is a device connected or not is shown in the WAILA tooltip.

This has been happening with RV1 stable 1 as well as with RV2 alpha 7.

Is this a bug? A feature? Or am I missing some configuration settings on client/server side?

thatsIch:
WAILA offers a new interface to sync the server data to the client.
Requires WAILA 1.5.7
Is back compatible, but will not provide all information

It is probably better to look at the resulting files. The original file was a giant `if instanceof` chain

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -18,7 +18,7 @@ fmp_version=1.1.0.311
 code_chicken_lib_version=1.1.1.105
 code_chicken_core_version=1.0.4.29
 nei_version=1.0.3.66
-waila_version=1.5.5_1.7.10
+waila_version=1.5.7a_1.7.10
 
 #########################################################
 # API Stubs                                             #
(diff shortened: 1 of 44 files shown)


thatsIch:
Aaand rebased again

yueh:
Looks good.

Some minor downside is that blocks accepting external power have two redundant lines for their power.
The one we add and which respects the player's setting and the general RF one.

For future changes this could be considered, if not added to the PR:

* #66 
* More support for full blocks regarding "Device Online/Offline", like Interface, Drive, Chest, QNB
* QNB linked state
* State of Toggle Buses
* State and strength of level emitters
* Current output of a Vibration Chamber
* [1.5.8] Add some icons through the custom renderer?
  * Cells inside a Drive/Chest
  * Interface configuration/patterns

TheJulianJES:
Finally!

Issue #118, opened 2014-09-27
Log to server log when a security terminal is broken.


AlgorithmX2:
Hello there

As it says in the description, it would be very helpful for server admins if, whenever a security terminal is broken, it logs the player name and security terminal location to the server log.

Regards,
ShamblerDK, owner of Awesomesauce 1.7.10

Originally Posted by ShamblerDK

This issue was migrated; For original details see http://ae-mod.info/Tracker/1099/

Technici4n:
Modern server tools allow monitoring when specific blocks are broken, and by whom. Hence this is unnecessary.

PR #627, opened 2014-12-24
Log to server log when security terminals are broken

riking:
Adds #118

### DIFF
--- a/src/main/java/appeng/block/misc/BlockSecurity.java
+++ b/src/main/java/appeng/block/misc/BlockSecurity.java
@@ -124,7 +124,7 @@ public class BlockSecurity extends AEBaseBlock
 			}
 			else
 			{
-				Optional<UUID> maybeUUID = WorldSettings.getInstance().getUUIDFromNumericalId( playerID );
+				Optional<UUID> maybeUUID = WorldSettings.getInstance().getUUIDFromPlayerID( playerID );
 
 				if ( maybeUUID.isPresent() )
 				{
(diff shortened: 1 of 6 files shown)


src/main/java/appeng/tile/misc/TileSecurity.java
```diff
@@ -275,6 +278,41 @@ public int getOwner()
 		return gridProxy.getNode().getPlayerID();
 	}
 
+	/**
+	 * Get a log-readable string representing the owner of this security terminal.
+	 *
+	 * @return string representing the owner
+	 */
+	public String getOwnerName()
```
riking: The main reason against putting it in Platform that I had was that this method isn't exactly useful to other code that isn't going directly into display to a human. Code wouldn't want to check if the string was "(no owner)". Perhaps a small data class is more appropriate?

src/main/java/appeng/block/misc/BlockSecurity.java
```diff
@@ -66,4 +72,38 @@ public boolean onActivated(World w, int x, int y, int z, EntityPlayer p, int sid
...
+		@SubscribeEvent
+		public void listenBreakBlock(BlockEvent.BreakEvent event)
+		{
+			// If logging is disabled, the listener isn't registered
+
+			// if ( ! AEConfig.instance.isFeatureEnabled( AEFeature.LogSecurityTerminalBreak ) )
+			//	return;
+
+			// Only listen to events for breaking BlockSecurity
+			if ( ! ( event.block instanceof BlockSecurity ) )
```
riking: `instanceof` doesn't get NPEs, and the second use is a cast in order to assign it to the variable. So I think it's fine here.

Issue #683, opened 2015-01-04
Issuing A Crafting Job From A Non-AE Terminal


Nividica:
Hello AE2 Team,
I have recently been working on integrating the auto-crafting system with Thaumic Energistics, and have run into an issue dealing with issuing crafting jobs from my terminal.
When attempting to start a crafting job from my terminal, I am running into a crash originating from the GuiCraftConfirm button list. When the constructor is called, the field 'OriginalGui' is not set, [Lines 90 to 100](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/63e375f2087c8c1f1f1aae94c17163b1bb9deb19/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java#L90). Because this field is not being set, the check on [line 122](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/63e375f2087c8c1f1f1aae94c17163b1bb9deb19/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java#L122) fails and the cancel button gets added to the button list as null. This in-turn causes a NullPointerException to occur from "net.minecraft.client.gui.GuiScreen.drawScreen" when it gets the button.

My original thought was to remove the check on line 122 so that the cancel button is never null, then put the check on line 169. If the OriginalGui is null in that location then just close the gui, if not continue as usual. However after testing with these changes I ran into another issue where after pressing Start, the gui does not close but instead only disables the start button (see attached image), but the crafting work does get performed.

Another idea I had was adding a method to the ITerminalHost interface to retrieve the original gui, such as "getHostGUI()". However this may have further reaching repercussions than I am aware of.

Any help with the issue or suggestions would be greatly welcome, as I have hit a roadblock.

![craftwindow](https://cloud.githubusercontent.com/assets/1585989/5604813/3d05639c-93a3-11e4-9322-f8a23be89c58.jpg)

thatsIch:
Looking at the code, the `this.OriginalGui` must be one of the following. So I think we can assume to set it `private final` and throw an `Exception` there, since it should not happen. That way, you can remove the null check.

Nividica:
Would that allow me to call on the GUI from my terminal without leading to a crash?
[...]

PR #688, opened 2015-01-04
Issuing A Crafting Job From A Non-Native Terminal

Nividica:
Related to issue #683

Ensures the cancel button is never null.
Adds code to close GUI if OriginalGui is null.

### DIFF
--- a/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java
+++ b/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java
@@ -119,8 +119,8 @@ public class GuiCraftConfirm extends AEBaseGui
 		this.selectCPU.enabled = false;
 		this.buttonList.add( this.selectCPU );
 
-		if ( this.OriginalGui != null )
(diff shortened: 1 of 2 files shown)


Nividica:
I simplified the changes as much as possible, as best I can tell the changes will only take effect when the GUI is called from a non-native AE part.

Cisien:
Looks good, though I have a few questions.

1. what I the use case for not passing the parent GUI into this method? AE uses this to return the user to the terminal GUI that the request was created with

yueh:
It is mostly a quick'n'dirty hack to not crash any addon terminal which is using the crafting GUI.
But to really fix it, we would have to refactor how we handle GUIs and provide a way to map tileentities back to GUIs dynamically and not static like now.

Compared to going back a step it will just close every gui instead of crashing.

Nividica:
Indeed. The issue I am facing is that ITerminalHost that is passed into the constructor does not matching the existing checks. That is to say I am not launching the Crafting GUI from an AE GUI/Terminal, but from my own GUI/Terminal. The proposed changes prevent a crash that occurs because of that, and instead close the GUI on conformation or cancellation.

thatsIch:
So, what about this PR?

Nividica:
I would be very appreciative if this PR was accepted ^.^

thatsIch:
Well, squash first, I guess?

PR #717, opened 2015-01-10
Synchronize displayname between server and client

yueh:
Also added a missing super call to the P2P tunnel

### DIFF
--- a/src/main/java/appeng/parts/AEBasePart.java
+++ b/src/main/java/appeng/parts/AEBasePart.java
@@ -40,6 +40,7 @@ import net.minecraft.util.Vec3;
 import net.minecraft.world.World;
 import net.minecraftforge.common.util.ForgeDirection;
 
+import cpw.mods.fml.common.network.ByteBufUtils;
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
 
@@ -171,12 +172,20 @@ public class AEBasePart implements IPart, IGridProxyable, IActionHost, IUpgradea
 	@Override
 	public void writeToStream(ByteBuf data) throws IOException
 	{
-
+		data.writeBoolean( this.hasCustomName() );
+		if ( this.hasCustomName() )
+		{
+			ByteBufUtils.writeUTF8String( data, this.is.getDisplayName() );
+		}
 	}
 
 	@Override
 	public boolean readFromStream(ByteBuf data) throws IOException
 	{
+		if ( data.readBoolean() )
+		{
+			this.is.setStackDisplayName( ByteBufUtils.readUTF8String( data ) );
+		}
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


AlgorithmX2:
Isn't DisplayName only synchronized on demand? ( in the gui ) this adds a lot of potential weight to general world data. Whats the intention here?

yueh:
Not always, parts like P2P tunnels are not synchronized between client and server. On a rejoin or something changes on the server it does not synchronize it.

Waila for example cannot display them. Of course there is a provider for server side data now in the latest one, but if this decreases the traffic? It would also be Waila only.

If we would want to show through something (maybe like the name in vanilla itemframes) we would need to account for server side changes. Memory cards for example overwrite the name, but no notifications of other players are doing it. Like continuously polling for changes (which is just bad).
The best option would probably be a more selective way to synchronize only changed tags, but that might be way out of the scope for AE2 currently.

If you have any objections against it, I will pull the PR.
But I still think the missing super call is not intentional, so this might be kept.

AlgorithmX2:
Well Waila should be able to pull the data from the server when it needs it, there was a bug several months ago that prevented that feature in Waila from working, I don't know what the current status of that is.

I'm pretty sure that any uses of the items display name should be synchronized at time of need rather then during chunk data sync.

The guis for instance all do this on demand when they do their initial data send. If theres an issue with walia not seeing the correct names, I'm sure that the walia plugin would be the place to solve that, rather then always on sending of the names?

I can't comment on the missing super, its possible I didn't care about the data that it sends for that part, but my memory is not a reliable source of information on that at this point.

yueh:
@thatsIch rewrote the whole Waila integration (there is an open PR for it)
They mostly added a (new?) API to fetch data directly from the server.

Maybe Waila is the better place currently, but it only solves the Waila issue. Not the one that parts are not synchronized between server and client in general (these without a GUI).

Querying the data if needed is perfectly fine, just the question about synchronizing a SMP environment as are no notifications about it, when another player changes something (besides polling everytime it calls readStream. meh.)
I would wish something with a low priority, like accumulate changes on the server and send a diff at most once per second or so.

The super call is for the NBT serialization of P2P tunnels, so they do not call `(writeTo|readFrom)NBT` on their proxy object and I have not seen any other class not calling super.

AlgorithmX2:
What other use does the DisplayName serve? other then Gui and Walia? You can only see it if you open the gui, or use Waila to look at it.

Yeah, theres no notifications about things like changing a name of an item in a tile because nothing can really see it, unless its specifically monitoring said value, ( Such as a gui. )

There is probably better ways this sort of thing could be handled, but this sort of thing is really a hidden value, with the exception of walia, which renders it upon you looking at it.

yueh:
For example some QoL changes. Like displaying the name of a P2P tunnel if the player holds a network tool. Similar to vanilla item frames, turtles or mystcraft books. This will make them a visible value and one which would need to be synchronized with the client on every change. It could be limited to a certain range of chunks around the player, but then we would need to query the server for changes each time the player visits a new chunk. In addition to the normal notifications.

Using something like the memory card just to fetch the name and read it in the tooltip would jst be annoying.
Color coding is not an option, as we would need 16^16 combinations with vanilla colors. There are these 4 conveniently placed spots each with 4 pixel on the back of a P2P tunnel, I am curious if it was intended to use them for something like that. But requiring to memorize 16 different colors on an object which can rotate into 6 different positions, no clear "front", maybe even require F3 to detect which side is north to compare the colors...

I opened the PR mostly to discuss the potential issues. :grinning:

PR #733, opened 2015-01-15
Disable level emitter on offline networks

yueh:
Turns off level emitters if the goes offline because it runs out of power, channels or similar.

This might break some builds, but it is not logical to let emitters output a redstone signal if they are offline.
It could consider the setting to invert the signal, but I do not think any device should be able to read its setting while offline.

### DIFF
--- a/src/main/java/appeng/parts/automation/PartLevelEmitter.java
+++ b/src/main/java/appeng/parts/automation/PartLevelEmitter.java
@@ -18,6 +18,7 @@
 
 package appeng.parts.automation;
 
+
 import java.util.Collection;
 import java.util.Random;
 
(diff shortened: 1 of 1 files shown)


AlgorithmX2:
Wasn't it coded to persist the previous value until power/channels were restored?

That is what I intended, since a redstone torch requires no power, it made sense that the level emitter would simply "stop updating" without network power/connection.

AlgorithmX2:
In retrospect, this change probably makes more sense, it probably also is more what people will think will happen, so it simplifies troubleshooting.

While I think that my original approach was "clever" I'll concede that this one makes more sense in the long run.

Issue #810, opened 2015-01-27
Inconsistant names: Stairs & Skystone


Deshiba:
Noticed something odd while working on some en_US.lang inconsistencies. All names follow the following pattern ->  ...[Function?].[Modname].[TypeMaterialForm].[Name]... <- for example:
- tile.appliedenergistics2.BlockQuartzChiseled.name
- item.appliedenergistics2.ToolNetherQuartzAxe.name
- etc.

Stairs and Skystone follow a different pattern:
- ChiseledQuartzStairBlock
- FluixStairBlock
- QuartzPillarStairBlock
- QuartzStairBlock
- SkyStoneBlockStairBlock
- SkyStoneBrickStairBlock
- SkyStoneSmallBrickStairBlock
- SkyStoneStairBlock
- BlockSkyStone.Block
- BlockSkyStone.Brick
- BlockSkyStone.SmallBrick

PR #813, opened 2015-01-28
Inconsistant code name fix

Deshiba:
Minor edit to bring the code names for *Stairs* in line with other blocks, as per #810


AlgorithmX2:
I'm pretty sure this will vaporize any of these blocks that are placed in the world.

Deshiba:
Didn't consider that. Probably not worth the rename then, seeing as it doesn't really change anything.

yueh:
Closing as it would break existing blocks, we might be able to reconsider it for 1.8.

PR #815, opened 2015-01-28
Add support for IConduitControl on quartz wrench

tterrag1098:
I assume this is how your @Interface works, as the BC methods don't have @Method on them. If that's not the case let me know and I'll fix it.


tterrag1098:
I've changed the maven dep to ignore transitive dependencies. This should fix EIO downloading a bunch of unneeded things.

Issue #817, opened 2015-01-29
Suggestion: Crafting color code to see what operation is currently done


RealStan:
The display of the current crafting process is fine. It is just sometimes confusing to quickly see what state the crafting process is in for a certain item (sceduled, stored or crafting).

I think of something like this:
- and I am sure you guys get this better visualized than me ;) -

![ae2_old](https://cloud.githubusercontent.com/assets/7492335/5959898/03b732a8-a7d2-11e4-9dc1-ab0e0e993601.jpg)
![ae2_new](https://cloud.githubusercontent.com/assets/7492335/5959897/03b55992-a7d2-11e4-856d-5be7fb13e15e.jpg)

PR #824, opened 2015-01-31
Added feature request #817

Eearslya:
Crafting CPU GUI now shows a yellow background on items being crafted, and a yellow background on scheduled items.


yueh:
We should probably consider adding a config option for it.
It flickers quite fast with fully upgraded assemblers and might be a bit irritating for some.

TheJulianJES:
@yueh I would like a config option ;)

Bestadd:
@yueh the company brig to turn it off is great but could there also be an option to change to color for those that are color blind?

yueh:
I already considered making it configurable as a color blind option. But I do not know of a green - yellow one, so we mostly need to target a total color blindness. (Also one reason, why I suggest these two colors)

Which then needs to deal with a lot more than just the two colors, we would also have to make the font color an option as there might be not enough contrast between the (very) dark gray font color and a potentially darker gray background color.

So it would be really nice to hear, if it is an actual issue for these persons.

PR #829, opened 2015-02-02
Improved OreHelper performance

yueh:
This might potentially break stuff. Please review thoroughly.

### DIFF
--- a/src/main/java/appeng/util/item/OreHelper.java
+++ b/src/main/java/appeng/util/item/OreHelper.java
@@ -18,8 +18,18 @@
 
 package appeng.util.item;
 
+
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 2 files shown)


KamranMackey:
Looks fine to me @yueh. Although to be honest, anything that improves performance is fine to me, but the code looks fine, at least in my opinion. I think it would be fine if you merged the PR.

thatsIch:
looked through,

any reason why in the `ItemRef` class, the fields are below the CTOR?


else code looks fine, its basically just a higher level cache put in between to reduce the look-up time for paying more space.

yueh:
Mostly personal habit to put (private) inner classes at the end, similar to private methods in general.

It is not necessarily a higher cache, but also reduces the memory usage immensely.
The old version creates a distinct list for each oredict entry containing every item as a hard copy.
For example all EnderStorage EnderChests are oredict with themselves. Meaning it has 4k oredict entries. Resulting in 16777216 unique itemstacks, which did consume about 1 gb of memory as I tested it.

PR #841, opened 2015-02-03
Controller config additions + Fixes for build scripts

Sureiya:
I've fixed the build script. It was copying jars into the run/mods folder that was causing an error upon startup (thanks @yueh).

The main reason for the PR is to add a max controller size configuration option.

### DIFF
--- a/gradle/scripts/autoinstallruntime.gradle
+++ b/gradle/scripts/autoinstallruntime.gradle
@@ -1,27 +1,3 @@
-// searches for NEI and Chicken stuff from compile set
-// and adds them to the mods in run dir
-task copyChicken(type: Copy, dependsOn: "extractUserDev") {
-    from { configurations.compile }
-    include "**/*Chicken*.jar", "**/*NotEnoughItems*.jar"
-    exclude "**/CodeChickenLib*" // because CCC downloads it anyways.. -_-
-    into file(minecraft.runDir + "/mods")
-    mustRunAfter "deobfBinJar"
-    mustRunAfter "repackMinecraft"
-}
-tasks.setupDevWorkspace.dependsOn copyChicken
-tasks.setupDecompWorkspace.dependsOn copyChicken
-
-// same for BC hacked file
-task copyBuildcraft(type: Copy, dependsOn: "extractUserDev") {
-    from { configurations.compile }
-    include "**/*buildcraft*.jar"
-    into file(minecraft.runDir + "/mods")
-    mustRunAfter "deobfBinJar"
-    mustRunAfter "repackMinecraft"
-}
-tasks.setupDevWorkspace.dependsOn copyBuildcraft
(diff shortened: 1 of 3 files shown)


AlgorithmX2:
Should there be a limit on the max size config?

The obvious reason why there is a limit, is to prevent people from making cable out of controllers, the less obvious reason is because testing controller "correctness" for larger structures will probably be slow; I don't think it would effect channel calculations much.

Sureiya:
What would be a fair max size for controllers? I was making an ME "server room" on my server, and kind of wanted the frame of the building to be the controller, hence the config options. 16 is quite a bit, but I can imagine someone using a big more if they were trying to keep their network very organized.

Cisien:
If the concern is about performance, the stress test it in low end hardware to see where it breaks, if the concern is gameplay, then cap it at 7x7. The config can still be used to increase the difficulty.

AlgorithmX2:
Gameplay can be modified by configs to a reasonable extent, 7x7x7 was my version of "that's plenty big" on the other hand I don't think we should support MAX_INT controllers that stretch for miles. config wise, 16 might be a good medium; it a little bit bigger for people who want that, but its not so bit that you can use it as a cable in larger bases.

Seeing where it breaks would be interesting if anyone wants to give it a try.

I just want to prevent sillyness in both cases really.

yueh:
7x7x7 is way more anyone needs. It should provide some 10k channels.

There was a build on reddit using a hilbert curve and it provides 16320 channels. And is still inefficent as there are multiple spots, which can trade 2 controller sides for 4.

If configurable, I would say a range of 1 to 7. This would enable modpacker makers to put some additional limits on the users.

AlgorithmX2:
I don't really disagree about the 7x7x7 being more then enough, @Sureiya  did make this PR specifically to get a larger controller however.

I'm 110% ok with smaller controllers, cause limits can be challenging, I'm just trying to keep an open mind about the more relaxed folks.

shobu9:
@AlgorithmX2 I think it's just because they want to use them as decoration for a base!

Chisel why don't you use normal ae-2 cntrollers textures !

Sureiya:
I didn't actually want to use them as decorations, I just wanted room to organize my drives crafting CPUs and P2P tunnels nicely. 7x7x7 made that hard to do in the scale I wanted. It was just a constraint I didn't really care for.

thatsIch:
I also have my doubts, giving the user access to all 3 coordinates. Should be just a cube

yueh:
I did not even consider the potentially performance issue in regards of the increase network sizes. There are already enough issues as people do not chunkload their network and the larger a network is, the more time it takes to boot up. Or other issues, which showed up lately.

So 7x7x7 should be really the hard cap, but it should be possible to limit them to smaller sizes (down to a single block)

PR #843, opened 2015-02-03
Refactored AEItemDefinition to use Optional instead of null

yueh:
The work is mostly done by @thatsIch, my contribution is just the small changes to migrate AE2 itself to the new api and squash his comments to be a bit more compact.

This will change the API greatly. This should be normally avoided in a non alpha state, but it is necessary to solve many issues with disabled feature in a sane way as well being able to push a stable build finally.
There is still a backward compatibility layer, so rv2 addons will not break.
For addon devs it is possibly the best option to use the deprecated way for now and plan a switch with rv3.

I personally would mark it for removal with the end of rv3 alpha/start of rv3 beta.
But I could see it stick around for rv3 completely.

### DIFF
--- a/src/api/java/appeng/api/AEApi.java
+++ b/src/api/java/appeng/api/AEApi.java
@@ -23,6 +23,7 @@
 
 package appeng.api;
 
+
 /**
  *
  * Entry point for api.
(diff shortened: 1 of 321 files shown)


yueh:
Updated version.

I needed to split the API construction into two phases. Some parts like the interface or formation plane require the helpers of the public API to be available. So it is impossible to construct the API successfully without constructing itself successfully before.

PR #853, opened 2015-02-05
Removed copy in favour of threadsafe collection

yueh:
**Test - do not merge**

### DIFF
--- a/src/main/java/appeng/crafting/CraftingJob.java
+++ b/src/main/java/appeng/crafting/CraftingJob.java
@@ -164,7 +164,7 @@ public class CraftingJob implements Runnable, ICraftingJob
 
 				Stopwatch timer = Stopwatch.createStarted();
 
-				MECraftingInventory craftingInventory = new MECraftingInventory( this.original, true, false, true );
+				MECraftingInventory craftingInventory = new MECraftingInventory( this.original, true, true, true );
 				craftingInventory.ignore( this.output );
 
 				this.availableCheck = new MECraftingInventory( this.original, false, false, false );
@@ -188,7 +188,7 @@ public class CraftingJob implements Runnable, ICraftingJob
 				try
 				{
 					Stopwatch timer = Stopwatch.createStarted();
-					MECraftingInventory craftingInventory = new MECraftingInventory( this.original, true, false, true );
+					MECraftingInventory craftingInventory = new MECraftingInventory( this.original, true, true, true );
 					craftingInventory.ignore( this.output );
 
 					this.availableCheck = new MECraftingInventory( this.original, false, false, false );
--- a/src/main/java/appeng/crafting/MECraftingInventory.java
+++ b/src/main/java/appeng/crafting/MECraftingInventory.java
@@ -18,6 +18,7 @@
 
 package appeng.crafting;
 
+
 import appeng.api.AEApi;
 import appeng.api.config.Actionable;
 import appeng.api.networking.security.BaseActionSource;
@@ -26,6 +27,7 @@ import appeng.api.storage.IMEMonitor;
 import appeng.api.storage.StorageChannel;
 import appeng.api.storage.data.IAEItemStack;
 import appeng.api.storage.data.IItemList;
+
 
 public class MECraftingInventory implements IMEInventory<IAEItemStack>
 {
@@ -57,7 +59,7 @@ public class MECraftingInventory implements IMEInventory<IAEItemStack>
 		this.par = null;
 	}
 
-	public MECraftingInventory(MECraftingInventory parent)
+	public MECraftingInventory( MECraftingInventory parent )
 	{
 		this.target = parent;
 		this.logExtracted = parent.logExtracted;
@@ -84,7 +86,7 @@ public class MECraftingInventory implements IMEInventory<IAEItemStack>
 		this.par = parent;
 	}
 
-	public MECraftingInventory(IMEMonitor<IAEItemStack> target, BaseActionSource src, boolean logExtracted, boolean logInjections, boolean logMissing)
+	public MECraftingInventory( IMEMonitor<IAEItemStack> target, BaseActionSource src, boolean logExtracted, boolean logInjections, boolean logMissing )
 	{
 		this.target = target;
 		this.logExtracted = logExtracted;
@@ -106,14 +108,16 @@ public class MECraftingInventory implements IMEInventory<IAEItemStack>
 		else
 			this.injectedCache = null;
 
-		this.localCache = AEApi.instance().storage().createItemList();
-		for (IAEItemStack is : target.getStorageList())
-			this.localCache.add( target.extractItems( is, Actionable.SIMULATE, src ) );
+		this.localCache = target.getStorageList();
+		// AEApi.instance().storage().createItemList();
+		// for (IAEItemStack is : target.getStorageList())
+		// this.localCache.add( target.extractItems( is, Actionable.SIMULATE,
+		// src ) );
 
 		this.par = null;
 	}
 
-	public MECraftingInventory(IMEInventory<IAEItemStack> target, boolean logExtracted, boolean logInjections, boolean logMissing)
+	public MECraftingInventory( IMEInventory<IAEItemStack> target, boolean logExtracted, boolean logInjections, boolean logMissing )
 	{
 		this.target = target;
 		this.logExtracted = logExtracted;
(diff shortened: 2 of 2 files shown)


AlgorithmX2:
I'm not really sure what your testing, but this is clearly broken, as the localCache is used to simulate the network after crafting steps are completed, using the network's copy to do that, will result in crafting calculations affecting the real networked storage.

As for real solutions, I originally intended to layer a "changes log" over the real network storage to prevent the copy, I don't recall why I didn't do that, probably because I just wanted to be done as most issues from rv1.

yueh:
To clarify it a bit more.

Before handing the crafting job to the thread, it does a (hard?) copy of the whole inventory including a copy of every item. This is done in the server thread.
The crafting thread itself then does create a copy of the cache (via `getAvailableItems()`) with references to the earlier created copy. 

There is already a "changes log" in place and used for extracted and missing items. But not inject, which operates on the localCache.
The question now is, why is it disabled? The code for handling seems to be existent and some simple tests with vanilla crafting recipes did not show any issue. No duped/missing items, the intermediate crafting results did not end up in the network, no changes to the network while calculating (second viewing the terminal).

As a major part of the cpu time is spend in `ItemList.findPrecise()` and `CSLM.get()`.
A `NavigableMap` is only needed for `ItemList.findFuzzy()`, so we could spend some memory for a `LoadingCache` from guava and cache the lookup, moving the complexity from `O(log n)` to `O(1)` (There is no real statement about the complexity of a `LoadingCache`, but anything other would make a cache pretty much useless). The only occurence of a removal is the `MeaningfulIterator`, so invalidating a cache entry would also be straightforward.

yueh:
Pushed an update, which also enables the changelog for injections instead of directly inject them into the original inventory/cache

AlgorithmX2:
Technically the reason the performance hog is findPrecise is..

for (IAEItemStack is : target.getStorageList())
     this.localCache.add( target.extractItems( is, Actionable.SIMULATE, src ) );

which pretends to extract every item from the network at its full quantity ( thus verifying the integrity of the storage list ) is it still this when since you have removed this integrity validation?

yueh:
And the actual reason for `findPrecise` being slow is `O(log n)` for `CSLM.get()`.
I am also testing a `LoadingCache`, which would trade some memory for a better complexity. But it requires some carefully placed invalidations.

Regarding the integrity validation, how can this be lost?
One of the first things which are done with the full copy, is create a copy of the list (referencing the same items). This will use the `MeaningfulIterator` and at least perform the self cleaning(/healing?).
Or is there something else I need to consider?

AlgorithmX2:
The Storage list ( inv.getStorageList() ) is not necessarily 100% accurate, its a best guess as provided by the inventories, it fails to be accurate for several inventories, such as delay updated storage buses.

Where as the extract method corrects any inaccuracies ( as best it can; since 100% accuracy cannot be provided via simulation ).

I'm not suggesting that trying another storage is bad, just merely pointing out that the majority of that issue is caused by the fully extraction, which could be avoided if we moved to a layered approach where we only test things and cache the results locally on demand, thus improving performance by limiting scope of the problem, rather then trying to optimize the code, which is clearly doing far more more heavy lifting then is necessary.

yueh:
As written earlier, there are already changelog/diffs/deltas for inject,extract and missing. But the injection one is not used. And the question is why? Was there an issue with CME as the `ItemList` did use a `TreeMap` instead of a `CSLM` earlier?

The `LoadingCache` by itself is not bad and it works pretty well for the OreHelper, as there is only one instance. But it would be needed in the `ItemList` and there are way to many small ones compared to only a few large. So we had to introduce something like a `CachedItemList` for a few occurences. As the overhead for small ones would be too much and probably way to much work to do use it correctly in the right places.

AlgorithmX2:
I can't really explain why things are the way they are in the Crafting Inventory, its been at least 6-7 months now since I messed with it, and I wanted nothing to do with crafting while I was working on it.

I doubt there was an issue with CME's with it, to this day i'm not really sure why there was CME's anyway ( since the crafting process didn't run concurrently - yes it was in its own thread, but it was executed for a fixed time period during the tick handler ) unless there was a bug with that logic, either case I quit before looking into it.

yueh:
I am suspecting [this](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/eb74fb33b823d0b3730600b862315acf1eeff5f2/src/main/java/appeng/me/storage/MEMonitorIInventory.java#L179) (Maybe also another inventory, anything which somehow modifies the inventory in the inner loop. Not related to threads. Modifying it while iterating already counts a concurrent)
```
for (IAEItemStack is : target.getStorageList())
   this.localCache.add( target.extractItems( is, Actionable.SIMULATE, src ) );
```
Iterates over the storagelist of an MEMonitorIInventory and each extract calls `onTick`, which also iterates over the same collection, causing the fast-fail iterator of a `TreeMap` to trigger the exception, if the (inner) MeaningfulIterator does modify the list.

But back to the the crafting. I have changed it to use the `logInections` and it works with fine with with normal crafting. I just fear these strange edge cases.
The extract could be also moved into the thread itself. reducing some pressure on the server thread, if it is really needed.

AlgorithmX2:
I suppose it all comes down to whats effective, I'd suggest creating a network full a huge piles of items ( from the item generator ) to do some comparison profiles.

and see if the changes made so far are making a big enough difference. I imagine that moving the copy to the thread would only cause issues rather then help.

yueh:
I have access to a testworld with that issue.

Sampling over about 105k ms using the rv2-beta-b8 consumes about 50k ms with copying the itemlist (server thread). A build with the injection change log + passing the storage list will not even list creating the CraftingJobs and the Crafting Threads themselves are not noticable slower.

Based on the build it probably needs a large network without a separate storage network.
Nothing I normally build...

yueh:
The whole issue is a bit tricky. The crafting calculation just makes it obvious as it uses extract on the whole inventory.

The actual issue is the ItemList using a NavigableMap with `O(log n)` for get, put, etc. and that just adds up. Creating a copy of the whole network just makes it obvious.
There is also no real difference between the `TreeMap` and `ConcurrentSkipListMap`

AlgorithmX2:
I Still think its an issue of the scope of the search, sure if you had an export bus for every item on your network, and you had 30,000 items in your network you would expect things to slow down.

The craft calculation is expected to be slow, thats why I have it coded as as threads that run for a segment of time each tick, rather then doing it inline with the normal processing.

If we could remove the overbearing stress of extracting the whole network, the thread can fun for its minor window of time each tick to do its work in peace.

When I tested different storage mediums for ItemList TreeMap ended up winning for performance of my test network; which was a fairly normal network, with a decent amount items you expect to see in the game stressed with lots of exports/imports and such.

If you have a data structure that can preform better, and also provide the ability to solve fuzzy searches with ranged results, then by all means.

I guess that about sums up my feedback on this issue however.

yueh:
Both `NavigableMap` perform about the same. The `TreeMap` might be a bit faster in some synthetic benchmarks, but with the downside of some random crashes. (But without the copy this would also not happen).

Switching collections/ripping the code apart is more about digging through it and understanding it. Like I have maybe spend 10-15 minutes to replace the `NavigableMap` complete with a normal HashMap and using the guava `Collections2` to create the submap. It was a minor performance increase, but it shifts the complexity from findPrecise to findFuzzy. So it probably just trades on issue for another.

Eliminating the copy is the way to go. There is no real alternative to it.
As the issues really show up in cases like copy the whole network or use the item generator with 6 fully upgraded import buses and dump that into a network. (That is the test I used for `TreeMap` vs `CSLM`)

FireBall1725:
Testing currently on ForgeCraft 2, fixes the lag issue.  So far no issues have shown up yet as far as unknown bugs.

yueh:
Potential issues are probably extremly hard to test/track down. I currently would assume the following situation could result in some errors:

1. Use an external inventory like a chest or barrel.
  1. Preferably something without a block update on inventory changes.
2. Attach it with a storage bus to the ME network.
3. It needs to share the content with something.
  1. Can be another ME network with another storage bus.
  2. Any other piping mod, which can import and export items.
4. In the exact same tick as the one issuing the crafting request:
  1. Completely/partly empty the inventory with the other network (below the requirement)
  2. A block update will not refresh the cache for at least 1 tick.
  3. Without a block update it can potentially take way longer and allowing a bit wider window for testing.

But I honestly have no idea how to reliably test this. At least not without writing real integration tests and mocking that exact behaviour. Or some obscure stuff like using the stacktrace to identify the caller and report a too high value for any `MECraftingInventory`. Potentially introducing false-positives, which actually would never occur under real conditions.

yueh:
Another issue (potentially) is `AEItemStack.compareTo()`, which normally will not show up via sampling (inlining etc). But sampling will show `CSLM.findPredecessor()` as major issue for `ItemList.add()` being slow. Which itselfs uses `compareTo`.

The profile I have done, counts about 250k invocations for `ItemList.add()` resulting in about 20 million invocations combined for `AEItemStack.compareTo()` (And another 8 million of `Object.hashCode()`. Of course it should be taken with a grain of salt in regards to profiling and the missing inlining etc.

About 3/4 of the additional calls are to compare two integers using `return l < m ? -1 : (l > m ? 1 : 0)` as non of them will overflow (itemid, metadata and displaydamage) it could be simplified with just `return l - m;`.

And 1/4 for comparing NBT data, even for items without NBT data. Which is using `System.identityHashCode()` and there are some regression issues reported for some JVM version.
Removing the call for NBT less items might be an idea.

It probably will not result in a major performance boost, but in regards of the amount of calls it probably will add up.

PR #854, opened 2015-02-06
Use factually superior File constructor

riking:


thatsIch:
`new File( world.getPath(), "AE2" );`

would be probably the correct choice. If you only provide the ´world` instance, it will auto converted by `toString()` and its probably something different than `getPath()`

yueh:
`File(File, File)` and `File(String, File)` (with `File.getPath()` does exactly the same, just that the first calls getPath internally (or better access the field directly).

But there are actually way more occurrences, which could be changed in general. Not just that single line.

riking:
Yeah, this came up in a EIO issue report and I made this half-facetiously.

yueh:
It is a bit more elegant as we do not need to deal with `File.separatorChar`. There are still other places using `File` with the String stitched together, so making it inconsistent.

But I do not see anything preventing the PR, if all occurrences are changed in a consistent way.

Sadly it will not fix the issue with the wrongly reported worldfolder..

PR #931, opened 2015-02-23
Changed biome temperature

FireBall1725:
Changed the biome temperature from cold to hot, to fix blizz spawns from TE.

### DIFF
--- a/src/main/java/appeng/spatial/BiomeGenStorage.java
+++ b/src/main/java/appeng/spatial/BiomeGenStorage.java
@@ -28,7 +28,7 @@ public class BiomeGenStorage extends BiomeGenBase
 		this.setBiomeName( "Storage Cell" );
 
 		this.setDisableRain();
-		this.temperature = -100;
+		this.temperature = 100;
 
 		this.theBiomeDecorator.treesPerChunk = 0;
 		this.theBiomeDecorator.flowersPerChunk = 0;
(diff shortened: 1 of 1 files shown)


yueh:
Any particular reason for it besides the spawning?
In regards to vanilla, there is no real benefit for a hot biome. Only the loss of snow farms, which can break builds.

I have no idea how TE decides to spawn blizz, but I do not think messing with the biome temperature to prevent them from spawning is a good solution.
I could imagine, that they directly inject it to the spawn lists, so overriding `getSpawnableList()` and just let it return null or an empty list is probably a better option to enforce no mob spawning.
If they do not respect this whitelist, then I would say it is not really an issue, we need to fix.

Both values are also very likely wrong, as the other biomes are using [0,1), maybe [0,1]? I did not find any documentation about the upper limit. So we should probably change it to 0.0.

FireBall1725:
@yueh changed the biome temp to 0.5F

AlgorithmX2:
I'd say this doesn't make much sense; Storage Cells are really cold ( there is no source of energy, so technically the only heat in the storage cell is introduced by the objects moved into them; there is no sun, or other heat sources ), and their data should reflect that; TE should probably check to see if the white list is empty before injecting mobs to spawn inside of biomes that doesn't have spawns.

TheJulianJES:
I will post this to the CoFH Tracker

yueh:
Closing for now.

1. Changing the temperature to prevent mob spawning is just a workaround. The next mod adding mobs to temperate biomes will just cause the same issue. Overriding `getSpawnableList()` is probably the better solution (If spawning still occurs, I would not assume it as our issue)
2. As the dimension has a roof/no sky, water will not freeze. (And even vanilla can prevent it)
3. We should add a IMC message to prevent MystCraft from using it (but that is another issue)
4. -100 as value is probably questionable, but there is no documentation saying only [0,1) is allowed. If it ever breaks vanilla minecraft or forge, 0 should be still ok to fit the theme.

Issue #932, opened 2015-02-23
Fire ExplosionEvent for Tiny TNT


MineMaarten:
Some people have reported to me that Tiny TNT can't be used anymore to get Compressed Iron from Iron in PneumaticCraft. Recently, as a refactor, I've instead of using custom EntityItems, hooked into forge's ExplosionEvent to spawn Compressed Iron. This works fine for normal explosions.

After investigation, it seems that Tiny TNT simulates Explosions in a custom way, and doesn't fire the ExplosionEvent. My suggestion would be to do this :). Or, a simpler solution is to use World#createExplosion instead. But, I assume there was a reason to not do that in the first place?

yueh:
´World.createExplosion()´ is not really an option. There is no guarantee that any explosion will happen as the explosion damage can be disabled in the configuration.

An `ExplosionEvent` is likely the way to go, but not available with the forge version we are currently using.

Cisien:
I seem to remember algo going this route because of mods or addond for bukkit that blocked explosions. This would allow players to still be able to craft singularities even when the explosions are disabled

AlgorithmX2:
Might need to either update forge, or reflect the event ; its likely the only solution without damaging the anti-explosion behavior.

yueh:
I am already planning on updating forge to the latest recommended version as well as stuff like NEI. (It is probably WAILA breaking NEI 1.0.3 for indev).
Just have not come around to do it.

Adding the event is pretty trivial after that.

JasonMcRay:
Sorry to bump this. but is something being done with this? It has been a year since the report and it still doesnt seem to be implemented.

thatsIch:
sorry, not of importance

PR #952, opened 2015-02-27
Updated Forge and other dependencies

yueh:
Updating forge to the newest recommended build is necessary for #932. It is probably possible to just update to the version adding the ExplosionEvent, but I personally prefer to just stick with a recommended version. (if possible).
As this is a hard dependency, people will need to update their forge version. Not ideal as we are in a beta phase, but there are already other mods depending on 1291 (like TE/TD). So we are not the only one pushing it.

CCC and NEI are not recommended builds and the WAILA version is also not really necessary.
Both the combination of these three is breaking NEI and making it unusable in the dev environment.
WAILA 1.5.8/9 adds some potentially interesting features, if we decide to use them. But it requires NEI 1.0.4.+ and to really work again also a newer CCC version.
And while I am already bumping, I have also bumped CCL and FMP.
As these are soft dependencies and we are currently not using newer features, there should be no reason to expect it to break.

### DIFF
--- a/build.properties
+++ b/build.properties
@@ -8,17 +8,17 @@ archivesBaseName=appliedenergistics2
 # Versions                                              #
 #########################################################
 minecraft_version=1.7.10
-forge_version=10.13.2.1230
+forge_version=10.13.2.1291
 
 #########################################################
 # APIs used for development                             #
 #########################################################
 cb_minecraft_version=1.7.10
-fmp_version=1.1.0.311
-code_chicken_lib_version=1.1.1.105
-code_chicken_core_version=1.0.4.29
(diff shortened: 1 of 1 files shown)


Adaptivity:
People might blame you for Forge 1291 since Cauldron doesn't support it. And as I know, TE/TD requires Forge 1291 only on client.

yueh:
We need a newer forge version on the server. According to the build date something between 1254 and 1264, but for performance reasons a server should use at least the 127x (the one fixing the chunk gen issue).

Regarding Cauldron, we cannot consider a dead project for blocking everything. Especially as we never supported it.

Issue #574, opened 2014-12-10
Crash with Wireless Terminal


TheJulianJES:
Something like #374.
If you right-click on the wireless terminal, switch to another item it will crash.
I'm playing on a server, maybe in the time it was a bit laggy, maybe the gui trys to open for another item, not for the wireless terminal.
I currently have OptiFine on my laptop, on my gaming pc I tried it without, still not working (crashing).
I also have ExtraCells2 installed, also crashes without.
Also portable cell, maybe [tested].

Version:
1.7.10 - AE2 rv2 alpha 28

Steps to reproduce:
1. Setup a server, maybe a bit laggy.
2. Right-click the wireless terminal.
3. Switch to another item/empty hand.
4. The wireless terminal gui will open shortly (no items and such).
5. Crash.

Crash report: http://pastebin.com/jp426z0k

TheJulianJES:
#889 is maybe referenced.

thatsIch:
probably resolved, closing

PR #964, opened 2015-03-03
Fixes #889 possible dupe bug

### LINKED ISSUE #889: Portable cell dupe
Using the latest beta build, a portable cell can overwrite another portable cell if there is enough lag to switch items between activating the source and the GUI opening.

yueh:
Adds a hotfix to prevent duping with portable cells.
Should possibly fix the NPE crash with switching slots on laggy servers. #574 (Might not fix every possible combination)

It is mostly a hotfix to prevent the bug.
The new API interface is mostly tentative and subject to change.

There should be a general GUI refactoring possibly using a registry/factory to create the GUIs instead of using reflection. This can potentially also allow AE2 addons to better reuse our GUIs (like crafting)

### DIFF
--- a/src/main/java/appeng/client/gui/implementations/GuiMEPortableCell.java
+++ b/src/main/java/appeng/client/gui/implementations/GuiMEPortableCell.java
@@ -30,7 +30,7 @@ public class GuiMEPortableCell extends GuiMEMonitorable
 
 	public GuiMEPortableCell( InventoryPlayer inventoryPlayer, IPortableCell te )
 	{
-		super( inventoryPlayer, te, new ContainerMEPortableCell( inventoryPlayer, null ) );
+		super( inventoryPlayer, te, new ContainerMEPortableCell( inventoryPlayer, te ) );
 	}
 
 	int defaultGetMaxRows()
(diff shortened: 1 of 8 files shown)


thatsIch:
is this really needed in the API itself?

yueh:
It depends on AE2 Addons, if some of them are providing their own portable cells and are reusing our GUI, then yes as these can be also affected by the bug. Otherwise not really.
And I did not want to break the IPortableCell and letting it implement the new (possible temporary) interface.

A real (longterm) solution would be a complete rework of the GUI handling.
For example thaumic energistics is reusing our crafting confirmation, but the cancel button is using hardcoded instanceof checks to reopen the previous GUI, which of course fails for unknown TileEntities.
For example a registry to add a mapping between the TileEntity and a GUI/Container.
But I have not really enough time for it currently, so at least having a possible hotfix for the next release available is better than nothing.

Items like the NetworkTool make it also quite tricky to deal with them. As they need to either encode the TileEntity they are used on or the slot index.
The Forge way is not really intented to deal with items opening GUIs, just TileEntities placed in the world. So either we need to abuse it and encode it somehow with the coordinates or the GUI id (`ID_ORDINAL`)

yueh:
I will make another pass over it and move the interface into the internal code.
There is currently no real reason to be externally available as there is no real way for addons reusing GUI parts without refactoring it completely.

PR #970, opened 2015-03-04
Introduction to the new Optional API

thatsIch:
**Changes**

* `blocks()`
* `parts()`
* `items()`
* `materials()`

and thus use the new re-direct via `definitions()`.
Also fields are marked as `@Nullable` to give direct IDE support, if somebody should ever use them by accident again

**Intention**

For future reference, it would be better to initialize the block and item definitions,
but intercept the registration, so the user cant access it, but the processing,
will not fail, because the definitions will still be valid. This could be an API change introduced in Minecraft 1.8+

**Old Way**
    `AEItemDefinition AEApi.instance().blocksQuartzOre`

**New Way**
    `Optional<AEItemDefinition> AEApi.instancen().blocks().quartzOre()`

It uses the Guava Optional Class, which is feature-wise more than the Java 8 version. It enables transformations and iterations

**Notice**

**Edit**
Format

### DIFF
--- a/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java
+++ b/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java
@@ -128,7 +128,7 @@ public class GuiCraftConfirm extends AEBaseGui
 	private void updateCPUButtonText()
 	{
 		String btnTextText = GuiText.CraftingCPU.getLocal() + ": " + GuiText.Automatic.getLocal();
-		if ( this.ccc.selectedCpu >= 0 )// && ccc.selectedCpu < ccc.cpus.size() )
+		if ( this.ccc.selectedCpu >= 0 )// && status.selectedCpu < status.cpus.size() )
 		{
 			if ( this.ccc.myName.length() > 0 )
 			{
--- a/src/main/java/appeng/items/parts/ItemFacade.java
+++ b/src/main/java/appeng/items/parts/ItemFacade.java
@@ -41,8 +41,11 @@ import cpw.mods.fml.common.registry.GameRegistry.UniqueIdentifier;
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
 
+import com.sun.corba.se.impl.oa.NullServantImpl;
+
 import appeng.api.AEApi;
 import appeng.api.parts.IAlphaPassItem;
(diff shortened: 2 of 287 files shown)


yueh:
Nice to see you back in action.

The API should be a definitive change for 1.8 (or maybe even rv4, depending on how we move forward)

I have not test it yet (no build for now).
But are PartInterfaces working? As I tested your old version, they are not created.
This is caused by their final DualityInterface, which needs a fully constructed API to construct them sucessefully. But as they are now created during the API construction, this adds a cyclic dependency making it impossible.

src/main/java/appeng/items/parts/ItemFacade.java
```diff
@@ -41,8 +41,11 @@
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
 
+import com.sun.corba.se.impl.oa.NullServantImpl;
```
yueh: I do not think this should be here.

src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java
```diff
@@ -128,7 +128,7 @@ public void initGui()
 	private void updateCPUButtonText()
 	{
 		String btnTextText = GuiText.CraftingCPU.getLocal() + ": " + GuiText.Automatic.getLocal();
-		if ( this.ccc.selectedCpu >= 0 )// && ccc.selectedCpu < ccc.cpus.size() )
+		if ( this.ccc.selectedCpu >= 0 )// && status.selectedCpu < status.cpus.size() )
```
yueh: Did some refactoring went wrong here?

yannik2k:
can i anywhere see, how can i use the api?

thatsIch:
I would rather put it into 1.8 so we can start fresh. Fixing code and patching it up on other side is troublesome.

I will take a look at it

yueh:
Besides the missing part interfaces, it looks good. (and maybe a organize imports + remove trailing whitespace)

yannik2k:
:/ can anyone say me how can i create a recipe with ae items?
it seems to be strongly changed

thatsIch:
@yannik2k take a look into the .recipe files in the resource folder.

thatsIch:
I will open a different solution, which is fully functional. I do not think that the `Optional` is the way to go in the end.

PR #981, opened 2015-03-07
Introduction to the new definitions API

thatsIch:
**Changes**

* `blocks()`
* `parts()`
* `items()`
* `materials()`

and thus use the new re-direct via `definitions()`. All fields do not need to be `@Nullable` anymore, since they can not be `null` anymore.

All definitions are now initialized, no matter what. But SubItems, Items and Blocks are not registered, if by chance are disabled.

There is also a suggestion for the new Definition handling. It is not hooked up yet, but it should not change much, since internally its already using different definitions depending on item, block, etc, but improve access errors (can not be null calling upon a normal decorative block for a tile entity)

**Intention**

In comparison to the `Optional` way, it has a straight forward implementation. No random `RunTimeException` to catch some cheese case, which propagates into some random `null` and thus finally do just nothing for a code-part. 

**Old Way**
    `AEItemDefinition AEApi.instance().blocks().blocksQuartzOre`

**New Way**
    `AEItemDefinition AEApi.instance().definitions().blocks().quartzOre()`

**Test**
Set all config settings to false. It result into a lot of NPEs which were based on the recipes, but all in all no crashes and no other messages I have seen so far.

Probably needs major testing.

**Notice**


yueh:
Part interfaces are back, I sadly cannot remember the other items affected by it. But I highly doubt that this will be an issue. 
Any particular important changes, which need to be considered? (compared to the previous PR). Say for adding new items?

There are still some imports of `Optional` left and the javadoc for `IBlocks` referes to something optional. Maybe this should be more precise and explicitly state that anything can return null?
It should also cover the other definitions, not just be blocks.

I would say that recipes fail is not relevant as these are made with the default settings in mind.
If anyone decides to disable major parts, they can still make AE2 usable by themselves with our recipe system. It would be an issue, if everything would be hardcoded. Then we would have to provide a fallback solution for every possible permutation. But I honestly do not want to even think about the dependency graph needed. For example disabling the grindstone will make AE2 competely unusable (no cables, no buses/interface), if no other mod provides a machine to grind dust or provide a recipe for it.

Catching the exception and provide a meaningful log message like "recipe X disabled because of missing item Y" is probably a good idea.


Could you please (shortly) elaborate about `Optional` not being the final solution?
The Guava `Optional` is a bit messy, but that is mostly caused by Java6/7 requiring to be pretty verbose. The JDK8 version is way more clean thanks to the stream API. No need to use the foreach approach or anonymous classes. Just a `definition.map( x -> foobar(x))` (or `ifPresent(Consumer)` or the other methods)

thatsIch:
> Part interfaces are back, I sadly cannot remember the other items affected by it. But I highly doubt that this will be an issue. 

I think it was the annihilation plane

> Any particular important changes, which need to be considered? (compared to the previous PR). Say for adding new items?

* Stairs are now decorative items, since disabling decorative items, blocks like quartz pillars are disabled, its only meaningful to keep it transitiv
* All P2P tunnels are P2P features now `AEFeature.P2PTunnel` so if you disable that, all P2P will not register anymore
* There is no easy way anymore to see if a feature is deactivated based on its definition, thus I changed the creative tabs icon from the controller (deactivated through channels features) to the fluix crystal. Can be reworked back by for example adding a dummy block with the textures of the controller and use that, or re-use a class and inject the dependency there, to search for an enabled texture.

> There are still some imports of Optional left and the javadoc for IBlocks referes to something optional. Maybe this should be more precise and explicitly state that anything can return null?
It should also cover the other definitions, not just be blocks.

oh forgot about the javadocs, branched it from the other branch... its on my todo

> Catching the exception and provide a meaningful log message like "recipe X disabled because of missing item Y" is probably a good idea.

its already done, but it still prints the NPEs

> Could you please (shortly) elaborate about Optional not being the final solution?

Actually since the Guava one is more powerful, its better suited than the JDK 8 version, at least atm. The problem is, that our current code is not programmed for that case (easily show-cased how many `null` were initially missing). `Optional` will propagate through the whole code which looks ugly as hell and can lead to overhead in certain areas.

There are also logic traps like, why do I need to check for a block definition to be valid in its own class. Obviously a failing OOP design, but rewriting such things will take much longer.

The problem arises due to the initial programming style, to aggregate a lot of classes and cases to have less code for a problem, but more verbosity on the other side and many more cases to handle.

yueh:
> There is no easy way anymore to see if a feature is deactivated based on its definition, thus I changed the creative tabs icon from the controller (deactivated through channels features) to the fluix crystal.

I probably need to look into the code how it is now handled (But currently not enough time). Before it would just iterate over a list and look for the first non null item. (But depending on how that icon is created, maybe it is too late?)
How would also an addon then check for a deactivated feature?

> Optional

I do not see any particular differences between both. Just that the Java8 one is designed with the Streams API in mind vs. Guava for Java6/7. But probably not the place to discuss it here.

Otherwise I mostly agree with it. One question would probably also, if it is really possible to make a clean design with minecraft at all.

thatsIch:
>  (But depending on how that icon is created, maybe it is too late?)

The creative tabs just crashes, if you try to render a definition which is not registered in the `GameRegistry`, so you just have to register a dummy at pre-init for example to elevate that. Annoying stuff will only happen with NEI, since it lists up all of the stuff in the `GameRegistry`.

> How would also an addon then check for a deactivated feature?

`GameRegistry` would be an easy way to do it I guess, other choice would be to enlarge the `IDefinition` by a method like `boolean isEnabled()` or so. Little changes would be needed for that, since the definitions contain already that information afaik.

> Optional

With Guava you can do `asSet` which lets you use a `for` loop to iterate over the `Optional` set, so it basically auto-checks if the contained is not empty and automatically unwraps it. With JDK8 you need to make it a bit more cumbersome.

> Clean

Probably not, but better scaling.

yueh:
`GameRegistry` as well as a changed `IDefinition` is a pretty major API change and can break current mods just checking for null. As the old API will now never return null (if I have not overlooked something)

`boolean isEnabled()` is probably the best option and should probably also be used to keep the old behaviour and return null for disabled items in the old API.

thatsIch:
After giving it some thoughts, its basically the same deal in the end, if we use some `isEnabled` method. 

Another option would be to use `Optional` on the `block()`, `item()` and `entity()` method or mark it as `@Nullable` (which they are already basically, since an item with access on the `block()` is `null`

yueh:
`isEnabled` is the more explicit way compared to the more implicit with Optional.

Not sure what the best approach would be. But we need to keep the old API stable, so we do not break any addon.

thatsIch:
is any of the Addons even using the definitions?

yueh:
I think RemoteIO does check for controllers or so being available. Maybe also DragonAPI or any other mod made by Reika.
And I would not be suprised, if EnderIO does some funky stuff. But then I do not really care as they more or less try to circumvent the API.

thatsIch:
ok advantage of the `Optional<Block>` approach would be, that stuff like `sameAs` does not need be behind an `Optional` statement, because they would be just `false` in that regard. Not sure how common it is, but I think you can return `new ItemStack( null )` for `stack( int )`,

PR #987, opened 2015-03-08
Improved AEItemStack.compareTo performance.

yueh:
Changed `AEItemDef.itemID` to public, mostly for caching the value. `Item.getIdFromItem()` is already backed by an `IdentityHashMap` and using `System.identityHashCode()`.
Also directly dealing with the internal id guarantees no overflows caused by using `a - b` to compare two integers.

`AEItemStack.compareTo()` will now return as soon as possible. For example there is no need to compare metadata, damage value and NBT data, if we already know that these are two item ids.
Whereever it is possible it also uses `a - b` to compare both values (No overflows).
Furthermore it now does an identity compare between the `NBTTagCompound` before actually comparing them. This results in not calling `compareNBT`, if both are null or the same object.
As we already use a shared `NBTTagCompound` (`AESharedNBT`) it also removes uneccessary calls to `compareNBT`.

These are two Snapshots using the old and new `compareTo`.
https://www.dropbox.com/s/mafdwiiviq82dc2/ae2-aeitemstack-compareto-old.nps?dl=0
https://www.dropbox.com/s/dhz3qhyyngbgxoq/ae2-aeitemstack-compareto-new.nps?dl=0

Both are using the same map with a noticeable impact with shift clicking massive amounts.
The Snapshots are done by filling the whole player inventory with the same crafting operation.

The important call is `ConcurrentSkipListMap.findPredecessor`, which is heavily using `compareTo`.
The reduction of over 50% should of course be taken with a grain of salt. But it is also clearly noticeable ingame.

### DIFF
--- a/src/main/java/appeng/util/item/AEItemDef.java
+++ b/src/main/java/appeng/util/item/AEItemDef.java
@@ -37,7 +37,7 @@ public class AEItemDef
 
 	public int def;
 
-	private final int itemID;
+	public final int itemID;
 	public final Item item;
 	public int damageValue;
 
@@ -62,7 +62,7 @@ public class AEItemDef
 
 	public AEItemDef(Item it) {
 		this.item = it;
-		this.itemID = System.identityHashCode( this.item );
+		this.itemID = Item.getIdFromItem( it );
 	}
 
 	public AEItemDef copy()
--- a/src/main/java/appeng/util/item/AEItemStack.java
+++ b/src/main/java/appeng/util/item/AEItemStack.java
@@ -261,11 +261,19 @@ public final class AEItemStack extends AEStack<IAEItemStack> implements IAEItemS
 	@Override
 	public int compareTo(AEItemStack b)
 	{
-		int id = this.compare( this.def.item.hashCode(), b.def.item.hashCode() );
-		int damageValue = this.compare( this.def.damageValue, b.def.damageValue );
-		int displayDamage = this.compare( this.def.displayDamage, b.def.displayDamage );
-		// AELog.info( "NBT: " + nbt );
-		return id == 0 ? (damageValue == 0 ? (displayDamage == 0 ? this.compareNBT( b.def ) : displayDamage) : damageValue) : id;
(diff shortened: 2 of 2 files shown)


thatsIch:
where in the profile do I see the difference? I cant seem to find the call on `compareTo` directly

yueh:
Just sampling will not show the `compareTo` used in `ConcurrentSkipListMap.findPredecessor()`. This is the important method.

Also the visualvm profiler (not sampler) seems to not pick them up for whatever reason (but then I usually encounter different problems with it itself...).
I have used JProfiler to profile it, which of course will distort the results to some degree. But at least it is way more detailed.

ConcurrentSkipListMap.cpr is the actual method calling `compareTo` called by `ConcurrentSkipListMap.findPredecessor`, which is the following method and the last which will show up in the snapshot (and only without ignoring `java.*` from the sampling).

So by default the visualvm profiler will not display anything below `ItemList.add()` or `CSLM.get()`

thatsIch:
Well you can expect from a commercial software to be better to some degree

yueh:
It is mostly about their different approaches (and JVisualVM is still commercial software developed by oracle. But then oracle :wink: ).
JVisualVM is more an incremental approach and will not transform all classes while it is attaching. Which is the way JProfiler goes. But compared to it, it will take ages to be finished and it is not possible to do a hot code replacement with it. So testing each little change is pretty time consuming.

thatsIch:
Just installed JProfiler 8 for evaluation, is it just me, but it seem not to be able to open the NPS files

yueh:
They have their own format. But I did choose to use visualvm to produce these as it does not affect the results as heavily as JProfiler.

It probably disables inlining and what not, so the actual results are distorted quite much in terms of actual used CPU time. Also it introduces some other issues like finalizing the grid will also take ages and this also causes the reevaluation of all itemlists etc.
So it is nice to see the invocation counts and how `compareTo(AEItemStack)` compares to `compare(int,int)` or `compareNBT()`. But in terms of a localized issue, it is quite limited.

PR #992, opened 2015-03-09
Fixes #743 crash with plugs, fixes #942 builder integration, closes #319 BC 6 prep

### LINKED ISSUE #319: BuildCraft 6.1 preparation
Hello,

Since BuildCraft 6.1 is slowly becoming a more and more stable release, I just wanted to open up an issue stating what has to be changed:

- BuildCraft 6's power API is only available in BuildCraft 6.0. The BuildCraft 5 API should be off by default with BC 6.1+, as while BuildCraft 6.1 supports it, it prefers Redstone Flux greatly. (Note: BuildCraft 6.0 also supports the BC 5 API, meaning the BC 6 API can just be removed at any moment.)
- BuildCraft 6.1.2 or 6.1.3 should bring the IFacadeItem API AlgorithmX2 requested, meaning it will not have to hook inside BC code any longer.

### LINKED ISSUE #743: Buildcraft integration crash
Server crashes whenever I place a BC pipe next to a storage bus. I either have to disable BC integration or delete the pipe using an editor to get it to boot again. Here is the crash log.

http://paste.ee/p/Nh4Y4

I am using:
Buildcraft 6.3.1
AE2 RV2-BETA-7

### LINKED ISSUE #942: AE2 BC Builder Integration
So last night I was working on a funky locomotion miner using annihilation planes (inspired by DW20) The base of this build (bottom two layers, top layer of frame blocks and the bottom being AE2 cables, interfaces, storage bus and annihilation planes).  Just getting two rows of the 18 rows done took me a while so I thought lets create a BC blue print and then use the path markers to let the builder build the remaining rows.  Sadly the builder wont place AE2 cables or parts.  As a low priority future enhancement would it be possible to get AE2 to work with the BC Builder.  

I'm just a user, so this may be a BC problem.  In which case let me know and I'll ask the BC devs. 

This would also be amazing for building SSD or extending them.

Just a thought I wanted to through out as I know there is already some integration with AE2 and BC when it comes to BC pipes.  

However, just in case this actually should have worked and is a config option or mod interaction problem I was in single player FTB Infinity 1.0.2

thatsIch:
Migrates to BC 6.4.2

* requires higher forge, min 1237 by BC, used the recommended 1291 instead ,
* no copying of BC needed anymore, removed that task
* Removes MJ5, MJ6, P2P for MJ

Builder needs testing, cause I have no clue how the builder works :8ball:


yueh:
I have pulled the other PR to upgrade forge and other dependencies. To "fix" the TinyTNT for Pneumaticraft, we need some version between 1250-1270. Afaik it uses 1254 itself. Sticking with the recommended version is still the best option.

If BC did not change the builder heavily, I should be able to test them. But not before this wednesday.
(Or that PR in general)

Moving away from the custom forge repo is also a good idea. I already discussed it with @FireBall1725, but lack of time and such things.

yueh:
Looks good, Builder is also working.

I am curious about the difference between `IPipeTile.hasPipePluggable` and `IPipeTile.hasBlockingPluggable`. Are there some PipePluggable not blocking the pipe and so allow connections?

thatsIch:
I would like to get rid of the `ItemFacade` reference somehow, this is the only access point to the core module as far as I can see, without that, we could just use the BC API and not the deobfuscated version

And from what I can see is the `hasBlockingPluggable` needed for 2 pipes to be aware of each other, that the other is blocked, thus it renders normally (not indicating that it is connected)

It does not matter to use though, since we do not render the Storage Bus different when it is connected.

yueh:
Ok, just looks a bit redundant.

Where is the reference to `ItemFacade`, I only see the `IFacadeItem` interface being used.

thatsIch:
`BC.createFacadePart()`

actually we use a lot of inner classes, like setting up the blocks for the attunement, not sure if we should change this

yueh:
I would say this is something for a new task.
Not crashing is more important for now.

Bestadd:
I know @yueh already tested builder; however, I confirm builder can place cables and parts if Forge 1291, BC 6.4.3 & AE2 PR #992 are installed in single player world and on server.

thatsIch:
Thanks for confirm

PR #996, opened 2015-03-09
Fixes #861 Wireless Terminal notifies player, if it is unlinked

### LINKED ISSUE #861: GUI of Wireless terminal flashes shortly on right-click when wireless terminal is unlinked
I reported this issue a while back, but back then it was closed and commented as "not valid", but it is a valid bug.

If you spam-rightclick a **charged** and **unlinked** wireless terminal, you will sometimes see a terminal interface flash up for a short moment and close immediately again. I assume there is a check missing for a existing link?

Also as it is opening a GUI every single time key pressings are cancelled (press shift while you spam, you will un-crouch whenever you click because a GUI opens and crouching stops working. Also try walking a rightclick, you will stop)

All demonstrated here: https://www.youtube.com/watch?v=wpa-C-428y0

AE [rv2-beta-1](http://prntscr.com/6249rd)
Forge [10.13.2.1272](http://prntscr.com/6249xm)

thatsIch:
Introduces 2 new lang entries, which were not existent before for:

* Device is not a wireless terminal (bug, should not happen)
* Device is not linked up

The trick is, to check the encryption key, if a terminal is linked up or not.


thatsIch:
um, not sure why the build is failing oO

yueh:
I have absolutely no idea and TeamCity omitting pretty much the complete log is not helpful.
Maybe it wants a purge of the gradle cache? (Forge Upgrade)

thatsIch:
No clue.. local build also fails, cause IIcon is not defined,
that makes no sense!

yueh:
Local build as well as my personal CI says everything is fine.
I highly recommend purging the ~/.gradle folder entirely (and not just do a `gradle cleanCache`)

thatsIch:
That does not explain why TC is failing though

thatsIch:
mh I just did that and forced pushed it again and now it works oO

thatsIch:
nvm

yueh:
Burn it. :fire:

But the code looks ok.

It might be an idea to also check the `ILocatableRegistry`, if there is still the access point available.

thatsIch:
From what I can see, its only there to see, if the security station is still there, not the access point.

yueh:
Meh. I wanted to write security station.

I also pinged @FireBall1725 on IRC.
I also now realize, that I use gradle 2.3 on my CI. No idea, if that is important.

thatsIch:
On used the grade the repo is providing, we can try to bump it up

thatsIch:
Added check for security station

deprecated `findLocatableBySerial` of the registry, because it returned for some reason an `Object` even though, it can return `ILocatable` which is part of the API.

PR #1016, opened 2015-03-12
Fixes #1015 no longer eat a single item when not able to satisfy the recipe

### LINKED ISSUE #1015: Pattern terminal and NEI integration is eating items
The pattern terminal will eat a single item when filling the pattern using NEI.
This only happens if it cannot satisify the recipe from the network itself and falls back to the player inventory. It also requires the player inventory to be not empty.

The issue was introduced by PR #575.

Fix is already done, I mostly want an issue to link/document it.

yueh:


src/main/java/appeng/core/sync/packets/PacketNEIRecipe.java
```diff
@@ -191,19 +193,23 @@ public void serverPacketData(INetworkInfo manager, AppEngPacket packet, EntityPl
...
-									if ( whichItem == null && playerInventory != null ) {
+									if ( whichItem == null && playerInventory != null )
+									{
+										ItemStack playerItemStack = null;
 										for ( int y = 0; y < playerInventory.getSizeInventory(); y++ )
 										{
 											// Put the item in the test inventory, and check if the recipe is satisfied
-											testInv.setInventorySlotContents(x, playerInventory.getStackInSlot(y));
-											if ( r.matches(testInv, pmp.worldObj) )
+											playerItemStack = playerInventory.getStackInSlot( y );
```
thatsIch: I think the comment does not match the code anymore I guess?

yueh: Changed. That diff is also a great example for the method is nested way too deep.

PR #1034, opened 2015-03-15
Improved MEInventoryHandler.getAccess()

yueh:
Changed the public fields in `MEInventoryHandler` to private and added setters and getters for them. I kept the my prefix to prevent them from clashing with the already existing `IMEInventoryHandler.getAccess()`.

The actual access rights are now only calculated once for every change as well as an additional booleans for read and write access.
Previously there were calculated multiple times with every access to the inventory (`extractItems/injectItems/getAvailableItems`).
`MEInventoryHandler.getAccess()` alone could contribute half of the total cpu time consumed by AE2. The actual calculation for the read/write access were part of other methods, so they are not showing up in the normal profiling, but I suspect they also contributed a good part to the actual cpu time.

It is quite similar to the `compareTo`, the actual methods are pretty fast. But the amount of invokes just add up.

I also changed the `CSLM` used in `NetworkInventoryHandler` back to a `TreeMap`. The only (currently) required use for a `CSLM` is the `ItemList`, only caused be the `MeaningfulIterator`.
The `TreeMap` is a bit faster and every bit helps. (As long as it will not start throwing `CMEs` like the `ItemList`)

### DIFF
--- a/src/main/java/appeng/me/storage/CellInventoryHandler.java
+++ b/src/main/java/appeng/me/storage/CellInventoryHandler.java
@@ -101,14 +101,14 @@ public class CellInventoryHandler extends MEInventoryHandler<IAEItemStack> imple
 					priorityList.add( AEItemStack.create( is ) );
 			}
 
-			this.myWhitelist = hasInverter ? IncludeExclude.BLACKLIST : IncludeExclude.WHITELIST;
+			this.setWhitelist( hasInverter ? IncludeExclude.BLACKLIST : IncludeExclude.WHITELIST );
 
 			if ( !priorityList.isEmpty() )
 			{
 				if ( hasFuzzy )
-					this.myPartitionList = new FuzzyPriorityList<IAEItemStack>( priorityList, fzMode );
+					this.setPartitionList( new FuzzyPriorityList<IAEItemStack>( priorityList, fzMode ) );
 				else
-					this.myPartitionList = new PrecisePriorityList<IAEItemStack>( priorityList );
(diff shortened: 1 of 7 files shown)


thatsIch:
Not sure, if I agree with naming the public accessors "My" too. My is generally used to differentiate between protected and super access, also only for internal affects. Ours is far as I understood only a promoted call.

yueh:
Providing some reliable performance data is quite difficult. If I use the world posted in the issue about looped storage buses/interfaces (and using a sane configuration, not resulting in a full grid traversal), then it looks like a slight performance gain.

The current master usually ticks between 11-12 ms, but can drop to 10.5ms or over 13ms over a longer time.
With this change it usually between 10-11ms, with drops to 9.5 (maybe even 9), but I have never seen values above 12.
So maybe about 10% on average in that case?

Regarding the naming, there is currently no call to "getMyAccess()", so we could just drop my in general and remove the access to the the unmodified `AccessRestriction`. But also not ideal.
Also I want to avoid having `getAccess()` defined by the interface and `getAccessRestriction()` for the actual getter.

thatsIch:
There is for example no difference between `getMyPriority` and `getPriority`. For the `AccessRestriction` I suggest something like `getInternalAccess` or `getUnrestrictedAccessRestriction` or `setBaseAccess`

yueh:
Pushed an update to remove `My` and renamed the methods for AccessRestriction to `(s|g)etBaseAccess()`

PR #1035, opened 2015-03-15
Added a simple time dilation to the tickmanager.

yueh:
**Do not merge!**

This is mostly a test for a simple time dilation. Instead of just just speeding up the tickrate of an `IGridTickable` based on conditions like "has work to do", it will also take the time to complete the work into account and potentially slow them down. The slowdown is still limited to the maximum tickrate.

It is disabled by default as wrong values can easily stall the performance of every network.

The settings can be mostly considered as guidelines and not hard caps. It will mostly use the `timePerTick` option as base value and multiply it with `sizeMultiplier` to give large networks more breathing space.
Each component can still speed up its tickrate, even with exceeding the limit. But the rate for speeding up will be a bit slower.

I have no idea, if I want to see it in AE2. The effect itself is quite interesting as it basically punishes bad network designs and it is separate for every network, so it will not affect other players (as long as the default settings are pretty sane). But the effect itself is not really obvious and clearly lacks visibility for the player (but this also applies to the dynamic tick rates..)

Maybe someone decides to test the CI build for it and leaves some feedback.


thatsIch:
Interesting idea, feel a bit biased towards it. I think the channel system let user think enough about setup already, not sure, if I want to see more pressure applied.

yueh:
It should never be active by default and probably always labeled as experimental.
But it might give some servers the option to lower the pressure.

It will also not provide an instant speedup, but gradually smoothen it.
For example the world used to test the AccessRestriction does initially ramp up to about 17ms per tick and then slowly adjust itself. With the default settings it will idle about 3.5-4.5ms (without the additional change). So they could be increased.

Also there is a clear decline noticable in relation of the distance to the actual storage.

yueh:
We are also currently using a `StopWatch` for the general `TickHandler` to break the loop when running longer than 50ms. This will not stop a single network from overloading the server, but it might be an idea to make it an option.

thatsIch:
Maybe a build considered for ForgeCraft

yueh:
ForgeCraft is always hard to get feedback back, without relaying it over 5 hops or so.
And their mostly volatile setup might not be the ideal one for a very focused test.

yueh:
Small update, mostly tweaked some settings and calculations.

Most important part is a new command to change it at runtime, which makes testing actually doable.
Syntax is `/ae2 timedilation [enable?] [timeLimit] [sizeMultiplier]`. Parameters can be omitted (starting at the last one) and the actual ones are printed back to the console.
Same access level as the chunklogger.

PR #1046, opened 2015-03-17
Fixes #976 Now uses GitHub to retrieve most current version

### LINKED ISSUE #976: appeng.serivces.VersionChecker (almost) constantly checking for updates.
While playing a slightly modified version of ME^4 (Added TD, XR, CC, and BM), as I was experiencing some lag (likely Fluxed Crystals' fault) so I took a CPU profiler snapshot with sfPlayer1's Sampler. In the roughly 10-15 minutes of sampling, appeng.services.VerionChecker.run() and .sleep() had been invoked 140,895 times each. While it may be on its own thread, it still takes up CPU time, and was invoked far too often, 10 times, 5 seconds apart would be smarter in my opinion. It might be that my internet has been unreliable lately, or that my 2 year old A10 is dying, but ever little thing helps.
-Mortvana

thatsIch:
Reworked whole Version Checker with an extensible interface to add any other service later on easier.
The version checker now has its own config file, to collect the different options and extract them from the main config file.
In that you can specify how fine the versions should be checked.

Edited the build file to replace the `@version@` only in the `AEConfig.java` file, to enable checks to determine if the code is executed in a PR, Dev or normal build.


yueh:
Did you consider the rate limitation? The response will then look like this : https://gist.github.com/yueh/c8e7199a50c15e9e41cf
GitHub will also report the expiration date as HTTP Header, so we might use it to prevent it from continously querying as long as it is still locked out.

I have not used GSON besides one or two quick tries, so I have no idea how it handles malformed input.
But expecting it just throws an exception and we use the `EXCEPTIONAL_RELEASE` in this cases.

As there is already a valid test resource (but I do not see it being used?) we should certainly add this case and be safe it will no crash in some rare conditons.

thatsIch:
Per default it only queries once per day

If it fails somehow to retrieve it, it throws an exception and a `MissingVersion` is returned which is basically the lowest you can get as a version `rv0-alpha-0`. This way any local version is higher than the fetched one.

thatsIch:
Regarding the resource, I just had not enough time to write test cases for it.
I would rather spend it on the API definition rework, maybe I can hook up a simple test for validity

yueh:
I would only expect some pretty rare cases, like a ISP heavily using NAT (mobile?) or maybe some server hosters using a HTTP proxy or similar.

The tests are not really important, I would just add the execeed rate limit one and maybe a TODO.
In case anyone has some time to write tests, there are is at least an indication on where to start and some actual data as test fixtures.

Oh and if we start using `assert`, we should also validate that gradle disables them for release and maybe keep active for the deobf build.

thatsIch:
Well, the old version checker did not much more, but I guess I can add the check on the rate limit. 
asserts are auto disabled if you do not provide the `-ea` tag

thatsIch:
The limit is `60` and it resets every hour. Even if somebody would set the interval to `0` hours, he would need to spam restart client with AE2 every minute to exceed this.

yueh:
It will just dump an IOException, but otherwise no effect.

As said, I never expect it to happen with normal use. Just some rare cases of ISPs abusing NAT for all customers or so.

PR #1065, opened 2015-03-21
instance() does not return null anymore, since else it would represent catching a major bug

thatsIch:
also does not catch any throwable anymore.

- Uses enum (as an alternative to public class with private constructor)
- Use static initializer, so it fails, if somebody tries to access the API before AE2 is instantiated.

Works based on [JLS](http://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.1), where a static block is only called, when another code is referencing the member or method.

Also tried to improve the thrown messages to actually explain what is happening.

Works in dev environment and stand-alone API so far


src/api/java/appeng/api/AEApi.java
```diff
@@ -23,17 +23,47 @@
...
- *
  * Entry point for api.
  *
  * Available IMCs:
- *
  */
-public class AEApi
+public enum AEApi
 {
+	;
```
yueh: superfluous?

yueh:
Not sure what I should think about the enum (semantically).
It feels a bit misused to save on a single line of code for explicitely declaring a constructor as private.

Maybe rename `CORE_API_PATH` to `CORE_API_FQN` as it matches the java terminology better?

thatsIch:
You need to declare the `;` since you have no enumerated fields.

Its either or, from coding standards you can either use the private ctor or the enum, The usage of enums avoid code coverage problems in private constructors (private ctor never called, instantiation never tested etc.). 

Also addition to this, lazy initialization creates opportunity for race condition if by some chance, 2 threads are calling that method at the same time.

It probably needs some testing, if I actually understood the JLS correctly. But since it did not crash yet, I guess it works. Normally people wrap it into another holder object and instantiate it from there (lazy final instantiation).

yueh:
I am mostly thinking about enum + static initializer vs private constructor + static initializer.

An enum does technically the same, but implicit (generate a static initializer to instantiate the objects + private constructor).
The more "enumy" way would probably also be having a single `INSTANCE` and then a custom constructor creating it. But that would break the API.

thatsIch:
its less the "enumy" way, but more of a singleton. There are obvious draw-backs for using enums

1. most obvious, its not an enumeration, using something out of its task, is probably bad
2. it offers other methods like, `valueOf` and such of `Enum` 

well, there is obviously this way and to not break the API

http://pastebin.com/pGvuriUS

yueh:
This does partly exist in #1073, so how does it affect this PR?

thatsIch:
Less to review in one chunk?

thatsIch:
Also this provides superior information why it could not access, also there is no need to `try` and `catch` the whole statement all the time, but its sufficient to do it upon first access.

That is at least the theory behind it.

PR #1073, opened 2015-03-22
Fixes #675 No disabled feature should log spam or crash anymore.

### LINKED ISSUE #675: disabling paintballs cause NPE
after disabling paintballs in the config i always get an NPE whilestarting MC.

Extract of the log file: http://pastebin.com/HjbyTzC9

thatsIch:
**Changes**

* `blocks()`
* `parts()`
* `items()`
* `materials()`

and thus use the new re-direct via `definitions()`. All fields do not need to be `@Nullable` anymore, since they can not be `null` anymore.

All definitions are now initialized, no matter what. But SubItems, Items and Blocks are not registered, if by chance are disabled and thus not obtainable.

`AEItemDefinition` was also retired in favor of `ITileDefinition extends IBlockDefinition extends IItemDefinition extends IComparableDefinition` to not pass more aggregated information than actually available. 

Added annotation marker `@Reflected` to indicate, that those constructors etc are used by reflection and thus are not to be removed or the signature to be edited (not complete). These indicate, that there needs to be some change in the future to improve the code calling and maybe performance.

**Intention**

In comparison to the `Optional` way, it has a straight forward implementation. No random `RunTimeException` to catch some cheese case, which propagates into some random `null` and thus finally do just nothing for a code-part. 

**Old Way**
    `AEItemDefinition AEApi.instance().blocks().blocksQuartzOre`

**New Way**
    `AEItemDefinition AEApi.instance().definitions().blocks().quartzOre()`
    `Optional<Block> .maybeBlock()`
    `Optional<Item> .maybeItem()`
    `Optional<Class<? extends TileEntity>> .maybeEntity()`
    `Optional<ItemStack> .maybeStack( int amount )`

are now the leafs of the call, pushing the call onto `Optional` as far as possible.

**Test**
Set all config settings to false. It result into a lot of NPEs which were based on the recipes, but all in all no crashes and no other messages I have seen so far.

Probably needs major testing.

**Notice**


yueh:
Looks ok.
But I would not be suprised, if some AE2 addons might be somehow affected (mosty likely an existing issue now showing up).

For further improvments, the meteor gen should probably moved into their own package instead of the current helpers.

thatsIch:
Any suggestions which package? or just top level `appeng.meteorite`

PR #1103, opened 2015-03-26
Fixes #1013 Energy acceptor can receive RotaryCraft Power again

### LINKED ISSUE #1013: Energy Acceptor does not accept Rotarycraft power
Rotarycraft has recently updated its API, and it appears AE will no longer accept Rotarycraft power directly.

thatsIch:
Needs testing, since I have actually no clue how RtC works :3


thatsIch:
Seems to work

PR #1117, opened 2015-03-28
Fixes #1115 Does not crash anymore without RotaryCraft

### LINKED ISSUE #1115: Start-up Crash with RV2 B20
Crash on startup with Beta 20
Note. Reduced number of mods installed for cleaner crash log. 
http://pastebin.com/EikVeh84

thatsIch:


thatsIch:
It basically did not strip the interface and tried to do some nasty stuff, which basically disabled the base class.

PR #1145, opened 2015-03-29
Using a unified number converter to display short versions of numbers

thatsIch:


thatsIch:
Should handle negative numbers

thatsIch:
Now handles negative numbers and is a singleton

thatsIch:
No, empty chars (zero length) do not exist.

yueh:
Just to make a note about it, the old postfixes were K, M and B. Probably for millions and billions. Now these are normal metric prefixes. So instead of displaying `2B` it is now `2G`.

thatsIch:
We could make it actually a config option

yueh:
There is an issue with Crafting patterns, it will overlay a number over the normal text indicating it is craftable.

PR #1146, opened 2015-03-29
Refactored scrollbars

thatsIch:
Due to the scrollbars being potentially `null` in the base class, I refactored the initaliation of the scrollbars into the specific classes, where they store their own references, which are 100% not `null`.


thatsIch:
Some strange bug with crafting monitor

PR #1167, opened 2015-03-31
Added a check to detect if AE2 and the API are loaded from the same source

yueh:
As the API is internally depending on AE2 (via reflection), it can potentially fail when another mod is also shipping and actually providing the API.

Normally this handled by Forge. But it will fail, if a coremod is shipping the API.
Forge 1.8 will actually crash in that case, but sadly not 1.7.10.

This still needs some testing as I am currently unable to reproduce it.
But it does not crash indev or a gradle build (except a guaranteed one when negating the condition).

It will probably not catch every rare condition, but the obvious ones should work.


thatsIch:
Have not tested, but I feel there should be more information for normal user, like that this is a bug, and shipping APIs it prohibited and such and maybe stating that this should be reported to the author of the corresponding mod.

yueh:
The feedback can be clearly better, but for now I mostly want a PR ready to test once I am able to reproduce it as well as for discussing it.

If I know it does, work I will most likely add a grace period for a few builds and log an error instead of crashing. Maybe also add a config option, to enable users to bypass it at their own risk. (And the state similar to the integration modules)

thatsIch:
I tried it indev with a cherry pick and the API of thaumic tinkerer did not even reach the `ApiContainer` meaning, that only the original AE2 API was loaded. Give me a pointer, with which mods it is supposed to work.
Also some things I noticed:

* since the order can be unpredictable/inconsistent, only fetching the first match may vary from time to time, would be better to store a list of entries and spit out a warning, that other mods tried to ship the AE2 API
* `aeApiContainer` can be  `null`, since the assignment is bound to an `if` clause in the `for` loop. In theory it is impossible, but that is the reason, why `runtime-exception` exist. Should probably outsource it in another method and throw an exception as the last task or so
* method javadoc should probably contain a `@throws` annotation

yueh:
As said, I am currently not able to reproduce it. Just AE2 + Thaumic Tinkerer does not trigger it.
I suspect some other mod interferes with it, the question is which one.

Regarding the issues

1. Not possible, the backing collectiong is a `HashMap` mapping the `API_ID` to `ModContainer`. So there is always exactly 1 container per API. (or none, but that case is ignored and will fail on its own.)
2. `aeApiContainer` can never be `null`, regardless what your IDE says. Most static analyzer are not able to follow the complex code path here. The outer condition with `ModAPIManager.hasAPI()` verifies that an entry exists. So the for will always succeed. It is identically to `if(HashMap.containsKey(API)) { return HashMap.get(API) }`. But whoever designed the API for `ModAPIManager` (or `APIContainer`) did not really put enough thought into it. So I have to iterate over `HashMap.values()` just to fetch an existing entry.
3. Should be done. But the exception (and location) is a bit tentative and I prefer no documentation over potentially wrong documentation (if forgotten to update after a change).

thatsIch:
From what I can see in the Forge code, not much is happening if an already existing API is found

https://github.com/MinecraftForge/FML/blob/master/src/main/java/net/minecraftforge/fml/common/ModAPIManager.java#L150 and https://github.com/MinecraftForge/FML/blob/master/src/main/java/net/minecraftforge/fml/common/ModAPIManager.java#L92

maybe simulate the behaviour, by removing own API? or just write testcase for potential cases

thatsIch:
Is this line really correct?

    final String offendingModId = aeApiContainer.getDependencies().toString();

imo this should be `getModId` cause else you get a trace like this

http://paste.ee/p/Twh33

yueh:
`getModId` returns the id of the actual api, which is always `appliedenergistics2|API` and so completely useless to identify the offending mod.
`aeApiContainer.getDependencies()` will either return an empty list or a list with a single entry of the owner mod id (the actual mod bundling it)

As said, the API is really broken. Most things are only possible with workarounds as the information itself is not publicly available (or only through reflection) and need to imply the behaviour of different things like a HashMap with only `containsKey()`, `values()` and the values itself having a field for the key. It is possible to implement `get(key)`, just needs the most tedious way possible.

PR #1192, opened 2015-04-03
Moved AEBaseBlock.setFeature() to the end of any constructor.

### LINKED ISSUE #1191: Sky Stone Block becomes Sky Stone when placed in world
Hi,

just tested with only appliedenergistics2-rv2-beta-23.jar and no other mod installed.
When you place one of the Sky Stone Variants (Meta 1,2 and 3) in the world it becomes a normal Sky Stone (Meta 0) like the meteorites are made of.

Forge: 10.13.2.1291

yueh:
Otherwise it can cause a registration of incomplete blocks, like missing subtypes.

Fixes #1191

### DIFF
--- a/src/main/java/appeng/block/crafting/BlockMolecularAssembler.java
+++ b/src/main/java/appeng/block/crafting/BlockMolecularAssembler.java
@@ -46,10 +46,10 @@ public class BlockMolecularAssembler extends AEBaseBlock
 	public BlockMolecularAssembler()
 	{
 		super( BlockMolecularAssembler.class, Material.iron );
-		this.setFeature( EnumSet.of( AEFeature.MolecularAssembler ) );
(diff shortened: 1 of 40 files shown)


thatsIch:
to prevent further issues, maybe its better to store only the feature set in a block, but the handler is constructed outside without the `this` parameter of the block.

yueh:
Using this as parameter in constructors is usually a bad idea as it makes parly constructed objects visible and can cause strange sideeffects.

Storing the EnumSet and then lazy evaluate the `AEBlockFeatureHandler` is probably better.
I would probably like to see in in combination of splitting the `AEBaseBlock` into a normal block and a block with a tileentity and no longer create a `TileDefinition` for every block, even the ones without a TE.

PR #1215, opened 2015-04-05
Closes #1054: Adds an Inscriber API for Developers

### LINKED ISSUE #1054: Make inscriber recipes accessible through the API
Currently only the grinder is (somehow) accessible. Inscriber should be also, since there are not many machines in AE2

thatsIch:


thatsIch:
Did some basic testing

Adds some basic interfaces to add either a pressing or an inscription to the registry. Its accessible via 

    AEApi.instance().registries().inscriber()

src/api/java/appeng/api/features/IInscriberRegistry.java
```diff
@@ -0,0 +1,52 @@
...
+	@Nonnull
+	Set<ItemStack> getInputs();
+
+	/**
+	 * add a new recipe the easy way, duplicates will not be added.
+	 * Added recipes will be automatically added to the optionals and inputs.
+	 *
+	 * @param recipe new recipe
+	 */
+	void addRecipe( @Nullable IInscriberRecipe recipe );
```
yueh: I do not think this should be `@Nullable`, adding null as recipe is completely pointless.
As the annotations are mostly hints for the IDE, throwing an exception like `IllegalArgumentException` in case of null should be considered.

src/main/java/appeng/recipes/handlers/InscriberProcess.java
```diff
@@ -0,0 +1,115 @@
...
+			}
+			else
+				throw new RecipeError( "Inscriber recipes cannot have rows, and must have more then one input." );
+		}
+		else
+			throw new RecipeError( "Inscriber recipes must produce a single output." );
+	}
+
+	@Override
+	public boolean canCraft( ItemStack reqOutput ) throws RegistrationError, MissingIngredientError
```
yueh: How can it throw any of these exceptions?

thatsIch: if the person entered a recipe wrongly

like 

    inscribe stuff -> output

missing an optional

thatsIch: oh l64, cause of `getItemStack`

src/api/java/appeng/api/features/IInscriberRegistry.java
```diff
@@ -0,0 +1,69 @@
...
+	@Nonnull
+	IInscriberRecipeBuilder builder();
+
+	/**
+	 * add a new recipe the easy way, duplicates will not be added.
+	 * Added recipes will be automatically added to the optionals and inputs.
+	 *
+	 * @param recipe new recipe
+	 *
+	 * @throws RuntimeException if null is added
```
yueh: Actually an `IllegalArgumentException`

thatsIch: fixed

PR #1240, opened 2015-04-08
Update zh_CN.lang

bakaxyf:


TheJulianJES:
Why do you want to change this back?

thatsIch:
good question

bakaxyf:
Technically speaking, "合成" (lit. compose, compound; assemble) cannot cover the meaning of "craft".
Then I need to have further discussion with my buddies.

TheJulianJES:
Okayyyy....

PR #1269, opened 2015-04-12
Fixes #1256: Using new size logic to determine the abbreviation for a stack size

### LINKED ISSUE #1256: Threshold for conversion on storage monitor at 1k, instead of 10k
Similar issue as described in #1202 
The storage monitor converts the item count to 1k, if more than 999 items are stored. The fix #1238 apparently did not fix it for the storage monitor.
Screenshot
Using version rv2-beta-28 on DW20 pack 1.7.10 1.2.1 (manually updated AE2)

thatsIch:
Removes the usage of the old methods, since they were pretty, but not applicable for our use case. The displayed strings are determined by the size of the to be rendered string. Now the algorithm tries to use as much width as possible before trying to trim it down.

Added tests to reflect the changes and expected behavior. Also using specific interfaces for the corresponding behavior to shield from potential calls to the underlying `enum` singleton implementation.


thatsIch:
Some border case are negative numbers like `-100000` which should be technically `-100K` but due to the width limitation are rendered as `-0M`, which covers technically are pretty wide range from `-100k` to `-999k`

yueh:
Are there even any cases with negative values?
With the limit of 3/4 chars and an additional char for the sign, it really needs to cover a large range, `-1M` would be really `9999 < x < 1999999` in case of the large font or `99999 < x < 1999999` for the small. The small one would be able to return something like `-.6k`, but the effort is not worth it when we do not use them.

thatsIch:
Well, you wanted to include negative numbers xD

yueh:
The previous one was a bit more sane to add negative values, we could discard it for now.

Currently I would only see negative values only for something like additional storage monitors. But that could maybe solved with something else (up/down arrows, colors, etc)

thatsIch:
Good point, altered the stuff to handle only positive and zero values.

PR #1275, opened 2015-04-13
Closes #1270: Improves NEI grinder overlay localization

### LINKED ISSUE #1270: More localization options
*gui.appliedenergistics2.OfSecondOutput* should have %s tag for swapping *any number*% with "Chance for second output.". Is there any other similar strings which may require swapping in other languages?

~~Also, AE2 builds doesn't have pack.mcmeta included. But it is on Github, maybe buildscript doesn't recognize it?~~

€: prioritized the localization problem in this issue -- @thatsIch

thatsIch:


Adaptivity:
Can you change "%1$d%% шанс вторичного выхода." to "Шанс вторичного выхода: %1$d%%" in ru_RU.lang? So I don't have to make another PR.

thatsIch:
I doubt that this is the proper localization? This sounds mainly your preference how to display the information.

Adaptivity:
In some rare cases first variant of translation can sound super weird. But the second variant is universal.

thatsIch:
Just do a PR for that, 
that way you wont miss the credits for that change (and also the blame :D)

Adaptivity:
Ok, after your PR gets pulled.

yueh:
Afaik russian has a pretty complex grammar regarding numbers. Not just depending on singular/plural, but more concrete values like hundreds. At least I have read enough examples where the russian translation will usually obliterate most system when not planned with it from the beginning.

Maybe displaying it in this way prevents some incorrect grammar?

PR #1276, opened 2015-04-13
Fixes #1186: Storage Cells drop upgrades upon disassembling

### LINKED ISSUE #1186: Dismantling an empty storage cell with a fuzzy card voids the card
Hi,

let me just list the steps to reproduce this bug:
1. Get any storage cell
2. Put it into a cell workbench
3. Install a fuzzy card into the cell
4. Take the cell out of the workbench
5. Shift-right click with the cell in your hand
6. You get a storage housing and a storage component, but the fuzzy card is gone.

This happens in version AE2 rv2-beta-8.

thatsIch:
only `ItemBasicStorageCell.java` contains the relevant code, rest was noise to detect the required code to fix the bug

### DIFF
--- a/src/api/java/appeng/api/implementations/items/IStorageCell.java
+++ b/src/api/java/appeng/api/implementations/items/IStorageCell.java
@@ -59,8 +59,20 @@ public interface IStorageCell extends ICellWorkbenchItem
 	 * @param cellItem item
 	 *
 	 * @return number of bytes
+	 *
+	 * @deprecated use {@link IStorageCell#getBytesPerType(ItemStack)}
 	 */
+	@Deprecated
 	int BytePerType( ItemStack cellItem );
+
+	/**
+	 * Determines the number of bytes used for any type included on the cell.
+	 *
+	 * @param cellItem item
+	 *
+	 * @return number of bytes
+	 */
+	int getBytesPerType( ItemStack cellItem );
 
 	/**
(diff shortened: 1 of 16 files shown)


src/api/java/appeng/api/implementations/items/IStorageCell.java
```diff
@@ -59,10 +59,22 @@
...
 	int BytePerType( ItemStack cellItem );
 
 	/**
+	 * Determines the number of bytes used for any type included on the cell.
+	 *
+	 * @param cellItem item
+	 *
+	 * @return number of bytes
+	 */
+	int getBytePerType( ItemStack cellItem );
```
yueh: Maybe bytes as it is never really singular?

thatsIch:
Changed it

yueh:
The cell and housing is directly placed in the player inventory and only dropped when not enough space is left. For example `ItemStack extraB = ia.addItems( this.component.stack( 1 ) );`.
Maybe also handle the upgrades in the same way?

thatsIch:
Added placing item into inventory

---

It should not, its just marked as deprecated and I gave them already a headsup. Internally its still using the old `BytePerType`. So not implementing it, should not crash them

€: tested it, it does not break with their most current version by placing their cells into a ME Drive (since the only call is in `CellInventory`

PR #1299, opened 2015-04-20
Improved exceptions

thatsIch:
Many exceptions got an improvement due to changed class or description or details it is providing.
Is not complete, needs to be done in patches in the regions, where it is needed, since some are just
swallowed.

Removed total usage of pure RuntimeExceptions to 0.


yueh:
Java 7 multi catch would be really nice for some cases.

thatsIch:
Indeed

PR #1312, opened 2015-04-23
Fixed interface dropping items into itself

### LINKED ISSUE #456: ME Interfaceblock push issue with single slot inventory upon crafting
Somebody told me on reddit. Needs investigation. Can somebody with Botania confirm this?

yueh:
Fixes #456


yueh:
This will also prevent part interfaces from dropping items into inventories which they do not actually face.

PR #1323, opened 2015-04-25
Makes maven dependencies de- and installable

thatsIch:
Provides a new configuration `mods` which can be used on Maven Artifacts. That way, we can install them without affecting them on runtime (less load time)

Put the optional tasks into `optional.gradle` where a install task is just a copy task and the deinstall task a delete task

Added maven local to the repositories if for some reason, somebody uses a local maven


thatsIch:
Changed it back to 2.3

PR #1332, opened 2015-04-26
Fixes #1331: Happened on deactivating features for intermediate crafting components

### LINKED ISSUE #1331: Crashed game, then disable some features of mod
Then i use that config:
http://paste.ubuntu.com/10892145/

Game crash:
http://paste.ubuntu.com/10892142/

Also using mod to remove some recipes, mod name CustomConfig.

thatsIch:
If a feature dependency of ItemMultiMaterial was disabled, the returned value was never assigned with the constructed. Pulling out the construction and setting it before checking it, prevents the NPE and also matches the behaviour in ItemMultiPart, where parts are constructed, but never registered.

### DIFF
--- a/src/main/java/appeng/core/features/DamagedItemDefinition.java
+++ b/src/main/java/appeng/core/features/DamagedItemDefinition.java
@@ -19,11 +19,14 @@
 package appeng.core.features;
 
 
+import javax.annotation.Nonnull;
+
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.world.IBlockAccess;
 
 import com.google.common.base.Optional;
+import com.google.common.base.Preconditions;
 
 import appeng.api.definitions.IItemDefinition;
 
@@ -32,9 +35,9 @@ public final class DamagedItemDefinition implements IItemDefinition
 {
 	private final IStackSrc source;
 
-	public DamagedItemDefinition( IStackSrc source )
+	public DamagedItemDefinition( @Nonnull IStackSrc source )
 	{
-		this.source = source;
+		this.source = Preconditions.checkNotNull( source );
 	}
 
 	@Override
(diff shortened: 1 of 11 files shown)


src/main/java/appeng/core/features/DamagedItemDefinition.java
```diff
@@ -32,7 +34,7 @@
 {
 	private final IStackSrc source;
 
-	public DamagedItemDefinition( IStackSrc source )
+	public DamagedItemDefinition( @Nonnull IStackSrc source )
```
yueh: Maybe add an assert as the other definitions are using it?

thatsIch: Probably even using a `Preconditions.checkNotNull`?

yueh: I do not think a `Preconditions` is actually needed here. Probably also not for `ItemMultiMaterial.createMaterial()`.

Both are only internal, therefore under our control. The general convention for `assert`/`throw Exception` is usually `assert` for internal code, which does not deal with external input, so it is enough to explicitly activate them for testing (manually, unit tests, etc) and `throw exception` (or `Preconditions`) in case there is an actual need to validate external data.

thatsIch: You generally define `internal` as `private` so any private method can use them, ctor and public methods are public thus callable from outside of the context. 

http://www.javapractices.com/topic/TopicAction.do?Id=102

yueh: If we draw a line between `private` and `public/protected` (and maybe `package private` somewhere) and not between `api` and say `do not implement`. Then yes, `Preconditions` are better, but should then apply to every `Definition`, as these are all using assert currently.

thatsIch:
Added preconds to every constructor of the definitions

PR #1358, opened 2015-04-29
Removes the unneeded information about the current class in all blocks

thatsIch:
Every Block had a call to its super, passing its own class.
This can easily be simulated by calling `this.getClass()` in the super class.
Also this was basically only used as using a name.
In the future it might be advisable to not use such methods,
since they are prone to refactoring.


yueh:
Could we merge it into master?
It works for me, but I could assume it could break addons if they just extend our internal classes and I would suggest avoid that for a stable build when not absolutely necessary.

thatsIch:
sounds legit

PR #1371, opened 2015-05-01
Adds several badges

thatsIch:


yueh:
AE2 license should be LGPLv3 not GPLv3

thatsIch:
Fixed

PR #1372, opened 2015-05-01
Fixes #1368: Vibration Chamber notifies neighbors on inventory change and state change

### LINKED ISSUE #1368: Comparator attached to vibration chamber does not shut off when last fuel item is consumed
When fuel is placed into a vibration chamber through a variety of methods, the comparator becomes powered as expected. However, when the fuel is consumed the comparator stays lit and only shuts off when a block adjacent to it is updated or when the fuel slot in the vibration chamber's GUI is clicked on.

I found that the comparator does update properly when a stack of fuel is placed in the vibration and then removed manually, so the problem seems to occur when the fuel slot empties after the fuel has been consumed.

I'm using a custom mod pack that includes the following mods:
- AE2 Beta 31
- NEI 1.0.4.95
- CodeChickenCore 1.0.4.35
- FastleafDecay 1.0

thatsIch:


thatsIch:
Changed notification to `markDirty` to indicate an inventory change. The other notifiction can stay to simulate furnace behavior.

PR #1396, opened 2015-05-05
 Generate a Forge Server side only PlayerInteractEvent before accepting PacketClick or PacketPartPlacement.

krwminer:
Since AE2 bypasses the normal server side playerInteract() Event that
can be used by Forge protection mods like MyTown2, this surfaces a Forge
protection event where AE2 is aleady doing a permission check.

### DIFF
--- a/src/main/java/appeng/core/sync/packets/PacketClick.java
+++ b/src/main/java/appeng/core/sync/packets/PacketClick.java
@@ -81,7 +81,6 @@ public class PacketClick extends AppEngPacket
 	@Override
 	public void serverPacketData( INetworkInfo manager, AppEngPacket packet, EntityPlayer player )
 	{
-<<<<<<< HEAD
 		ItemStack is = player.inventory.getCurrentItem();
 		final IItems items = AEApi.instance().definitions().items();
 		final IComparableDefinition maybeMemoryCard = items.memoryCard();
@@ -90,17 +89,6 @@ public class PacketClick extends AppEngPacket
 		if( is != null )
 		{
 			if( is.getItem() instanceof ToolNetworkTool )
-=======
-		PlayerInteractEvent event = ForgeEventFactory.onPlayerInteract( player, Action.RIGHT_CLICK_BLOCK, this.x, this.y, this.z, this.side, player.worldObj );
-		if ( !event.isCanceled() )
-		{
-			ItemStack is = player.inventory.getCurrentItem();
-			final IItems items = AEApi.instance().definitions().items();
-			final IComparableDefinition maybeMemoryCard = items.memoryCard();
-			final IComparableDefinition maybeColorApplicator = items.colorApplicator();
-	
-			if( is != null )
->>>>>>> branch 'master' of https://github.com/krwminer/Applied-Energistics-2.git
 			{
 				PlayerInteractEvent event = ForgeEventFactory.onPlayerInteract( player, Action.RIGHT_CLICK_BLOCK, this.x, this.y, this.z, this.side, player.worldObj );
 				if ( !event.isCanceled() )
@@ -109,7 +97,6 @@ public class PacketClick extends AppEngPacket
 					tnt.serverSideToolLogic( is, player, player.worldObj, this.x, this.y, this.z, this.side, this.hitX, this.hitY, this.hitZ );
 				}
 			}
-<<<<<<< HEAD
 
 			else if( maybeMemoryCard.isSameAs( is ) )
 			{
@@ -123,8 +110,6 @@ public class PacketClick extends AppEngPacket
 				ToolColorApplicator mem = (ToolColorApplicator) is.getItem();
 				mem.cycleColors( is, mem.getColor( is ), 1 );
 			}
-=======
->>>>>>> branch 'master' of https://github.com/krwminer/Applied-Energistics-2.git
 		}
 	}
 }
--- a/src/main/java/appeng/core/sync/packets/PacketClick.java
+++ b/src/main/java/appeng/core/sync/packets/PacketClick.java
@@ -81,9 +81,9 @@ public class PacketClick extends AppEngPacket
 	@Override
 	public void serverPacketData( INetworkInfo manager, AppEngPacket packet, EntityPlayer player )
 	{
-        PlayerInteractEvent event = ForgeEventFactory.onPlayerInteract(player, Action.RIGHT_CLICK_BLOCK, this.x, this.y, this.z, this.side, player.worldObj);
-        if (!event.isCanceled())
-        {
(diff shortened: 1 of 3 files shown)


thatsIch:
It makes no sense to add this kind of functionality to a check if a user has permission to mine it. So anytime I just want to check that, it will always post a break event?

AfterWind:
We are doing this for a server-side only mod to be able to cancel the event and stop any changes to the AE blocks. Although I do think that another parameter should be added to the hasPermission function to tell it if the action is supposed to modify land or not.

yueh:
The event itself is even the type and it will also send fake events causing potentially issues with other mods as we might later decide to not break them.

Just image some other mod listening for real block break events and use it to replace the broken block with something else. So suddenly removing a terminal from a cable will cause the whole cable to be replaced with something else. Or whatever else side effect.

Also if this would actually be the correct implementation, I am pretty sure that forge would have already patched `canMineBlock()` to post a `BlockEvent.BreakEvent`.

krwminer:
The heart of the problem here is that AE2 does not allow the normal server side Forge PlayerInteractEvent. Placing a Terminal on a Cable does not generate a Forge PlayerInteractEvent. Right clicking with any mods wrench causes a client side Forge PlayerInteractEvent that then cancels the Forge PlayerInteractEvent so the server side never sees it. AE2 uses its own client to server packets, (PacketPartPlacement and PacketClick), to have the server perform the function rather than a server side Forge PlayerInteractEvent.

I would be glad to talk about other possible solutions if you feel this would not be the right approach.

thatsIch:
That problem you are currently describing is totally different from the implementation of the PR.

In the PR you are injecting a `BlockBreakEvent` every time something is checking if a player has the required permissions to mine it, or even opening an AE2 GUI (which is hardly applicable for a block breaking event in my opinion)

The real issue you are describing is, that we need to actually know the requirements, so that Forge will fire a `PlayerInteractEvent`.

As you see in `PartPlacement`, we cancel this event after reaching the correct destination. We basically send a `PlayerInteractEvent` from the Client to the `EventBus` and consumed the event on the Server. 

So are you actually telling us, that the correct way to implement this, is to let the event free roam until the end of day? Generally the correct way to handle this, if the man in the middle wants to, is to increase the priority on the `EventBus`. 

The third option might be, that we lower our priority, but since the integration of our service needs to be ensured and the primary goal of the part placement functionality, this would not be my choice of the three

AfterWind:
The issue is not that you cancel the event after you've consumed it. The issue is that it is only client-side and for us to detect it we need some kind of event fired on the server-side.

yueh:
Sure we could just let it propagate to the server and just ignore it on our side. But the only change this will provide, is that you will know your protection has failed and not just silently fail.
Handling both is very likely not possible. The packets are most likely asynchronous, meaning it would only work in some cases they reach the server in the right order, else it will have some nasty side effects.

The issue is just that minecraft/forge is not really working on a lower level than blocks and maybe block sides. AE2 needs at least multiple interaction points per side and this is just not possible with default forge tools and it can even interact with a part on the opposite side of a block as we are using smaller collision boxes.

krwminer:
Take a look at this solution.

yueh:
Let us start with the easy part. Please use our code style and do not introduce your own.

About your actual problem.

The change to `PacketClick` will not solve anything, just break other things.
For example it will prevent players from accessing the inventory for upgrades of their network tool or clearing the settings of a memory card. The only world modifying thing it will prevent is rotating some AE2 blocks.

`PacketPartPlacement` will basically break everything related to interactions with parts.
Even worse it will actually make the protection obsolete. If some players actually want to share some items with friends or even any other player, it is actually impossible to do it without simultaneously giving them the rights to destroy their network or even the whole base depending on how your protection works.

I doubt it is actually possible without a serious refactoring of how AE2 handles the interaction with parts.
Even then this will most likely not work with normal forge events.
For example you cannot distinguish with a `PlayerInteractEvent` between someone opening a public terminal (allowed), opening the GUI for an export bus located on the same cable (most likely not allowed and can be protect by a security station) or wrenching a part of the same cable (world protection).
So we need to identify which event types we need, their permission levels, implement these things, refactor the current code to use it, and finally every protection mod needs to use AE2 as dependency and actively support it.

AfterWind:
We are aware that this is not the best implementation and it might break some things. But this is better than no implementation for any protection. With the current event handling you have there is absolutely no way to protect against the AE blocks which is a major issue for server owners without having something like Bukkit on top.

The player should be aware of the permission that the player gives when let's others access one of the AE blocks.

I do agree that the PacketClick event firing might not even be needed but the PartPlacement is certainly needed.

This might not be a big issue for you guys, but for the server owners it is a big issue. Since a big chunk of the community is driven by servers I think it's a good idea to solve this one way or the other.

krwminer:
First off, I am sorry I submitted badly formed code. The settings on my version of eclipse hid the fact that the indentation was all messed up, (using spaces and tabs), so I did not notice that. I plain missed the spacing difference around ()'s. It was my intention to keep the code style the same.

Secondly I wanted to address some of points you made.

I agree that putting the PlayerInteract check in PacketClick might be a bit much. I could refine that rather easily to just the rotate block case. I do consider it an issue worth addressing since I am sure no one wants their AE2 blocks rotated unless they have permission.

Instead of PacketPartPlacement I could also move the PlayerInteract check inside the right part of the place() method if there is a case I missed where the place method does not either place a part or wrench off a part.

It is possible to allow someone to use a terminal without giving them rights to modify/destroy your network/base. When you catch the PlayerInteract RIGHT_CLICK_BLOCK event simply look at what the current held object is. If it is a wrench, then cancel it to prevent block modification/breaking. Otherwise allow it, (or in the case of MyTown2 it will fall to the container protection).

I also want to point out that the server side PlayerInteract Event has to be caught by a mod AND canceled for it to effect AE2. Most mods do not even catch PlayerInteract Events. As with AE2 why would you do that unless you know for sure you want to.

Lastly please consider that having a security terminal does not do much good at all if anyone can walk up to it and pop it off with a wrench.

We are not trying to break your mod in any way, we love your mod and want it to be used everywhere. At some point you may make your own physical security layer for AE2. But for now, unless there is some way to protect against physical attacks to AE2, it will likely become less popular for servers.

joserobjr:
Tank you @krwminer I patched the JAR in my server and it seems to be working fine, people was stealing cables and terminals in my server and destroying the fun, I hope they accept your pull request, it helped me.

ParadoxDE:
@joserobjr I tried the same and used the repository from krwminer but it doesn't work on my server users can go into other peoples claim and wrench their cables and even the complete machines with the network tool.
Is their a posilibity to fix this thing?

AfterWind:
Currently none, since AE2 does not handle its events properly and leaves us with nothing to work on the server-side, sorry.

ParadoxDE:
Thats bad, but thanks for the answer :)

PR #1408, opened 2015-05-07
Closes #1404: Added logo to AE2

### LINKED ISSUE #1404: Add logo to mcmod.info
One of the little support you have in `mcmod.info` is providing the logo.
Other nice features in the future would be screenshots and update link. These are current unimplemented though.

thatsIch:
the logo needs to be a bit bigger, since the scaling happens depending on the window size and resolution. Technically we only need 250x65, but it will scale it up for low resolution screens.


yueh:
Is there a reason to swap the gradle wrapper with the archive containing the source?

thatsIch:
mh no, it seems it was just updating it, when I was trying to make it work for the missing `mcmod.info`.

yueh:
The download is about 50mb vs 40mb... So dunno about additional sources and docs.

thatsIch:
reverted the properties back.

PR #579, opened 2014-12-11
Added P2P tunnel part for OpenComputers.

fnuecke:
As discussed with @FireBall1725 on IRC, here's a PR adding a P2P tunnel part for OpenComputers networking.

Tested with and without OC, in dev-env and normal MC via vanilla launcher.

One thing that needs specific review are the adjusted enums. I believe the ones where I did not append the entry to the end to be order irrelevant, but verification of that is needed.

PR #1410, opened 2015-05-07
Added P2P tunnel part for OpenComputers.

fnuecke:
See #579.

Fixed the PR, now against master, also some small improvements (e.g. made it use the dynamic tick handling instead of rescheduling to check for connections each time it failed, so that should be nicer).

Re-tested and looking good.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -34,6 +34,7 @@ code_chicken_lib_version=1.1.3.127
 code_chicken_core_version=1.0.4.35
 nei_version=1.0.4.90
 bc_version=6.4.6
+opencomputers_version=1.5.9.21
 
 #########################################################
 # Self Compiled APIs                                    #
--- a/gradle/scripts/dependencies.gradle
+++ b/gradle/scripts/dependencies.gradle
@@ -49,6 +49,12 @@ repositories {
         name = "RX14 Proxy"
         url = "http://mvn.rx14.co.uk/repo/"
     }
+
+    maven {
+        name "OpenComputers Repo"
+        url = "http://maven.cil.li/"
+    }
+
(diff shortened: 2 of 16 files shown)


thatsIch:
So far looks good, I am testing it tomorrow,
you doing quick work! :D

yueh:
`GridFlags.COMPRESSED_CHANNEL` might not be necessary. Except you want to prevent them from being tunneled through another network. But besides ME no other tunnel has this limitation.

fnuecke:
> GridFlags.COMPRESSED_CHANNEL

Ah, right. I oriented myself on the ME tunnel for the `IGridTickable` implementation and copied that over for testing, forgot to remove it again. Removed it.

fnuecke:
Also, if you test with OC the way I did (computer on one end, screen on the other), in particular that it keeps working after exiting and re-entering the world, make sure to [use the latest OC dev build](http://ci.cil.li/job/OpenComputers-dev-MC1.7.10/)'s OpenOS. Depending on the time it takes for things to initialize on chunk load, there was a scenario where GPUs could forget about their screens, but the screens could be back before the computer resumed for the first time, leading to the OS assuming the GPU was still aware of its screen, where it really wasn't. Fixed that in OpenOS now. So this implementation turned out to be a really good test case for that behavior!

thatsIch:
Tested it, working fine for me,
send you a PR which you can merge,
this includes some style fixes and other random crap, the commit message should help

fnuecke:
Merged, thanks for the cleanup!

thatsIch:
just `rebase -i` them out. I think thats just nicer and faster than somebody bossing around to fix this and that

fnuecke:
Sure thing, done!

thatsIch:
Thank you for your contribution!

PR #1417, opened 2015-05-08
Removes the unneeded information about the current class in all blocks

thatsIch:
Every Block had a call to its super, passing its own class.
This can easily be simulated by calling `this.getClass()` in the super class.
Also this was basically only used as using a name.
In the future it might be advisable to not use such methods,
since they are prone to refactoring.


thatsIch:
new for master regarding https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/1358

PR #1425, opened 2015-05-09
Use weaker collection types

thatsIch:


yueh:
Careful with changing overridden methods of vanilla minecraft like getDrops.

thatsIch:
fixed

PR #1426, opened 2015-05-09
reduces overly strong type cast

thatsIch:


thatsIch:
notice https://github.com/thatsIch/Applied-Energistics-2/blob/e-overly-strong-cast/src/main/java/appeng/me/GridNode.java#L628-628

this makes no sense, since they have neither a common ancestor, nor a possible common object.

PR #1428, opened 2015-05-09
Splits AEBaseBlock into AEBaseBlock + AEBaseTileBlock

yueh:
**First iteration not ready to be merged**

Splits the `AEBaseBlock` (with TE support) into a `AEBaseBlock` (no TE) and `AEBaseTileBlock`.

The `BaseBlockRender` is now a generic as some subclasses need a `AEBaseTileBlock` and this is a bit more sane to change the types of the method parameters.
But I will probably rename `BaseBlockRender` to `GenericBlockRender` add 2 intermediate classes `BaseBlockRender extends GenericBlockRender<AEBaseBlock>` and `BaseTileBlockRender extends GenericBlockRender<AEBaseTileBlock>`.

Also a separate commit/PR with an optimise import would be recommended to reduce the clutter.

### DIFF
--- a/src/main/java/appeng/block/AEBaseBlock.java
+++ b/src/main/java/appeng/block/AEBaseBlock.java
@@ -22,27 +22,19 @@ package appeng.block;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.List;
-import javax.annotation.Nullable;
 
 import com.google.common.base.Optional;
-import com.google.common.collect.Lists;
 
 import net.minecraft.block.Block;
-import net.minecraft.block.BlockContainer;
 import net.minecraft.block.material.Material;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.renderer.texture.IIconRegister;
 import net.minecraft.client.resources.IResource;
 import net.minecraft.creativetab.CreativeTabs;
 import net.minecraft.entity.Entity;
-import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.entity.player.EntityPlayer;
-import net.minecraft.inventory.Container;
-import net.minecraft.inventory.IInventory;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
-import net.minecraft.nbt.NBTTagCompound;
-import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.AxisAlignedBB;
 import net.minecraft.util.IIcon;
 import net.minecraft.util.MovingObjectPosition;
@@ -52,17 +44,11 @@ import net.minecraft.world.IBlockAccess;
 import net.minecraft.world.World;
 import net.minecraftforge.common.util.ForgeDirection;
 
-import cpw.mods.fml.relauncher.ReflectionHelper;
 import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
 
-import appeng.api.implementations.items.IMemoryCard;
-import appeng.api.implementations.items.MemoryCardMessages;
-import appeng.api.implementations.tiles.IColorableTile;
-import appeng.api.util.AEColor;
 import appeng.api.util.IOrientable;
 import appeng.api.util.IOrientableBlock;
-import appeng.block.networking.BlockCableBus;
 import appeng.client.render.BaseBlockRender;
 import appeng.client.render.BlockRenderInfo;
 import appeng.client.render.WorldRender;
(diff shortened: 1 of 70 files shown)


thatsIch:
Dont forget to squash

src/main/java/appeng/block/AEBaseBlock.java
```diff
@@ -73,41 +59,32 @@
...
+	protected final Optional<String> featureSubName;
 	@SideOnly( Side.CLIENT )
 	public IIcon renderIcon;
 	protected boolean isOpaque = true;
 	protected boolean isFullSize = true;
 	protected boolean hasSubtypes = false;
 	protected boolean isInventory = false;
+	protected IFeatureHandler handler;
 	@SideOnly( Side.CLIENT )
 	BlockRenderInfo renderInfo;
```
thatsIch: rather loosening the fields contract, better use `private` on all and use `final` getter and setter where required.

src/main/java/appeng/block/AEBaseBlock.java
```diff
@@ -73,41 +59,32 @@
...
-	private IFeatureHandler handler;
-
-	@Nullable
-	private Class<? extends TileEntity> tileEntityType = null;
 
 	protected AEBaseBlock( Material mat )
 	{
-		this( mat, Optional.<String>absent() );
+		this( mat, Optional.<String> absent() );
```
thatsIch: might rework this now, since the split. Can be postpone though to remove the `absent()`

src/main/java/appeng/block/AEBaseBlock.java
```diff
@@ -150,8 +121,8 @@ public String toString()
 
 	public void registerNoIcons()
 	{
-		BlockRenderInfo info = this.getRendererInstance();
-		FlippableIcon i = new FlippableIcon( new MissingIcon( this ) );
+		final BlockRenderInfo info = this.getRendererInstance();
```
thatsIch: That method is based on `this.getRenderer()` which returns an unparameterized `BaseBlockRenderer`. Least contract should be the base block and tile.

src/main/java/appeng/block/AEBaseBlock.java
```diff
@@ -258,25 +215,21 @@ public IIcon getIcon( int direction, int metadata )
 		return this.getRendererInstance().getTexture( ForgeDirection.getOrientation( direction ) );
 	}
 
+	public ICustomCollision getCustomCollision( World w, int x, int y, int z )
```
thatsIch: why `public` also it returns `null` outside of the own object, might be dangerous (potential new NPEs)

src/main/java/appeng/block/AEBaseBlock.java
```diff
@@ -73,41 +59,32 @@
...
 
 
-public class AEBaseBlock extends BlockContainer implements IAEFeature
+public abstract class AEBaseBlock extends Block implements IAEFeature
 {
 	private final String featureFullName;
-	private final Optional<String> featureSubName;
+	protected final Optional<String> featureSubName;
 	@SideOnly( Side.CLIENT )
 	public IIcon renderIcon;
```
thatsIch: I think `renderIcon` is never used anywhere, basically its always `null`

src/main/java/appeng/block/AEBaseStairBlock.java
```diff
@@ -40,7 +40,7 @@ protected AEBaseStairBlock( Block block, int meta, EnumSet<AEFeature> features )
 	{
 		super( block, meta );
 
-		this.features = new StairBlockFeatureHandler( features, this, Optional.<String>absent() );
+		this.features = new StairBlockFeatureHandler( features, this, Optional.<String> absent() );
```
thatsIch: random formatter change :laughing:

src/main/java/appeng/block/AEBaseTileBlock.java
```diff
@@ -0,0 +1,349 @@
...
+import appeng.tile.storage.TileSkyChest;
+import appeng.util.Platform;
+import appeng.util.SettingsFrom;
+
+
+public abstract class AEBaseTileBlock extends AEBaseBlock implements IAEFeature, ITileEntityProvider
+{
+
+	@Nonnull
+	private Class<? extends TileEntity> tileEntityType;
```
thatsIch: is only the case if `setTileEntity` is enforced being called in the constructor.

src/main/java/appeng/block/misc/BlockLightDetector.java
```diff
@@ -70,4 +91,82 @@ public void randomDisplayTick( World w, int x, int y, int z, Random r )
 	{
 		// cancel out lightning
 	}
+
+	@Override
+	protected Class<? extends BaseBlockRender> getRenderer()
```
thatsIch: raw

src/main/java/appeng/block/qnb/BlockQuantumBase.java
```diff
@@ -0,0 +1,78 @@
...
+		if( bridge != null )
+		{
+			bridge.breakCluster();
+		}
+
+		super.breakBlock( w, x, y, z, a, b );
+	}
+
+	@Override
+	protected Class<? extends BaseBlockRender> getRenderer()
```
thatsIch: raw

src/main/java/appeng/block/AEBaseBlock.java
yueh: `protected` should be sufficent. The code currently using it was already dealing with `null`, so NPE will only be a issue with future code. Maybe annotating it `@Nullable` to indicate it.

Otherwise it is really out of scope for this PR and should be dealt separately in one refactoring collisions, if it is really an issue.

src/main/java/appeng/block/AEBaseStairBlock.java
yueh: Formatter fixed, will be part of the next commit.

src/main/java/appeng/block/AEBaseBlock.java
yueh: If I recall correctly, It is currently not really possible to make the IFeatureHandler final.
So it is either loosening the contract or adding getter + setting for it to be able to change it from a subclass when private.

It is probably design decision, if we move to private fields + getter/setter completely or keep the current way with some public/protected/package-private fields.

src/main/java/appeng/block/AEBaseTileBlock.java
yueh: Intentionally it is nonnull, but `BlockCableBus` is preventing it from being final and can potentially create a race conditions in which it is temporarily null until every P2P layer is transformed into the cable.

src/main/java/appeng/block/AEBaseBlock.java
thatsIch: not the field `final` but the methods to ensure the base class contract.

src/main/java/appeng/block/AEBaseBlock.java
yueh: Welcome to the magic of java generics and reflection. :grinning: 

The best option is probably to refactor `AEBaseBlock.getRenderer()` to no longer return a `Class<? extends BaseBlockRender>` and instead directly construct a `BaseBlockRender<Block, Tile>`

The only usage of it is actually in `AEBaseBlock.getRendererInstance()` which is basically a lazy evaluated factory for a singleton. So there is nothing really breaking because it actually needs to know the `Class<?>`

PR #1431, opened 2015-05-09
Changed things in de_DE.lang

TheJulianJES:
If the review is accepted I will also create a PR for rv2 if wanted.


plumpudding:
Naja über das mit "Fluixkristall" oder "Fluix-Kristall" kann man sich streiten..

TheJulianJES:
@plumpudding Why? It's name as item is "Fluix-Kristall" and I just changed other "Fluixkristall" to "Fluix-Kristall".

yueh:
The - in "Fluix-Kristall" is also called "Deppenbindestrich".

"Kristall" as translation for seed is also pretty wrong.

TheJulianJES:
Woops, the 2. is a derp.

TheJulianJES:
Also @yueh do you think I did something wrong with the "Deppenbindestricen"?

plumpudding:
So 4 Germans here ^^
actually if you look at http://www.deppenbindestrich.de/ it explains pretty much why I wouldn't do what you did. Additionally it is inconsistent because the other words are not seperated. 

Fazit: "Das kannste schon so machen aber dann isses halt kacke!"
Nicht zu ernst nehmen wie gesagt kann man schon so machen und ist Ansichtssache.

TheJulianJES:
@plumpudding I would also prefeer to leave the "-" out but if you take a look at http://de.wikipedia.org/wiki/Viertelgeviertstrich#Bindestrich I think it's okay.

thatsIch:
I also would prefer `Fluixkristall`,
you also call it Quarzkristalle for example

TheJulianJES:
@thatsIch Yeah, so should I change all things to that?

thatsIch:
All things?

plumpudding:
No only change back "Fluix-Kristall" to "Fluixkristall"
As both versions are ok and most people like the version without seperation it would be a logic decision to change it back.

TheJulianJES:
@plumpudding But it is also called "Certusquarzkristall" and not "Certusquarz-Kristall"

plumpudding:
You seem to think the wrong way: You changed it from "Fluixkristall" (the better one) to "Fluix-Kristall".

TheJulianJES:
@plumpudding Yeah because the normal item was called "Fluix-Kristall" and other things were just called "Fluixkristall" but I like "Fluixkristall" more than "Fluix-Kristall".

plumpudding:
You still don't understand. You removed "Fluix-Kristall" and added "Fluixkristall" instead.

TheJulianJES:
@plumpudding I know..

plumpudding:
Now I'm beginning to mix them up too. xD

TheJulianJES:
@plumpudding Can happen sometimes ;)

TheJulianJES:
So to make it clear again:
The current name for normal Fluix Crystal is "Fluix-Kristall"
for normal Certus Quartz it is currently "Certusquarz".
So I changed all things with "Fluixkristall" to "Fluix-Kristall" so it is like the current item name.
But I would prefer "Fluixkristall" and even would change it to that if you all ok with that.

plumpudding:
Ok now I got it. Change it all.

TheJulianJES:
Change it all to "Fluixkristall"? @plumpudding

plumpudding:
Yes.

TheJulianJES:
Woops, not done :P

thatsIch:
Could you re-add the last `\n`? Else there might be issues when people are reading the files on different OS

TheJulianJES:
@thatsIch For me there actually is an empty line 450 but GitHub isn't recognizing it..

TheJulianJES:
Nvm was on the wrong fork :P

TheJulianJES:
Another question @thatsIch should I also PR this to rv2 or is it not necessary?

thatsIch:
please do

PR #1432, opened 2015-05-09
Fix warnings in eclipse

TheJulianJES:
This fixes warnings in eclipse, yes some people are still using eclipse :)
If the review is accepted I will also create a PR for rv2 if wanted.

### DIFF
--- a/src/main/java/appeng/services/version/VersionParser.java
+++ b/src/main/java/appeng/services/version/VersionParser.java
@@ -86,7 +86,11 @@ public final class VersionParser
 	{
 		assert PATTERN_VALID_REVISION.matcher( rawRevision ).matches();
 
-		final int revision = new Scanner( rawRevision ).useDelimiter( PATTERN_REVISION ).nextInt();
+		final Scanner scanner = new Scanner( rawRevision );
+
+		final int revision = scanner.useDelimiter( PATTERN_REVISION ).nextInt();
+
+		scanner.close();
 
 		return revision;
 	}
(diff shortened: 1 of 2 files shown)


thatsIch:
What errors are they fixing?

plumpudding:
Yup I was getting that too. There is an error because the resources are included twice in the Eclipse classpath.

thatsIch:
What does eclipse consider a resource?

TheJulianJES:
http://imgur.com/Fqltju1

thatsIch:
There are no resources? Tell eclipse its drunk!

thatsIch:
nvm its probably about the `Scanner` usage

TheJulianJES:
Also @plumpudding to the double resource issue: http://stackoverflow.com/questions/26755674/importing-project-build-path-contains-duplicate-entry-src-for-project

plumpudding:
My Eclipse even showed an error (not a warning). I can reproduce it if you want.

TheJulianJES:
@plumpudding Yeah, that link fixes the error. This PR fixes warnings I had in my two eclipse instances.

thatsIch:
@TheJulianJES it does not fix them, you just hide them, which is bad. Warnings are a good thing, cause they can detect potential errors and bugs.

TheJulianJES:
@thatsIch I thought they are just issues in a eclipse (1. and 2.) and not in idea.

thatsIch:
Generally its not bad, since this object will be cleaned up anyways after the code block is executed. 
Also here it is applied to a String, so no external resources like files are in use, so it is actually eclipse messing up.

TheJulianJES:
Okay, now I'm messed up...

thatsIch:
I accept the `all` fix, the `resource` is just wrong here, since there are no resources in the first place. But closing the `Scanner` is still a good thing, if you do not want to do it, I can do it.

TheJulianJES:
@thatsIch Is it okay?

thatsIch:
Make the `Scanner` final and you are done :)

TheJulianJES:
I usually make things never final..

thatsIch:
Good practice, prevents you from modifying code without your consent

TheJulianJES:
@thatsIch Should I also pull this to rv2 or is it not necessary?

thatsIch:
please do

TheJulianJES:
Will do tomorrow

PR #1434, opened 2015-05-09
Error eclipse for rv2 branch

plumpudding:
Close scanner instead of doing it with a warning

Make the scanner final..

for rv2 branch


TheJulianJES:
I guess you can also do it ;)

PR #1438, opened 2015-05-10
Meteorites can now spawn in stained clay biomes

TheJulianJES:
This add stained clay to the meteorite whitelist so they can spawn in clay biomes.
If requested I can also PR this for rv2.


thatsIch:
Dont think rv2 needs that.

bugfixes and localizations are only thing which are gonna pulled into there.

TheJulianJES:
@thatsIch Okay, good :)

Issue #1379, opened 2015-05-02
Slabs 


Yulife:
Slab variants for the decorative blocks in AE2 would be brilliant, especially the Sky Stone variant. Right now we've just got stairs so slabs would be very neat ~ :smile: 
Thanks in advance

plumpudding:
You can assign me, I'd try to do that

thatsIch:
You can not assign people outside of the organisation

plumpudding:
:O didn't know that. 
So do we have some kind of ReflectionHelper util?

thatsIch:
like the Forge `ReflectionHelper`?

plumpudding:
Forge has one? I'll try that one

plumpudding:
Very little progress but that's what I've done until now: https://github.com/AppliedEnergistics/Applied-Energistics-2/commit/ac076577668299b65fd6970c203ad3cce14e11e4

thatsIch:
nice, the worst part comes, when you need to handle the merging of the slabs upon double place.

yueh:
The properties of slabs should probably be handle inside the slabs and not set through the `IFeatureHandler`.

plumpudding:
@yueh better? https://github.com/AppliedEnergistics/Applied-Energistics-2/compare/master...plumpudding:rv2?diff=unified&name=rv2

plumpudding:
Houston, we have a problem

getting double slabs up and running is not very difficult either, ***but*** we need an ItemBlock which requires the Block to be constructed before the Items are. How would you do that? I'm not very familiar with the AE Api code..

thatsIch:
There should be an `AEBaseItemBlock` which accepts a `Block`.
You probably should extend that class and use that as a return type of the DoubleBlock

FMP is not AE

PR #1439, opened 2015-05-10
add slabs

plumpudding:
see #1379 

could defenitely be done better, but works perfectly fine like this
I tried to keep this compliant to your syntax, shouldn't be too much to fix
do you want this for rv2 too?


TheJulianJES:
Please use the right format, sometimes you used (something) and not ( something )

src/main/java/appeng/block/AEBaseItemBlockSlab.java
```diff
@@ -0,0 +1,15 @@
+package appeng.block;
+
+import net.minecraft.block.Block;
+import net.minecraft.item.ItemSlab;
+
+public class AEBaseItemBlockSlab extends ItemSlab {
+
+	public AEBaseItemBlockSlab(Block b, AEBaseSlabBlock sb, AEBaseSlabBlock db, Boolean id) {
+		super( b, sb, db, id );
+		System.out.println(b);
```
yueh: Remove these.

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,127 @@
...
+	public void registerBlockIcons(IIconRegister reg) { }
+
+	@Override
+	public Item getItemDropped(int meta, Random rand, int fortune)
+	{
+	    return Item.getItemFromBlock(this.slabs);
+	}
+
+	@Override
+	public int quantityDropped(int meta, int fortune, Random random)
```
yueh: Why override it?

plumpudding:
@yueh which exactly do you mean?

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,127 @@
...
+	{
+	    return this.field_150004_a ? 2 : 1;
+	}
+
+	@Override
+	public ItemStack getPickBlock(MovingObjectPosition target, World world, int x, int y, int z)
```
yueh: Do not implement deprecated API, use the alternative.

plumpudding:
more specific? which of them is deprecated?

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,127 @@
...
+        this.setStepSound( block.stepSound );
+        this.useNeighborBrightness = true;
+		if (!isDoubleSlab)
+		{
+			this.dSlabs = new AEBaseSlabBlock( block, meta, features, true, name + ".double" ).setSlabs(this);
+			this.slabs = this;
+		} else {
+			this.dSlabs = this;
+		}
+		this.features = isDoubleSlab ? new SlabBlockFeatureHandler( features, this ) : new SlabBlockFeatureHandler( features, this );
```
yueh: What is the difference here?

plumpudding:
-.- tabs and spaces and eclipse

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,127 @@
...
+
+	public AEBaseSlabBlock( AEBaseBlock block, int meta, EnumSet<AEFeature> features, boolean isDoubleSlab, String name )
+	{
+        super( isDoubleSlab, block.getMaterial() );
+		this.block = block;
+		this.meta = meta;
+		this.name = name;
+		this.isDoubleSlab = isDoubleSlab;
+		this.setBlockName( "appliedenergistics2." + name );
+        this.setHardness( block.getBlockHardness( null, 0, 0, 0 ) );
```
yueh: Why call the getters, if you do not change anything? Same for the `setResistance`.

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,127 @@
...
+        this.setResistance( block.getExplosionResistance( null ) * 5.0F / 3.0F );
+        this.setStepSound( block.stepSound );
+        this.useNeighborBrightness = true;
+		if (!isDoubleSlab)
+		{
+			this.dSlabs = new AEBaseSlabBlock( block, meta, features, true, name + ".double" ).setSlabs(this);
+			this.slabs = this;
```
yueh: Why reassign `this` multiple times?

plumpudding:
will fix all this later today or tomorrow

now have to do homework -.-

src/main/java/appeng/block/AEBaseSlabBlock.java
yueh: It does exactly the same as `BlockSlab.quantityDropped()`.

And use the "Add a line note" button, it is easier to keep the context.

TheJulianJES:
@plumpudding lol

plumpudding:
not funny

yueh:
Line notes are always related to the last displayed line. No idea why github does not highlight them and also show a snippet of the code after it.

TheJulianJES:
@plumpudding O_o I know, good thing I don't have to deal with that anymore :)

plumpudding:
Is that better? @yueh

src/main/java/appeng/block/AEBaseSlabBlock.java
plumpudding: blockhardness and resistance are private in the block. I have to use the getter.

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,131 @@
...
+		this.meta = meta;
+		this.name = name;
+		this.isDoubleSlab = isDoubleSlab;
+		this.setBlockName( "appliedenergistics2." + name );
+		this.setHardness( block.getBlockHardness( null, 0, 0, 0 ) );
+		this.setResistance( block.getExplosionResistance( null ) * 5.0F / 3.0F );
+		this.setStepSound( block.stepSound );
+		this.useNeighborBrightness = true;
+		if (!isDoubleSlab) this.dSlabs = new AEBaseSlabBlock( block, meta, features, true, name + ".double" ).setSlabs(this);
+		this.features = !isDoubleSlab ? new SlabBlockFeatureHandler( features, this ) : null;
```
plumpudding: Double slabs don't need a handler. They are handled by their non-double equivalent's handler.

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,131 @@
...
+	public void registerBlockIcons(IIconRegister reg) { }
+
+	@Override
+	public Item getItemDropped(int meta, Random rand, int fortune)
+	{
+		return Item.getItemFromBlock(this.slabs);
+	}
+
+	@Override
+	public ItemStack getPickBlock(MovingObjectPosition target, World world, int x, int y, int z)
```
plumpudding: The non-deprecated API doesn't work.

src/main/java/appeng/block/AEBaseSlabBlock.java
```diff
@@ -0,0 +1,131 @@
...
+
+	public AEBaseSlabBlock( AEBaseBlock block, int meta, EnumSet<AEFeature> features, boolean isDoubleSlab, String name )
+	{
+		super( isDoubleSlab, block.getMaterial() );
+		this.block = block;
+		this.meta = meta;
+		this.name = name;
+		this.isDoubleSlab = isDoubleSlab;
+		this.setBlockName( "appliedenergistics2." + name );
+		this.setHardness( block.getBlockHardness( null, 0, 0, 0 ) );
```
plumpudding: Hardness and Resistance are private in the block class. I have to use the getter.

plumpudding:
any updates on the review of this?

thatsIch:
It is probably better for scaling reason to use 2 different blocks for double and single slab.

The rest looks fine to me after a squash.

thatsIch:
Rebase required

plumpudding:
I am unable to seperate it into 2 blocks. I can push the broken version to a seperate branch of my fork so you can clone it and work on it if you want.

thatsIch:
what restriction are you reaching?

plumpudding:
I have to access the double slab to register the single slab and the other way arround. So I have to get the other block from the API, but I always get Nullpointers because they are not reachable through the API at that point. It's been some time since I've written that so this is just roughly what the problem was.

thatsIch:
I just did a quick refactor to see the problem. I noticed the stopper, else it would have a cyclic dependency due to the nature of `IAEFeature`

Then a working implementation would suffice for me. We can probably fix this later, if `IAEFeature`s are gonna reworked.

plumpudding:
So tomorrow I'll do some more testing and then squash. Would that be OK then?

thatsIch:
Sure, take your time.

yueh:
The main issues is probably abusing the `AEBaseSlabBlock ` as collection, block and `IAEFeature`.
It might be a good example for moving the `IAEFeature` into its own class, but probably something we should change in general when reworking them and also structuring the packages in a different way.

plumpudding:
I had to recreate the master. See #1500.

PR #1440, opened 2015-05-10
Also adds tag and release shield for rv2

TheJulianJES:


thatsIch:
There is no need for that on rv2, since the display is only visible on the default branch which is the current master.

PR #1441, opened 2015-05-10
Added EMC to SkyStone and CertusQuartz

FireBall1725:


FireBall1725:
Don't pull in yet.

yueh:
You should bug Pahimar about a real api jar, so we do not have to pull all that useless stuff into indev.

TheJulianJES:
@FireBall1725 Why has SkyStone a low EMC value? It would be better if it is higher than obsidian because SkyStone is really uncommon I think. (Just my Opinion)

src/main/java/appeng/integration/IntegrationType.java
```diff
@@ -19,8 +19,12 @@
 package appeng.integration;
 
 
+import scala.Int;
```
TheJulianJES: Is this Scala import needed?

FireBall1725:
@yueh meh

FireBall1725:
Hold on merging this pr,
having issues with AE2 detecting EE3
"Equivalent Exchange 3 - Integration Disabled"

FireBall1725:
@TheJulianJES probably, maybe ill just add it to a config somewhere so you can just set it to what ever you want...

TheJulianJES:
@FireBall1725 Yeah, I would like that :)

FireBall1725:
Right now trying to figure out why it doens't detect EE3

FireBall1725:
@TheJulianJES found out that there is no need for a config in AE2, EE3 creates the config file.  We just set the defaults, and if you want to configure any of the default EMC for any of the mods just edit the one file.

TheJulianJES:
@Fireball1725 Oh, that's nice :)

FireBall1725:
This is good to go, the only thing left is to figure out if we want to add EMC to other things, thinking maybe yes...

FireBall1725:
@TheJulianJES and the scala import is on error as well, just noticed your comment about the import.

TheJulianJES:
@FireBall1725 Okay :)

yueh:
`IIntegrationModule.preInit()` is completely unnecessary. Both `IntegrationStage.PRE_INIT` and `IntegrationStage.INIT` is called during `FMLInitializationEvent` and `PRE_INIT` is really only being used to setup the integration itself. This also means if some integrations are depending on eachother. This also means when they are not fully initialized at this point, it might fail.

The EE3 API sadly does not state when `EnergyValueRegistryProxy.addPreAssignedEnergyValue()` needs to be called (PreInit/Init). If it needs to be called during PreInit, then it is not possible with our current integration system.
But I currently cannot say, if we could move it to PreInit without causing issues.

I also tried a few things with it, but for whatever reasons I could not convince it to assign an EMC value for charged certus nor add any recipe via `RecipeRegistryProxy.addRecipe()`. For example charged certus, the cutting knifes or dusts.

Btw. there is also `certusQuartzCrystal().maybeStack(size)`, so you do not have to tinker with `new ItemStack` directly and things like the metadata.

FireBall1725:
Im going to close this in favor of a different design of adding EMC, will reopen a new PR once that is completed.

Issue #1388, opened 2015-05-04
More infinite power with RotaryCraft - Flywheel


jeremiahwinsley:
I found this one while looking at #1384. This is a separate issue.

I'm using version rv2.stable build 1 of AE2, and version v6f of RotaryCraft, in single player.

Steps to reproduce:

- Place down a RotaryCraft flywheel and a RotaryCraft power source.
- Charge the flywheel, then remove the power source. Using the Angular Transducer, you should see the flywheel speed remain constant.
- Place an Energy Acceptor at the flywheel's output. 
- Using the Network Tool, you'll see that the Energy Acceptor is producing power. Using the Angular Transducer, you'll see that the Energy Acceptor does not actually consume any power from the flywheel.

This may be due to [this function](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/de069980ced2b6525f6102735839271f079f963f/src/main/java/appeng/tile/powersink/RotaryCraft.java#L165-L168) returning 0, where [the api](https://github.com/ReikaKalseki/RotaryCraft/blob/f650baf0d6b88b8bb0e333da3fa86f6eab7832eb/API/Power/PowerAcceptor.java#L23-L26) suggests it should return 1.

yueh:
This is most likely a duplicate of the other issue and there is no need to report every single machine causing issues.

At least until the other is actually fixed and there are still issues left. But that might be caused by the actual issue missing some important information hidden in another issue.

PR #1447, opened 2015-05-11
Fix for #1384 - Infinite Power

### LINKED ISSUE #1384: Infinite power with Rotarycraft
The Energy Acceptor and ME Controller can produce infinite power with no input. 
I'm using version rv2.beta build 32 of AE2, and version v6f of RotaryCraft, in single player.
Steps to reproduce:
- Place down an Energy Acceptor and a RotaryCraft power source.
- Once the power is going into the Energy Acceptor, use a Screwdriver to rotate the RotaryCraft power source. 
- Now remove the RotaryCraft power source. 
- With a Network Tool, you should see the Energy Acceptor producing power by itself.

jeremiahwinsley:
I'm not a Java programmer, but the gradle tests run and the power works, so here's an attempt at fixing this.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -40,7 +40,7 @@ opencomputers_version=1.5.9.21
 # Self Compiled APIs                                    #
 #########################################################
 mekansim_version=8.0.1.198
-rotarycraft_version=V6f
+rotarycraft_version=V6e
 
 
 #########################################################
--- a/src/main/java/appeng/integration/modules/RotaryCraft.java
+++ b/src/main/java/appeng/integration/modules/RotaryCraft.java
@@ -28,7 +28,8 @@ public class RotaryCraft extends BaseModule
 
 	public RotaryCraft()
 	{
-		this.testClassExistence( Reika.RotaryCraft.API.Power.ShaftPowerReceiver.class );
+		this.testClassExistence( Reika.RotaryCraft.API.Power.AdvancedShaftPowerReceiver.class );
+		this.testClassExistence( Reika.RotaryCraft.API.Interfaces.Transducerable.class );
 	}
 
 	@Override
--- a/src/main/java/appeng/tile/powersink/RotaryCraft.java
+++ b/src/main/java/appeng/tile/powersink/RotaryCraft.java
@@ -19,20 +19,25 @@
 package appeng.tile.powersink;
 
 
+import appeng.transformer.annotations.Integration;
+import net.minecraft.world.World;
 import net.minecraftforge.common.util.ForgeDirection;
 
-import Reika.RotaryCraft.API.Power.ShaftPowerReceiver;
+import Reika.RotaryCraft.API.Interfaces.Transducerable;
+import Reika.RotaryCraft.API.Power.AdvancedShaftPowerReceiver;
 
 import appeng.api.config.PowerUnits;
 import appeng.tile.TileEvent;
 import appeng.tile.events.TileEventType;
 import appeng.transformer.annotations.Integration.Interface;
+import appeng.transformer.annotations.Integration.InterfaceList;
 import appeng.transformer.annotations.Integration.Method;
 import appeng.util.Platform;
 
+import java.util.ArrayList;
 
-@Interface( iname = "RotaryCraft", iface = "Reika.RotaryCraft.API.Power.ShaftPowerReceiver" )
-public abstract class RotaryCraft extends IC2 implements ShaftPowerReceiver
+@InterfaceList( value = { @Interface( iname = "RotaryCraft", iface = "Reika.RotaryCraft.API.Power.AdvancedShaftPowerReceiver" ), @Interface( iname = "RotaryCraft", iface = "Reika.RotaryCraft.API.Interfaces.Transducerable") } )
+public abstract class RotaryCraft extends IC2 implements AdvancedShaftPowerReceiver, Transducerable
 {
 
 	private int omega = 0;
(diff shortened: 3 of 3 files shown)


yueh:
Thanks for the PR.

[RotaryCraft.java](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/master/src/main/java/appeng/integration/modules/RotaryCraft.java#L31) also needs to be changed. For example if it is ever renamed, to disable the integration instead of crashing.

I would probably prefer a ticking acceptor instead of directly injecting the power into the network.
There a few rare cases of injecting power needing a reasonable amount of time, mostly in deeply nested networks with hundreds of smaller networks (maybe a bit less, I do not have a real number at hand). Because it needs to traverse every grid in the worst case.
If it takes 1-2ms per tick, it is probably still fine. But there are 5 sides usable on any acceptor, so this has the chance to suddenly take 5-10ms per tick or maybe more. Summing some longs and injecting it once is a bit more forgiving.
Sure it is really rare, but unnecessary calculation should be avoided when possible.

Did you also test it in regards to #1388? As far as I could see, rotarycraft will ensure that the torque is never below. But else I have absolutely no idea how the API works. For example why `addPower` returns a boolean and what it indicates, or if the comment for `AdvancedShaftPowerReceiver` about "forgo this" is related to reseting the power or implementing the methods (maybe including the methods of the supertypes).

jeremiahwinsley:
This does not fix #1388, even with `getMinTorque` returning 1 as recommended in the API documentation. I figured I probably messed something up somewhere, I'll try again later. 

For the ticking acceptor, do you mean a method that runs every few ticks and calls `injectExternalPower` with the total amount since last call? That should be simple to add as long as I can figure out how to tick it. I'm not sure how the power system works, would this cause any problem if the system had no power storage and ran out partway through the tick delay, causing a cycle of shutting down and starting back up? Or is that even possible?

yueh:
Just accumulate the power and inject it once per tick. An acceptor powered by 5 sides would currently inject it 5 times per tick. Maybe more, but I have no idea about the calculations in RC.

For example add amount passed with `addPower()` to the field, and then inject that once per tick and reset it back to 0.
`@TileEvent( TileEventType.TICK )` marked the old method to be called once per tick, but `Tick_RotaryCraft()` is probably a suboptimal name. (And do not annotate `addPower()` with it.)

But else it is really just guessing about how it could actually behave. Which holds me back from fixing it myself. Mostly because I do not have time to dig through the rotarycraft code and try to deduce it.

jeremiahwinsley:
Got it, I've amended my commit to use the original `Tick_RotaryCraft()` instead of replacing it.

jeremiahwinsley:
Missed the other file, I've added that to the commit as well.

yueh:
Also nice that someone knows amend/squashing and we do not have to request it.

Have you any new ideas about #1388? I am really clueless here. Maybe we should not set omega, torque, and power? There is no documentation about it being only used to display these informations?
So as we actually "destroy" the energy through converting it, it might actual expect them to be 0?
But I am just guessing here...

Else it looks fine for me, but I will not be able to really test it today or tomorrow.

jeremiahwinsley:
Still working on #1388. It looks like I'm going to need to compile with a more recent version of the API in order to replace the angular transducer functionality - that's not a blocker, though. Not sure who runs the maven repo, would it be feasible to get the latest version added?

yueh:
Updating the API is sadly not that trivial, because RotaryCraft does not provide a real API.
So we have to compile our own version for each update and strip all the useless stuff away. Mostly because adding every integration with the full mod as dependency increases the loadtime quite a bit and no one wants to wait 5 minutes to start the client and just crash 10 secs later.
As far as I can remember, the API was not even really self-contained, so we still have to keep some non api parts.

jeremiahwinsley:
I've written a [gradle script](https://github.com/jeremiahwinsley/rotarycraft-api) to generate an API jar, with Reika's permission. Would you be able to use this to update?

jeremiahwinsley:
#1388 is due to `getMinTorque` originally returning 0, and a RotaryCraft bug. This pull request, and the next version of RotaryCraft, should see it fixed.

thatsIch:
This would require you to build for us all the time since Reika restricted the cloning of his repo to special authenticated users.

I am looking for a more general purpose solution

thatsIch:
The other solution would be that you use HTTPS url for the submodule

jeremiahwinsley:
Done. You may need `git submodule sync

jeremiahwinsley:
Done. You may need to run `git submodule sync` if you've got it cloned already.

thatsIch:
I implemented this, 
this kicks out the step of using the external jars, 
we cant host them on the CI

https://github.com/AppliedEnergistics/AE2-RotaryCraft-API

thatsIch:
Just need  to setup TeamCity to deploy it to Artifactory

jeremiahwinsley:
It looks like this just pulls in the power API. Could you pull in `Interfaces.Transducerable` too? I think that's the only one I needed.

thatsIch:
sure

thatsIch:
you can already set it to V6f to get the latest API or fetch the jar from TeamCity

jeremiahwinsley:
I've updated it with the latest version, and implemented `Transducerable`. I think that should be everything necessary to replace the old one. Compiles and runs fine with the new jar on my local machine, not sure why it's failing on TeamCity.

thatsIch:
Please add the second interface to the existence test.

jeremiahwinsley:
Done.

thatsIch:
I think you need to rebase, else the build will keep failing

jeremiahwinsley:
Rebased, but still failing. Gradle pulls in the new jar just fine on my machine, not sure what the problem is.

thatsIch:
can you try to switch using the uploaded `V6e` version? I just hope it is because of the dependency resolution.

yueh:
TC probably still uses the old cached build without the interface. Or maybe no redeploy?

thatsIch:
I already tried cleaning the maven dependencies, but I think it also has an intermediate cache layer, thus having no effect, thus I uploaded that version just decremented.

jeremiahwinsley:
That appears to fix it. Running a gradle build with `--refresh-dependencies` should force an update, if you haven't tried that already..

thatsIch:
That's what I tried first.

thatsIch:
You also need to strip down the `Transducerable` interface (else it will crash without RotaryCraft)

jeremiahwinsley:
Done.

thatsIch:
Do I miss something or does the industrial coil not work anymore?
can you show me your test setup?

jeremiahwinsley:
I'm using the creative coil next to an energy acceptor.

thatsIch:
ok, can you squash the commits or did you split them on purpose?

jeremiahwinsley:
I left that one separate because it's not directly related to the feature, but I can squash it if you'd prefer.

PR #1453, opened 2015-05-13
Closes #1452: Implements auto-generation of custom recipes

### LINKED ISSUE #1452: Auto-Generate custom recipes
If custom recipes is set to `true` it expects the user to copy the custom recipes into the config folder. This is rarely applicable, because it requires usage-knowledge.

Auto-Exporting will highly improve the QoL for custom recipes user.

The should copy all files named with `*.recipe` in a specific path to the config location. This is extendable for having `*.recipe` files all over the package system if required.

thatsIch:
Removes config option for `B:customRecipes`

It now generates two folders in your config folder:

* `generated recipes`
* `user recipes`

the entry point is still `index.recipes` but if a replacement can be found in the `user recipes` these are prioritized. That way you do not need to replace all recipe files, but only the one you want to replace.

### DIFF
--- a/src/main/java/appeng/core/RecipeLoader.java
+++ b/src/main/java/appeng/core/RecipeLoader.java
@@ -1,0 +1,99 @@
+/*
+ * This file is part of Applied Energistics 2.
+ * Copyright (c) 2013 - 2015, AlgorithmX2, All rights reserved.
+ *
+ * Applied Energistics 2 is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Applied Energistics 2 is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Applied Energistics 2.  If not, see <http://www.gnu.org/licenses/lgpl>.
+ */
+
+package appeng.core;
+
+
+import java.io.File;
+import java.io.IOException;
+import java.net.URISyntaxException;
+import javax.annotation.Nonnull;
+
+import com.google.common.base.Preconditions;
+
+import org.apache.commons.io.FileUtils;
+
+import appeng.api.recipes.IRecipeHandler;
+import appeng.recipes.loader.ConfigLoader;
+import appeng.recipes.loader.JarLoader;
+import appeng.recipes.loader.RecipeResourceCopier;
+
+
+/**
+ * handles the decision if recipes should be loaded from jar, loaded from file system or force copied from jar
+ *
+ * @author thatsIch
+ * @version rv3 - 12.05.2015
+ * @since rv3 12.05.2015
+ */
+public class RecipeLoader implements Runnable
+{
+	private final IRecipeHandler handler;
+
+	/**
+	 * @param handler handler to load the recipes
+	 *
+	 * @throws NullPointerException if handler is <tt>null</tt>
+	 */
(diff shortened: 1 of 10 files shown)


yueh:
This actually has some pretty serious flaw as it will prevent us from updating any recipe after the user has started it the first time and I doubt most players will manually enforce a reset every time they update AE2.

Maybe add two folders for the recipe to the config folder? One for the default recipes, which will be overwritten with every start (or store the version somewhere and only overwrite with an update) and a second one for the actual custom recipes.

thatsIch:
Self modified recipes are always prioritized, 
this does not change.

thatsIch:
thought about `generated recipes` and `user recipes` folders.
Probably need somehow to highlight the `index` somehow, by either pushing the rest up one higher or something along these lines.

yueh:
The recipes should simultaneously receive an overhaul themselves and be better organized.

thatsIch:
I thought about removing the ´index` and have for each recipe a single file. 
That would make priority recipes maybe easier.

yueh:
One file per recipe might get really tedious to maintain for someone actually wanting to change recipes.
Maybe group them a bit? Like one file for all dense cables (colors), one for upgrade cards, on for QNB related and so on?

And then there is still the question on where to place new recipes. Like crafting processors without inscribers, or recipes for presses.

thatsIch:
Updated description

thatsIch:
This PR handles the auto generation of recipes. There is another issue open for the flattening and the general rework of recipes.

src/main/java/appeng/core/RecipeLoader.java
```diff
@@ -0,0 +1,94 @@
...
+		}
+		// on failure use jar parsing
+		catch( IOException e )
+		{
+			e.printStackTrace();
+			this.handler.parseRecipes( new JarLoader( "/assets/appliedenergistics2/recipes/" ), "index.recipe" );
+		}
+		catch( URISyntaxException e )
+		{
+			e.printStackTrace();
```
yueh: Should probably be changed to a more meaningful `AELog`. Else people will start complaining about "stacktraces are crashing my game" :grin:

src/main/java/appeng/core/RecipeLoader.java
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	@Override
+	public void run()
+	{
+		// setup copying
+		final RecipeResourceCopier copier = new RecipeResourceCopier( "assets/appliedenergistics2/recipes/" );
+		final File configDirectory = AppEng.instance().getConfigDirectory();
+		final File generatedRecipesDir = new File( configDirectory, "generated recipes" );
+		final File userRecipesDir = new File( configDirectory, "user recipes" );
```
yueh: We should probably replace the spaces with something else. For example traversing the folders on linux with a console is slightly annoying as spaces needs to be escaped. So either something like `userRecipes` or `user-recipes` is a bit more convenient.

yueh:
It might also be a good idea to add a readme to each folder.

For example to warn them that the `generated recipes` will be replaced and changes are then lost as well as some information to copy the file over to `user recipes` as well as a link to [custom recipes](http://ae2.ae-mod.info/Custom-Recipes/) for further information.

src/main/java/appeng/core/RecipeLoader.java
```diff
@@ -0,0 +1,94 @@
...
+
+			copier.copyTo( generatedRecipesDir );
+
+			// parse recipes prioritising the user scripts by using the generated as template
+			this.handler.parseRecipes( new ConfigLoader( generatedRecipesDir, userRecipesDir ), "index.recipe" );
+		}
+		// on failure use jar parsing
+		catch( IOException e )
+		{
+			AELog.error( e );
```
yueh: `AElog.error()` itself does a `e.printStackTrace()` and an additional line with the exception class + message. Maybe we should also the stacktrace there.

thatsIch: verb missing xD

thatsIch:
Added readme and copying it over to `generated` and `user` directory. 
Should always be a fresh readme from JAR / FS

PR #1467, opened 2015-05-16
Fix #1362 - Parts display in WAILA

### LINKED ISSUE #1362: All cables, including buses and facades, are now "AE2 cable and/or Bus"
I discovered that with rv2 beta build 30 all my cables are now "AE2 cable and/or bus", so now I can't see how many channels are available on my dense cable that maintain my crafting network.
Some Images

I don't know if this help, I have installed thaumic energistics

TheJulianJES:
I will squash this if the review is accepted. (Fix for #1362)
I would not consider this as a workaround because that is actually done for other things in AE2 and this is also not a WAILA bug. This basicly picks the item from the AE2 multipart and display it to WAILA.
Can also do this to rv2 but I don't think this is really needed.

### DIFF
--- a/src/main/java/appeng/integration/modules/Waila.java
+++ b/src/main/java/appeng/integration/modules/Waila.java
@@ -38,6 +38,7 @@ public class Waila extends BaseModule
 	{
 		final IWailaDataProvider partHost = new PartWailaDataProvider();
 
+		registrar.registerStackProvider( partHost, AEBaseTile.class );
 		registrar.registerBodyProvider( partHost, AEBaseTile.class );
 		registrar.registerNBTProvider( partHost, AEBaseTile.class );
 
--- a/src/main/java/appeng/integration/modules/waila/PartWailaDataProvider.java
+++ b/src/main/java/appeng/integration/modules/waila/PartWailaDataProvider.java
@@ -39,6 +39,7 @@ import appeng.api.parts.IPart;
 import appeng.integration.modules.waila.part.ChannelWailaDataProvider;
 import appeng.integration.modules.waila.part.IPartWailaDataProvider;
 import appeng.integration.modules.waila.part.PartAccessor;
+import appeng.integration.modules.waila.part.PartStackWailaDataProvider;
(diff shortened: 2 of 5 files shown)


thatsIch:
copyright is `2013 - 2015` but it is working nicely and codewise looks straight to me too

TheJulianJES:
Woopss, the other files was wrong, will fix and squash

yueh:
Maybe rename `partDisplay` to just `partStack` or similar? As it is a StackProvider and display nowhere else mentioned.

TheJulianJES:
Will do

TheJulianJES:
Fixed partDisplay -> partStack @yueh 
Fixed copyright @thatsIch

yueh:
Hm. I considered every `Display` => `Stack`. Not just the single parameter.

TheJulianJES:
Ohh, can also do that :)

TheJulianJES:
I think I fixed all @yueh

thatsIch:
pull it if you are ok with it. I am through

TheJulianJES:
If AE already supports WAILA for channels, power etc. it should also support display the cables right. I won't call this a workaround because a workaround is for me using the depracted getPickBlock method or something like that. :)

yueh:
You should also remove the trailing whitespaces.

TheJulianJES:
@yueh Will search and do :)

yueh:
Your IDE should have an option to remove all trailing whitespaces on save. But that is probably not part of the formatter but some general setting. It is a simple way to avoid merge conflicts caused by useless whitespace.

TheJulianJES:
@yueh I think I fixed all :)

TheJulianJES:
Yey ^_^

thatsIch:
You want to do that for rv2 too?

TheJulianJES:
Can vutvrhen I have to change the rv3 Brand to say since rv2, right

TheJulianJES:
? (Handy)

TheJulianJES:
Sure I could, but then I have to change in the rv3 branch the since to rv2, right?

thatsIch:
right

PR #1479, opened 2015-05-18
Make declarations final

thatsIch:
That is probably somewhere too much due to stuff like Immibis Microblocks, 
either make it non-`final` again or strip it via ASM (already happening)


yueh:
Might be a bit too restrictive. Especially making things like `toString()` final. (We should at least create meaningful `toString()` before, that would help a bit for debugging).

thatsIch:
Removed the `final` attribute on `toString` of abstract base classes.

PR #1482, opened 2015-05-18
Closes #1344 Added a silk touch annihilation plane

### LINKED ISSUE #1344: Silk Touch Annihilation Planes
Hello everybody,

on my server, I usually use a CC controlled frame with annihilation planes on it to make a quarry.
What I would love to be able to do is have a card for the annihilation planes that makes them pick up the blocks as if you're using Silk Touch and/or Fortune perhaps.
This would greatly improve the efficiency of frame quarries built in such a manner. 
For balancing reasons one could make the recipe require loads of resources if this seems too overpowered of a feature.

What are your thoughts on that? Would it be too difficult to implement due to the way the annihilation planes are coded? =)

Sincerely, 

Infi

yueh:
Increased power consumption and name is still tentative.
Probably should also be moved to its own feature.


yueh:
Not sure about the indention of the enums in `PartType`.
2 tabs feels a bit strange and other inner classes are not indented

thatsIch:
Just tested this:

1. I think silky feels not like a very *techy* name, maybe something like `Identity Annihilation Plane` sounds better
2. the recipe is shaped, shapeless might be better

yueh:
1. Clearly needs a better name. `Identity` might work, but it is pretty abstract and not really indicating a silk touch functionality.
2. Probably, sometimes the recipes fell a bit inconsistent about shaped/shapeless.
3. Will change it later.

thatsIch:
The idea of Identify was that it mines this block **as is**

yueh:
Personally I am mostly fine with `Identity`. It just might be hard to understand without actually reading the website. Especially as silk touch ist not always "mine this block as is", for example it will not work with wheat (the planted version)

thatsIch:
Could add this information to the item information string list, we are missing a bunch of extra information in-game anyways (hold shift for more information stuff)

yueh:
"Hold shift to replace this line with another actually useful line" :smile:

thatsIch:
well, could be multiple :expressionless: 

(did not know, that GitHub supports Emoji's)

Issue #1450, opened 2015-05-13
Dense Cables show up as "Unnamed"


Iomega0318:
In appliedenergistics2-rv2-stable-1 I have Channels set to False and the Dense Cables are showing up as "Unnamed".

Not a big issue but I thought I would report it.

XFactHD:
When channels are disabled, you actually shouldn't be able to get dense cable.

Iomega0318:
Which makes sense of course, but just wanted to let you know lol. I haven't tried to craft them it's just how they show up in game. Also perhaps if channels are disabled the dense cables shouldn't be disabled? Otherwise how would you use the Quantum Network? Just a thought..

XFactHD:
The dense cable has to be disabled because the renderer that renders the lines for the channel usage needs a serious number between 0 and 32 to work, otherwise they misbehave.

Iomega0318:
Ahh that I did not know, well then perhaps a modification that when channels are disabled the crafting recipe for the bridge change from requiring dense cables to something else? Also I just checked and the "Unnamed" dense cables are craftable in game..

XFactHD:
If they are craftable, this is a serious, and when placing the cable, probably also a world corrupting bug.

Iomega0318:
Good thing I didn't place it then LOL but it is indeed craftable.

yueh:
How about adding some more information like screenshots? Because I cannot reproduce it.

Iomega0318:
I will see if I can get you a short video tomorrow, I don't know why it allows me to craft it but it did..

Iomega0318:
I made sure I was using the latest version of AE and I am not op'd on the server or in creative. Here is a video showing me crafting it in AE.

https://vid.me/Yu6p

thatsIch:
Can you post your configs for AE2 please?

Iomega0318:
Client:
http://paste.ubuntu.com/11252579/

Server:
http://paste.ubuntu.com/11252578/

border999:
Hold up, I have encountered something like this in my work on a private modpack, it would appear that certain versions of NEI cause items to appear to be named unnamed. Not saying this is the same as what's happening to you, just though it would be something you should look into.

yueh:
rv3 will fix it. We should try to cherrypick it into rv2, but the changes are pretty large so it might simply be not feasible.

PR #1483, opened 2015-05-19
Fixes #1474: Prevents crash after placing terminal

### LINKED ISSUE #1474: Crash after placing Crafting terminal:
I had a crafting terminal on a wall not connected to anything and have been waiting for channels to get fixed so I could disable them, after disabling channels and logging in I picked up the crafting terminal and placed it down, I crashed but it didn't give an accurate crash report however on trying to log back in I now get the following crash and cannot log back into the server:

http://paste.ubuntu.com/11196205/

thatsIch:
Should be impossible, but could technically happen.

### DIFF
--- a/src/main/java/appeng/items/parts/ItemMultiPart.java
+++ b/src/main/java/appeng/items/parts/ItemMultiPart.java
@@ -1,6 +1,6 @@
 /*
  * This file is part of Applied Energistics 2.
- * Copyright (c) 2013 - 2014, AlgorithmX2, All rights reserved.
+ * Copyright (c) 2013 - 2015, AlgorithmX2, All rights reserved.
  *
  * Applied Energistics 2 is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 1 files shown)


yueh:
Not really sure about it. It should prevent crashing, but for me it looks like suppressing an actual issue.
Crashing here is really indicating something is wrong and should be further investigated and finally be fixed.

For example the player uses an ancient build of ThaumicTinkerer built at the beginning of january.
This ships the AE2 api from beta b7 or earlier and I have no idea what conflicts can it cause.
 #1450 from the same player might be also related. But without any useful information, it is hard to track down.

thatsIch:
Not sure..

this should be impossible, as I already stated. There is a clear distinction between registered and unregistered items. Unregistered items are items which are disabled by features. 

These should not be seem anywhere (like in NEI or CreativeTab) thus these methods should never be called on a client or server. That given, #1450 might be really related.

yueh:
If there is an issue, it is really somewhere else. Like something messing up the registration of blocks/items. Especially as there should be no way to obtain these ingame.

I mostly want to prevent quick'n'dirty fixes to ignore null or similar and so potentially hiding the real issue.

PR #1486, opened 2015-05-19
Closes #1259 ETA for crafting jobs

### LINKED ISSUE #1259: Estimated Time of Crafting Job Completion
So whenever you have a "huge" crafting job that would take some real world hours, display a estimated time of completion for each CPU and maybe for all CPUs. So you have a rough idea of when it's going to be done.

yueh:
First iteration..


src/main/java/appeng/client/gui/implementations/GuiCraftingCPU.java
```diff
@@ -65,6 +67,8 @@
 	GuiButton cancel;
 	int tooltip = -1;
 
+	private ContainerCraftingCPU craftingCpu;
```
thatsIch: may be `final`

src/main/java/appeng/client/gui/implementations/GuiCraftingCPU.java
```diff
@@ -166,6 +171,12 @@ public void drawFG( int offsetX, int offsetY, int mouseX, int mouseY )
 	{
 		this.fontRendererObj.drawString( this.getGuiDisplayName( GuiText.CraftingStatus.getLocal() ), 8, 7, 4210752 );
 
+		if( this.craftingCpu.eta > 0 )
+		{
+			final String etaString = this.getGuiDisplayName( GuiText.ETA.getLocal() ) + " " + DurationFormatUtils.formatDuration( TimeUnit.MILLISECONDS.convert( this.craftingCpu.eta, TimeUnit.NANOSECONDS ), this.getGuiDisplayName( GuiText.ETAFormat.getLocal() ) );
```
thatsIch: might for readability split this line, looks a bit convoluted

src/main/java/appeng/client/gui/implementations/GuiCraftingCPU.java
```diff
@@ -166,6 +171,12 @@ public void drawFG( int offsetX, int offsetY, int mouseX, int mouseY )
...
+			this.fontRendererObj.drawString( etaString, 8, 165, 4210752 );
```
thatsIch: might outsource to constants for re-use in the future or better description

src/main/java/appeng/container/implementations/ContainerCraftingCPU.java
```diff
@@ -133,6 +137,7 @@ protected void setCPU( ICraftingCPU c )
 				this.list.resetStatus();
 				this.monitor.getListOfItem( this.list, CraftingItemList.ALL );
 				this.monitor.addListener( this, null );
+				this.eta = 0;
 			}
```
thatsIch: can simplify this block due to `null` check in `instanceof`. That way `this.monitor` is never `null`

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -109,6 +111,11 @@
 	private int remainingOperations;
 	private boolean somethingChanged;
 
+	public long lastTime;
+	public long elapsedTime;
+	public long startItemCount;
+	public long remainingItemCount;
+
```
thatsIch: all really needed to be `public` and mutable?

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -1242,6 +1265,34 @@ public void breakCluster()
...
+		this.remainingItemCount = itemCount;
+	}
+
+	private void updateETA( IAEItemStack is )
+	{
+		final long nextStartTime = System.nanoTime();
+		this.elapsedTime += nextStartTime - lastTime;
+		this.lastTime = nextStartTime;
+		this.remainingItemCount -= is.getStackSize();
+		System.out.println( this.remainingItemCount + " " + this.elapsedTime );
```
thatsIch: debug?

src/main/resources/assets/appliedenergistics2/lang/en_US.lang
```diff
@@ -184,6 +184,8 @@ gui.appliedenergistics2.Lumen=Lumen
 gui.appliedenergistics2.Empty=Empty
 gui.appliedenergistics2.Stored=Stored
 gui.appliedenergistics2.Cancel=Cancel
+gui.appliedenergistics2.ETA=ETA:
+gui.appliedenergistics2.ETAFormat=HH:mm:ss
```
thatsIch: Is it possible to give translators only one String? (so on need, they can switch orders or whatever)

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
yueh: public for convenience currently. Mostly to fine tune the calculation.
Mutable is required as they are changing and the multiblock is not destroyed after each job is finished.
Basically first get it working, then make it pretty and use the PR to discuss.

src/main/resources/assets/appliedenergistics2/lang/en_US.lang
yueh: No, the duration format needs to be in ISOwhatevernumber format otherwise it will fail.
We could change it to `ETA: %s` so the order could be changed. But the only other option for format would be actually not make it available.

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
yueh: Yeah, it is a nightmare to actually find the correct itemstacks and let `remainingItemCount` tend towards 0 and not -5134 or +1234. I am not even convinced that it currently works perfect.

src/main/java/appeng/client/gui/implementations/GuiCraftingCPU.java
yueh: Already considered it. If I would know what `4210752` actually means...
It is really something which should be done completely or not at all. Some half done state where some magic numbers are moved to constants with meaningful names and some not is actually worse in my opinion.

thatsIch:
Tested it in-game and it worked fine.

thatsIch:
Squash required

TheJulianJES:
:+1:

thatsIch:
can you add the issue number to the commit. 
Headers will not auto close issues.

PR #1487, opened 2015-05-19
Adds source jar artifact to gradle

thatsIch:


yueh:
Is it intended, that this only covers `main` but not `api`?

thatsIch:
`sourceSets.main` automatically covers the API based on the definition in `sourceSets.main.java.srcDirs += ...`

An execution of `gradlew sourceJar` will quickly proof that statement.

yueh:
Sometimes it is easier to have a single build script to not miss definitions in other files.

PR #1501, opened 2015-05-23
Added support for EE3 and EMC on all AE2 Items

FireBall1725:


TheJulianJES:
:+1:

plumpudding:
Squash

TheJulianJES:
@plumpudding It's easier to review with more commits

FireBall1725:
:+1: good to go...

FireBall1725:
:+1: forge 1401 is good to go...

FireBall1725:
@plumpudding Forge was updated because of stuff with EE3.

FireBall1725:
@thatsIch @yueh testing on ForgeCraft, to check compatibility and stability.  Also, waiting for an update from Pahimar to hopefully fix the tools not having EMC as well as a few of the other items.  There are a few other things that need to be fixed like the water bucket as well, so this should be considered an Alpha of the EE3 compatibility.

FireBall1725:
After testing on ForgeCraft, EMC is 2 on the colored cables, so will need to fix this.

yueh:
I am a bit reluctant on the forge version. Especially the ls branch is still known to cause issues for some users and the normal branch is a bit fluctuating as something gets added and reverted the next day... (The ls branch is even already merged into the main 1.7.10 one)
It is probably ok to use an unstable forge for the rv3 alpha, if there where some indications about the next stable release. But I really do not see any point on making a release (of rv3) build against a non-recommended forge release, especially in case of taking ages until the next recommended release.

But for now it is probably ok while testing it.

FireBall1725:
@yueh I can try to roll back the forge version and see what happens

yueh:
Afaik EE3 requires a later forge version. So it really depends on how EE3 is added indev.

If it provides a real api and not just the full mod, it would not be an issue. Of course directly trying it indev will not work and you have to build a .jar an drop it into a real instance.
So it is a bit more limited, but it will not prevent us from compiling against a lower forge version.

But if it is just the full mod as dependency it will also be load indev, thus completely failing building and launching without using the same (or later) forge version.

FireBall1725:
pushed 947e61a to forgecraft for testing, should be all fixed.

TheJulianJES:
There are still debug messages in it.

yueh:
These comments are not helpful as long it is still developed and debugged.

FireBall1725:
@TheJulianJES there should be no debug messages in it, i moved all debug stuff to the debug log.  Anything on the info log should be stuff that is worthwhile to know.  I'll take a look to make sure, I have been known to accidentally leave debug code in from time to time.

FireBall1725:
@TheJulianJES I don't see anything that would be visible.  AELog.debug will output on the debug channel, meaning you have to modify your log4j file to see them.  It is handy for if there is a problem to leave those in since normally you wont see them.

TheJulianJES:
@FireBall1725 Ah, sorry. I didn't know that.

FireBall1725:
Closing this, will make into a stand alone mod so it can be released.

PR #1505, opened 2015-05-24
Fixes #1474: Prevents crafting of disabled recipes and deletes invalid parts

### LINKED ISSUE #1474: Crash after placing Crafting terminal:

thatsIch:
This will not handle items in inventory

### DIFF
--- a/src/api/java/appeng/api/definitions/IItemDefinition.java
+++ b/src/api/java/appeng/api/definitions/IItemDefinition.java
@@ -1,10 +1,33 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2013 - 2015 AlgorithmX2
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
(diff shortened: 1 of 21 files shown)


yueh:
Not sure why it is closed as it looks nearly complete and I have pulled and improved it a bit.

`ItemMultiPart.processMetaOverlap()` only needs to ignore the disabled parts, then they will no longer appear in NEI or creative tabs. But still exists as invalid items.

If we add a basic `NullPart`, we can even keep disabled items in world and do not need to delete them. Which means reenabling them will bring them back. Also we probably have to move the metadata from -1 for the InvalidType to Short.MAX_VALUE.

I basically copied the cable anchor for a trivial implementation and it looks like this with storage buses.
nullpart
Reenabling will bring the storage buses back, but they lose their config.

thatsIch:
cause I was sick and tired of this xD
Feel free to improve it

PR #1513, opened 2015-05-26
Improve CraftGuide integration

Uristqwerty:
A bug had crept in that caused the CraftGuide to NPE once for each non-AE2 crafting recipe, extending CraftGuide recipe loading time by tens of seconds.

While doing that, I cleaned up the rest of the file by removing two unnessecary interface implementations, and added support for inscriber and grinder recipes.

I hope that I adequately imitated the code style of AE2.

### DIFF
--- a/src/main/java/appeng/integration/modules/CraftGuide.java
+++ b/src/main/java/appeng/integration/modules/CraftGuide.java
@@ -24,7 +24,6 @@ import java.util.List;
 
 import com.google.common.base.Optional;
 
-import net.minecraft.init.Blocks;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.CraftingManager;
 import net.minecraft.item.crafting.IRecipe;
@@ -34,7 +33,6 @@ import uristqwerty.CraftGuide.DefaultRecipeTemplate;
 import uristqwerty.CraftGuide.RecipeGeneratorImplementation;
 import uristqwerty.CraftGuide.api.ChanceSlot;
 import uristqwerty.CraftGuide.api.CraftGuideAPIObject;
-import uristqwerty.CraftGuide.api.CraftGuideRecipe;
 import uristqwerty.CraftGuide.api.ExtraSlot;
 import uristqwerty.CraftGuide.api.ItemSlot;
 import uristqwerty.CraftGuide.api.RecipeGenerator;
@@ -44,6 +42,7 @@ import uristqwerty.CraftGuide.api.Slot;
 import uristqwerty.CraftGuide.api.SlotType;
 import uristqwerty.gui_craftguide.texture.DynamicTexture;
 import uristqwerty.gui_craftguide.texture.TextureClip;
+
 import appeng.api.AEApi;
 import appeng.api.IAppEngApi;
 import appeng.api.definitions.IBlocks;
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/integration/modules/CraftGuide.java
```diff
@@ -22,38 +22,42 @@
 import java.util.Arrays;
 import java.util.List;
 
+import com.google.common.base.Optional;
+
 import net.minecraft.init.Blocks;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.CraftingManager;
 import net.minecraft.item.crafting.IRecipe;
-
```
yueh: The blank lines between the import groups are intentional.

src/main/java/appeng/integration/modules/CraftGuide.java
```diff
@@ -152,56 +158,51 @@ else if( recipe instanceof ShapelessRecipe )
...
 	{
-		this.parent.addRecipe( template, crafting );
+		ItemStack handle = api.definitions().blocks().crankHandle().maybeStack(1).orNull();
+		Slot[] grinderSlots = new ItemSlot[] {
+				new ItemSlot( 3, 21, 16, 16 ).drawOwnBackground(),
+				new ItemSlot( 41, 21, 16, 16, true ).drawOwnBackground().setSlotType( SlotType.OUTPUT_SLOT ),
+				new ChanceSlot( 59, 12, 16, 16, true).setRatio( 10000 ).setFormatString( " (%1$.2f%% chance)" ).drawOwnBackground().setSlotType( SlotType.OUTPUT_SLOT ),
+				new ChanceSlot( 59, 30, 16, 16, true).setRatio( 10000 ).setFormatString( " (%1$.2f%% chance)" ).drawOwnBackground().setSlotType( SlotType.OUTPUT_SLOT ),
+				new ExtraSlot( 22, 12, 16, 16, handle ).clickable( true ).showName( true ).setSlotType( SlotType.MACHINE_SLOT ),
+				new ExtraSlot( 22, 30, 16, 16, grindstone ).clickable( true ).showName( true ).setSlotType( SlotType.MACHINE_SLOT ) };
```
yueh: Could the be extracted to a constant similar to `smallCraftingSlots`? Or can it not share the same instance, which makes `smallCraftingSlots` also questionable? It is a slight performance and memory regression when creating an actual constant object for each recipe.

src/main/java/appeng/integration/modules/CraftGuide.java
```diff
@@ -152,56 +158,51 @@ else if( recipe instanceof ShapelessRecipe )
...
+	private void addInscriberRecipes( IAppEngApi api, ItemStack inscriber, RecipeGenerator generator )
 	{
-		this.parent.addRecipe( recipe, craftingType );
-	}
+		Slot[] inscriberSlots = new ItemSlot[] {
+				new ItemSlot( 12, 21, 16, 16 ).drawOwnBackground(),
+				new ItemSlot( 21, 3, 16, 16 ).drawOwnBackground(),
+				new ItemSlot( 21, 39, 16, 16 ).drawOwnBackground(),
+				new ItemSlot( 50, 21, 16, 16, true ).drawOwnBackground().setSlotType( SlotType.OUTPUT_SLOT ),
+				new ExtraSlot( 31, 21, 16, 16, inscriber ).clickable( true ).showName( true ).setSlotType( SlotType.MACHINE_SLOT ) };
```
yueh: Same as `grinderSlots`

src/main/java/appeng/integration/modules/CraftGuide.java
```diff
@@ -100,11 +92,22 @@ public void generateRecipes( RecipeGenerator generator )
 
 		RecipeTemplate shapelessTemplate = new DefaultRecipeTemplate( this.shapelessCraftingSlots, RecipeGeneratorImplementation.workbench, new TextureClip( DynamicTexture.instance( "recipe_backgrounds" ), 1, 121, 79, 58 ), new TextureClip( DynamicTexture.instance( "recipe_backgrounds" ), 82, 121, 79, 58 ) );
 
-		RecipeTemplate furnaceTemplate = new DefaultRecipeTemplate( this.furnaceSlots, new ItemStack( Blocks.furnace ), new TextureClip( DynamicTexture.instance( "recipe_backgrounds" ), 1, 181, 79, 58 ), new TextureClip( DynamicTexture.instance( "recipe_backgrounds" ), 82, 181, 79, 58 ) );
+		this.addCraftingRecipes( craftingTemplate, smallTemplate, shapelessTemplate, generator );
+
+		final IAppEngApi api = AEApi.instance();
+		IBlocks aeBlocks = api.definitions().blocks();
```
yueh: We usually mark variable/fields as `final` to prevent unintentional mutation of them.
As this is an ongoing project, not every class has already be changed. 
It would be nice, if you could change it and save us a bit of work.

Uristqwerty: Do you mean just mark aeBlocks as final? Or every variable possible?

JackTheRedCreeper:
So this'll fix the huge load time? Finally! :D

thatsIch:
I send you a PR,
you probably require a rebase before you can pull it,
squash that,

then I can pull it.

JackTheRedCreeper:
Hopefully this damn freeze gets fixed because half the AE2 recipes won't show anyways :l

PR #1519, opened 2015-05-26
Fixes #1517 improved performance of TESRWrapper

### LINKED ISSUE #1517: TESR chest/charger/inscriber performance issue
The TESR for chargers, skystone chests and inscribers (notably the wrapper `TESRWrapper`) has a really bad performance caused by calling `glPushAttrib()` and `glPopAttrib()` when a large number of them need to be rendered

[Sampling snapshot with 256 of them](https://www.dropbox.com/s/0c3uif0fnhyxcfd/tesr.nps?dl=0)
(And do not look at the server thread spending 50% of the time with bats...)

yueh:
Instead of spending the majority of the time with `glPushAttrib()` and `glPopAttrib()`, the bottleneck is now the the skystone chest itself. The other blocks like chargers or inscribers are not that taxing in comparison.

This is mostly caused by the chests reusing the vanilla chest model. So this could potentially also be optimized. As far is I understand it, `glPushAttrib()` and `glPopAttrib()` is actually only needed for the chest itself and none of the other blocks. At least I cannot notice any issue with moving it into `RenderBlockSkyChest`.

### DIFF
--- a/src/main/java/appeng/client/render/blocks/RenderBlockSkyChest.java
+++ b/src/main/java/appeng/client/render/blocks/RenderBlockSkyChest.java
@@ -20,6 +20,7 @@ package appeng.client.render.blocks;
 
 
 import org.lwjgl.opengl.GL11;
+import org.lwjgl.opengl.GL12;
 
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.model.ModelChest;
@@ -38,7 +39,11 @@ import appeng.tile.storage.TileSkyChest;
 public class RenderBlockSkyChest extends BaseBlockRender<BlockSkyChest, TileSkyChest>
 {
 
-	final ModelChest model = new ModelChest();
+	private final static ResourceLocation SKY_STONE_CHEST = new ResourceLocation( "appliedenergistics2", "textures/models/skychest.png" );
+	private final static ResourceLocation SKY_BLOCK_CHEST = new ResourceLocation( "appliedenergistics2", "textures/models/skyblockchest.png" );
+	private final static ResourceLocation METADATA_TO_TEXTURE[] = new ResourceLocation[] { SKY_STONE_CHEST, SKY_BLOCK_CHEST };
+
+	private final ModelChest model = new ModelChest();
 
 	public RenderBlockSkyChest()
 	{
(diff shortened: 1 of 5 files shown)


thatsIch:
looking fine

on the note

http://www.cs.rutgers.edu/~decarlo/428/gl_man/pushattrib.html

it is required to push and pop attributes, where you want to save the state of the corresponding variables masked by the respective attribute. Pushing and popping `GL_ALL_ATTRIB_BITS` is very defensive programming, where you can modify any variable. You might want to pull the mirrored calls (begin, end; push & pop etc) more near to the respective calls.

I will mark something in the code as an example

src/main/java/appeng/client/render/blocks/RenderBlockSkyChest.java
```diff
@@ -91,25 +95,26 @@ public void renderTile( AEBaseBlock block, AEBaseTile tile, Tessellator tess, do
...
 
 		if( !skyChest.hasWorldObj() )
 		{
 			return;
 		}
 
-		GL11.glEnable( 32826 /* GL_RESCALE_NORMAL_EXT */ );
+		GL11.glPushAttrib( GL11.GL_POLYGON_BIT );
+		GL11.glEnable( GL12.GL_RESCALE_NORMAL );
 		GL11.glColor4f( 1.0F, 1.0F, 1.0F, 1.0F );
```
thatsIch: marker

src/main/java/appeng/client/render/blocks/RenderBlockSkyChest.java
```diff
@@ -91,25 +95,26 @@ public void renderTile( AEBaseBlock block, AEBaseTile tile, Tessellator tess, do
...
 		if( tile.getWorldObj().getBlockMetadata( tile.xCoord, tile.yCoord, tile.zCoord ) == 1 )
 		{
-			loc = new ResourceLocation( "appliedenergistics2", "textures/models/skyblockchest.png" );
+			loc = SKY_BLOCK_CHEST;
 		}
 		else
 		{
-			loc = new ResourceLocation( "appliedenergistics2", "textures/models/skychest.png" );
+			loc = SKY_STONE_CHEST;
 		}
```
thatsIch: should be able to be moved here 
also best-practive to use `{...}` to indicate the mirror

yueh:
Yeah. It is extremely defensive for a few rare cases. Like broken itemrenderers.
But this is not really a valid reason to wrap everything into `glPushAttrib()` and save the complete state to the stack. Just do be lazy and not enable/disable `GL_?` after they are needed.

I have tinkered a bit more with it, and the issue why the `glPushAttribut` + `glPopAttrib` is needed is because the culling is reverse for the chests (buggy vanilla crap) but afterwards not switched back
An explicit call for it will also elimate the ones in the skychest renderer.

These renders are fairly trivial to change. But I do not recommend looking at the `BaseBlockRender` or the ones for cables. :grinning:

thatsIch:
indeed, these should be moved to the modifying code though, except these are private methods.

PR #1522, opened 2015-05-27
Closes #12 Added visual lock state to monitors

### LINKED ISSUE #12: Indication that Conversion/Storage Monitor are locked/unlocked
Originally Posted by thatsIch

This issue was migrated; For original details see http://ae-mod.info/Tracker/0014/

yueh:
Added a (simple) way of indicating a locked monitor, texture is mostly tentative for now.

Also some general refactoring of every monitor including panels and terminals.
Moved the texture to constants instead of a field in `PartMonitor` and added getters for them.
So they can for example return a different texture based on the state.

This makes it necessary to override the getters in each subclass and which cannot be enforced.
Except if we introduce an abstract class for every type (panel, monitor and terminals), which then handle important parts and just leave the getter for texture abstract.
But that might be a bit overengineered.

Disabled glPushAttrib and glPopAttrib for StorageMonitor as this can be a performance issue and seems to be unecessary/not causing issues for me.
Probably needs further testing.

Removed the double assignment of `this.spin` from the netcode.

Closes #12


src/main/java/appeng/parts/reporting/PartMonitor.java
```diff
@@ -52,18 +52,18 @@
...
 	private static final int BOOTING_FLAG = 8;
 	// CableBusTextures frontSolid = CableBusTextures.PartMonitor_Solid;
-	CableBusTextures frontDark = CableBusTextures.PartMonitor_Colored;
-	CableBusTextures frontBright = CableBusTextures.PartMonitor_Bright;
-	CableBusTextures frontColored = CableBusTextures.PartMonitor_Colored;
-	boolean notLightSource = !this.getClass().equals( PartMonitor.class );
-	byte spin = 0; // 0-3
-	int clientFlags = 0; // sent as byte.
-	float opacity = -1;
+	private boolean notLightSource = !this.getClass().equals( PartMonitor.class );
```
thatsIch: should use the same get style as with the textures.

thatsIch:
I see the problem with the need for an `abstract` base class to enforce the behavior and protect the lifecycle of the parent partmonitor class besides using the super constructor.

Some fields may be applicable to be moved to the constructor if every child needs it.

Besides the code smell with the abstraction problem via `PartMonitor.class` I have nothing to say.

:+1:

src/main/java/appeng/parts/reporting/PartMonitor.java
yueh: Yeah. I am am bit annoyed by the isNot..()

yueh:
We need 3 or 4 abstract classes. One for the base monitor, dealing with the hitboxes and the common models as well as one for each type (panel, monitor and terminal).

The storage monitor and conversion monitor then needs to be splitted and extend the abstract class and no longer conversion extending storage. The same might apply for the other parts.

yueh:
It is probably also questionable, why parts have their own rendering code compared to blocks. These are separate `BlockRenderers` and exists in the client package (where it should belongs)

yueh:
`PartMonitor` is some crazy god class...
Currently I have split it into 4 abstract classes and probably need 1 or 2 two more.

The hierarchy looks currently like:

* **AbstractPartReporting** | Basics, network, nbt, grid handling, old `PartMonitor` stripped down
  * **AbstractPartPanel** | Light sources, panel rendering, shared texture
    * **PartPanel** | brightness
    * **PartSemiDarkPanel** | brightness
    * **PartDarkPanel** | brightness
  * **AbstractPartMonitor** | storage watcher, item rendering, additional nbt/network
    * **PartStorageMonitor** | texture
    * **PartConversionMonitor** | texture, extract items
  * **AbstractPartTerminal** | viewcell, additional nbt/network
    * **PartInterfaceTerminal** | interfaces, does not really fit here
    * **PartTerminal** | textures, terminal specific stuff
    * **PartCraftingTerminal** | textures, terminal specific stuff
    * **PartPatternTerminal** | textures, terminal specific stuff

It really needs something like an `AbstractPartNetworkTerminal` between the `AbstractPartTerminal` and everything network inventory related to separate the `PartInterfaceTerminal` correctly as well as a shared superclass for `AbstractPartMonitor` and `AbstractPartTerminal` to handle the rendering (backside, channel indicator). `PartInterfaceTerminal` really needs some things which are in the current `PartMonitor` but more or less only need for the interface terminal.

Also it affects the GUI handling itself, as this is currently hardcoded to `PartMonitor`, but actually only applies to terminals. So some gui specific interface might also be nice.
Which then brings us to the point of reworking the whole GUI handling...

src/main/java/appeng/parts/reporting/AbstractPartDisplay.java
```diff
@@ -0,0 +1,132 @@
...
+import net.minecraftforge.common.util.ForgeDirection;
+
+import cpw.mods.fml.relauncher.Side;
+import cpw.mods.fml.relauncher.SideOnly;
+
+import appeng.api.parts.IPartRenderHelper;
+import appeng.client.texture.CableBusTextures;
+
+
+public abstract class AbstractPartDisplay extends AbstractPartReporting
```
thatsIch: Might spend some more time writing the javadoc of such classes

why is it there,
what is it doing,
other possible abstract classes you should look into when you consider extending this class.

src/main/java/appeng/parts/reporting/AbstractPartDisplay.java
```diff
@@ -0,0 +1,132 @@
...
+		super( is, true );
+	}
+
+	@Override
+	@SideOnly( Side.CLIENT )
+	public void renderInventory( IPartRenderHelper rh, RenderBlocks renderer )
+	{
+		rh.setBounds( 2, 2, 14, 14, 14, 16 );
+
+		rh.setTexture( CableBusTextures.PartMonitorSides.getIcon(), CableBusTextures.PartMonitorSides.getIcon(), CableBusTextures.PartMonitorBack.getIcon(), this.is.getIconIndex(), CableBusTextures.PartMonitorSides.getIcon(), CableBusTextures.PartMonitorSides.getIcon() );
```
thatsIch: re-use stuff?

src/main/java/appeng/parts/reporting/AbstractPartDisplay.java
```diff
@@ -0,0 +1,132 @@
...
+		rh.renderInventoryBox( renderer );
+	}
+
+	@Override
+	@SideOnly( Side.CLIENT )
+	public void renderStatic( int x, int y, int z, IPartRenderHelper rh, RenderBlocks renderer )
+	{
+		this.renderCache = rh.useSimplifiedRendering( x, y, z, this, this.renderCache );
+
+		rh.setTexture( CableBusTextures.PartMonitorSides.getIcon(), CableBusTextures.PartMonitorSides.getIcon(), CableBusTextures.PartMonitorBack.getIcon(), this.is.getIconIndex(), CableBusTextures.PartMonitorSides.getIcon(), CableBusTextures.PartMonitorSides.getIcon() );
```
thatsIch: or here

src/main/java/appeng/parts/reporting/AbstractPartDisplay.java
```diff
@@ -0,0 +1,132 @@
...
+
+		Tessellator.instance.setColorOpaque_I( this.getColor().mediumVariant );
+		rh.renderFace( x, y, z, this.getFrontDark().getIcon(), ForgeDirection.SOUTH, renderer );
+
+		Tessellator.instance.setColorOpaque_I( this.getColor().blackVariant );
+		rh.renderFace( x, y, z, this.getFrontColored().getIcon(), ForgeDirection.SOUTH, renderer );
+
+		renderer.uvRotateBottom = renderer.uvRotateEast = renderer.uvRotateNorth = renderer.uvRotateSouth = renderer.uvRotateTop = renderer.uvRotateWest = 0;
+
+		rh.setTexture( CableBusTextures.PartMonitorSidesStatus.getIcon(), CableBusTextures.PartMonitorSidesStatus.getIcon(), CableBusTextures.PartMonitorBack.getIcon(), this.is.getIconIndex(), CableBusTextures.PartMonitorSidesStatus.getIcon(), CableBusTextures.PartMonitorSidesStatus.getIcon() );
```
thatsIch: or here

src/main/java/appeng/parts/reporting/AbstractPartMonitor.java
```diff
@@ -0,0 +1,424 @@
...
+import appeng.util.ReadableNumberConverter;
+import appeng.util.item.AEItemStack;
+
+
+/**
+ * @author AlgorithmX2
+ * @author thatsIch
+ * @version rv2
+ * @since rv0
+ */
```
thatsIch: would be nice not copy that, since we did not create this stuff and add your own version, so anybody asking questions for that, can refer to you

src/main/java/appeng/parts/reporting/AbstractPartMonitor.java
```diff
@@ -0,0 +1,424 @@
...
+	private IAEItemStack configuredItem;
+	private boolean isLocked;
+	private IStackWatcher myWatcher;
+	@SideOnly( Side.CLIENT )
+	private boolean updateList;
+	@SideOnly( Side.CLIENT )
+	private Integer dspList;
+
+	@Reflected
+	public AbstractPartMonitor( ItemStack is )
```
thatsIch: general pre-conds are missing

yueh: Added the `Preconditions` to `AEBasePart` as this is really the place it belongs.

src/main/java/appeng/client/gui/implementations/GuiMEMonitorable.java
```diff
@@ -139,7 +140,7 @@ else if( te instanceof IMEChest )
 		{
 			this.myName = GuiText.Chest;
 		}
-		else if( te instanceof PartTerminal )
+		else if( te instanceof AbstractPartTerminal )
```
thatsIch: any chance we can elevate that, while we are at it?

src/main/java/appeng/items/parts/PartType.java
```diff
@@ -111,11 +111,11 @@ public boolean isCable()
...
 	QuartzFiber( 140, EnumSet.of( AEFeature.Core ), EnumSet.noneOf( IntegrationType.class ), PartQuartzFiber.class ),
 
-	Monitor( 160, EnumSet.of( AEFeature.Core ), EnumSet.noneOf( IntegrationType.class ), PartMonitor.class ),
+	Monitor( 160, EnumSet.of( AEFeature.Core ), EnumSet.noneOf( IntegrationType.class ), PartPanel.class ),
 
-	SemiDarkMonitor( 180, EnumSet.of( AEFeature.Core ), EnumSet.noneOf( IntegrationType.class ), PartSemiDarkMonitor.class ),
+	SemiDarkMonitor( 180, EnumSet.of( AEFeature.Core ), EnumSet.noneOf( IntegrationType.class ), PartSemiDarkPanel.class ),
 
-	DarkMonitor( 200, EnumSet.of( AEFeature.Core ), EnumSet.noneOf( IntegrationType.class ), PartDarkMonitor.class ),
+	DarkMonitor( 200, EnumSet.of( AEFeature.Core ), EnumSet.noneOf( IntegrationType.class ), PartDarkPanel.class ),
```
thatsIch: now feels a bit strange, since they are named differently

thatsIch:
Left some comments, nothing major. Just feedback required.

src/main/java/appeng/items/parts/PartType.java
yueh: Yeah, but changing the enum will actually purge these items from the world. It is really just a small inconvenience like the names for stairs.

src/main/java/appeng/client/gui/implementations/GuiMEMonitorable.java
yueh: To which class? This case is shared for every terminal and previously all did inherit `PartTerminal`, howerever `AbstractPartTerminal` is now the common superclass for every terminal.
Raising it to `AbstractPartDisplay` makes no sense, as these do not use a terminal like GUI.

thatsIch: just meant the `instanceof` shit

yueh: Probably another task related to rework the GUI as this will affect mostly unrelated classes (compared to panels)

Issue #1283, opened 2015-04-14
TickHandler / Callable memory leak


sfPlayer1:
The system mentioned above can leak memory when worlds unload if the callable references a world object, even though it's using a WeakHashMap. MeteoriteWorldGen.MeteoriteSpawn is where I've noticed it.

I'd replace Callable with a custom interface that passes the World object with the invocation, removing the need to store it at the target, e.g. void call(World).

PR #1526, opened 2015-05-28
Improves the message generated when generating the folder for the compass data

thatsIch:
Did some internal cleaning of the class


src/main/java/appeng/services/compass/CompassRegion.java
```diff
@@ -85,18 +79,20 @@ private void openFile( boolean create )
 
 	private File getFileName()
 	{
-		String folder = this.rootFolder.getPath() + File.separatorChar + "compass";
-		File folderFile = new File( folder );
+		assert this.rootFolder.isDirectory();
```
yueh: Not sure about the assert here. These are usually disabled and thus not really providing any information outside of indev. Maybe raise as `PreCondition` to the constructor?

yueh:
I am also curios, if this (and the rare meteor issue) could be related to #1283.
Maybe even make a valid/existing folder as precondition (not the guava ones) before actually generating compass/meteor data.

src/main/java/appeng/services/compass/CompassRegion.java
thatsIch: this is actually just checking the invariant, so adding the PreCondition to the current solution is the way to go

thatsIch:
Sure, but then the folder generation needs to be handled before the world generation

yueh:
Which is already the case (for meteors).
It either saves the data directly to disc in case of meteors disabled or at least 1 tick later with meteors. So the folder generation needs to be already done or it would fail more often.

It is not that bad for the compass data as it will just recheck the chunks but not modify them.
But has a way higher impact on meteor generation. Which is of course something for another issue.

thatsIch:
I am currently rewriting this shit.

yueh:
So much code to review :grinning: 

It might be an idea for the future to make some parts available through the public API. Like exposing the meteors, so addons could easily find them and change them somehow. Like additional world gen.

thatsIch:
open new issue for that xD

src/main/java/appeng/core/AppEng.java
```diff
@@ -189,14 +190,14 @@ private void handleIMCEvent( FMLInterModComms.IMCEvent event )
 	@EventHandler
 	private void serverStopping( FMLServerStoppingEvent event )
 	{
-		WorldSettings.getInstance().shutdown();
+		WorldData.onServerAboutToStart();
```
yueh: `onServerAboutToStart` for the stopping event? Is that either incorrectly named or the wrong place?

src/main/java/appeng/core/AppEng.java
```diff
@@ -189,14 +190,14 @@ private void handleIMCEvent( FMLInterModComms.IMCEvent event )
...
 		TickHandler.INSTANCE.shutdown();
 	}
 
 	@EventHandler
 	private void serverAboutToStart( FMLServerAboutToStartEvent evt )
 	{
-		WorldSettings.getInstance().init();
+		WorldData.instance().onWorldStart();
```
yueh: `onServerAboutToStart` is probably the better name for it.

src/main/java/appeng/core/worlddata/SpawnData.java
```diff
@@ -0,0 +1,222 @@
...
+
+	private NBTTagCompound loadSpawnData( int dim, int chunkX, int chunkZ )
+	{
+		if( !Thread.holdsLock( SpawnData.class ) )
+		{
+			throw new IllegalStateException( "Invalid Request" );
+		}
+
+		NBTTagCompound data = null;
+		File file = new File( this.spawnDirectory, dim + "_" + ( chunkX >> 4 ) + '_' + ( chunkZ >> 4 ) + ".dat" );
```
yueh: It might be an idea to move the string generation into its own method. So at least we can test the generated string and guarantee it will always stay consistent.

src/main/java/appeng/core/worlddata/WorldData.java
```diff
@@ -0,0 +1,199 @@
...
+	private final IWorldCompassData compassData;
+	private final IWorldSpawnData spawnData;
+
+	private final List<IOnWorldStartable> startables;
+	private final List<IOnWorldStoppable> stoppables;
+	private final File ae2directory;
+	private final File spawnDirectory;
+	private final File compassDirectory;
+
+	private WorldData( @Nonnull final File worldDirectory )
```
yueh: Maybe also add a `Preconditions` to validate, if the world/level name matches the directory name?

src/main/java/appeng/services/compass/CompassRegion.java
```diff
@@ -85,18 +82,9 @@ private void openFile( boolean create )
 
 	private File getFileName()
```
yueh: The name is a bit misleading as it returns a file and not a file name.

src/main/java/appeng/core/worlddata/WorldData.java
thatsIch: We retrieve it via the Forge API,
no reason to double check it on their side, also it does not matter in which folder we land in the end, since we do not modify the folder itself, but add content to it.

yueh: Mostly yes.
I more or less considered it as a debugging option to indicate if forge does report the wrong name or if our objects are not correctly destroyed. But this probably needs to be in a few more places. (The issue with meteors generating everyhwere because unable to write the spawndata)

thatsIch: Not sure if that makes any sense, since both are retrieved from their API.

The correct solution here, would be to remove the singleton access, since it is totally not a singleton.

PR #1527, opened 2015-05-28
Fixes #1521: Renames release to stable

### LINKED ISSUE #1521: VersionParser wrong tag for stable channel
The version checker currently expect `alpha`, `beta`, and `release` as channel names, but `release` actually needs to be named `stable`.

thatsIch:


yueh:
Not sure if the readme is intentional.

thatsIch:
removed

PR #1551, opened 2015-06-05
No longer invalidate the cache on simulate.

### LINKED ISSUE #1036: Auto-Crating Simulation issue with Storage Buses
Hello I just wanted you to be aware that this is still an issue, I'm running the latest Beta rv2-11 and I'm having issues similar to #682 where having a storage bus attached to a barrel causes simulation hang ups, although the difference is that I don't have multiple barrels, just a single cobblestone barrel and it will lock up any crafting recipe at random.

I know other people have already commented about this, but it has plagued my world for quite a long time now, I just recently figured it out because I went nuts smashing a bunch of cables trying to find the cause.

yueh:
Should fix #1036

### DIFF
--- a/src/main/java/appeng/me/storage/MEMonitorIInventory.java
+++ b/src/main/java/appeng/me/storage/MEMonitorIInventory.java
@@ -86,7 +86,10 @@ public class MEMonitorIInventory implements IMEMonitor<IAEItemStack>
 			out = this.adaptor.addItems( input.getItemStack() );
 		}
 
-		this.onTick();
+		if( type == Actionable.MODULATE )
+		{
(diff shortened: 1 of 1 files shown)


thatsIch:
How to reproduce?

yueh:
I am basically using this setup:

xu nodes

Transfer nodes with max speed upgrades and cycling cobblestone in the chest.
Chest stores some logs and crafting recipes for furnaces, planks and sticks in the interface.

The transfer nodes are fast enough to basically block it nearly every tick. If not even every one. So a chance of 95% or more to reproduce it with requesting sticks.

PR #1561, opened 2015-06-08
Updates BuildCraft to 7.0.9

thatsIch:
Most noise is generated due to the fact, that I renamed the interface and class for being more descriptive

- [x] Update BC 7.0.8
- [X] Split up the dependencies
- [x] Add Javadocs
- [x] Update BC 7.0.9 (11 June 2015)


yueh:
Seems like it still disables the BC integration.

thatsIch:
investigating

thatsIch:
Ok, the problem was, that the names of the stripped interface are assumed to be the name of the `IntegrationType` enum. 

If you refactor them, you need to replace the strings too. 
This can be part of a future improvement for more type-safety.

yueh:
Just to throw it in. Does `BuildCraft|Core` also provide the pipes? Or do they explicitly depend on `BuildCraft|Transport`.

Mostly because we use pipes as storage and can insert items with a storage bus.

thatsIch:
We actually use multiple, 

Core is used for the P2P attunement (engines definition),
also the pipes.

should we split that?

yueh:
Probably a good idea, as BC is now modular and pipes can potentially be missing. Which if course should not disabled other parts.

thatsIch:
Split the BC module class into 3 used classes,

* BC Core is used for wrench stuff (actually not used, can be actually deleted, but maybe someday..)
* BC Builder for builder integration of AE2 stuff
* BC Transport for facade and pipe stuff

Tasklist done

src/main/java/appeng/facade/FacadePart.java
```diff
@@ -342,6 +343,124 @@ else if( this.side == ForgeDirection.NORTH || this.side == ForgeDirection.SOUTH
...
+		ItemStack is = this.getTexture();
+		if( is == null )
+		{
+			return 0;
+		}
+		return is.getItemDamage();
+	}
+
+	@Override
+	public boolean isBC()
```
yueh: Maybe rename it to `isBuildCraftFacade`?

PR #1571, opened 2015-06-11
Update zh_CN.lang

bakaxyf:
Need help
- ```gui.tooltips.appliedenergistics2.ItemsStored=Items Stored: %s
gui.tooltips.appliedenergistics2.ItemsRequestable=Items Requestable: %s```
 
      I cant figure out where are them in game
-  What does "A" in ETA stand for?

### DIFF
--- a/src/main/resources/assets/appliedenergistics2/lang/zh_CN.lang
+++ b/src/main/resources/assets/appliedenergistics2/lang/zh_CN.lang
@@ -194,7 +194,6 @@ gui.appliedenergistics2.Lumen=光通
 gui.appliedenergistics2.Empty=空
 gui.appliedenergistics2.Stored=现存
 gui.appliedenergistics2.Cancel=取消
-gui.appliedenergistics2.ETA=ETA: %s
 gui.appliedenergistics2.ETAFormat=HH:mm:ss
 gui.appliedenergistics2.Crafting=正在合成
 gui.appliedenergistics2.Scheduled=计划合成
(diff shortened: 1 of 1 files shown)


yueh:
The first strings are used in tooltips (see their keys). The `Items Stored: %s` should only appear after the icon flips to using prefixes like K or M (so over 1 or 10k items depending on our settings).

Similar for `Items Requestable: %s`, but I am not not sure if these are actually possible to view currently.
It is mostly a preparation for a logistics pipes integration. I think the only option which could show them is with autocrafting. If I remember correctly, overflow items are still stored inside the crafting cpu until done, but are requestable for other crafting jobs. For example crafting sticks from a single log will also produce 2 left over planks. If you can somehow prevent the sticks from being crafted, it might be possible that the tooltip will also show 2 requestable planks.

ETA stands for "estimated time of arrival", it is certainly not the perfect term for on based on the actual behaviour. Estimated Time To Complete (ETC) would probably be better but is less known.
But there is a good chance that this string will go away in the near future. So it might be pointless to translate it.

bakaxyf:
Oh,thanks. I will update after test.

thatsIch:
Depending on https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/1577

bakaxyf:
I has tested, but I still couldn't find ``Items Requestable: %s``, Maybe there is something I did wrongly
.More detail?
I translated it literally, and will keep test. If have some mistake, I will fix in next update.

bakaxyf:
I think can merge now.

thatsIch:
@yueh, you are required here

thatsIch:
You need to squash them first

bakaxyf:
okey

PR #1591, opened 2015-06-13
Updates BC to 7.0.9 for rv2

thatsIch:
Fixes disabling of BC integration


yueh:
Looks like my issue was caused by another mod also shipping a (old?) BC API, thus breaking the integration.

Just AE2 + BC 7.0.9 works fine.

PR #1604, opened 2015-06-15
Changed to immutable list to prevent direct modifcations

yueh:


src/api/java/appeng/api/features/IInscriberRegistry.java
```diff
@@ -12,14 +13,15 @@
...
  * @since rv2
  */
 public interface IInscriberRegistry
 {
 	/**
-	 * Current list of registered recipes, you can modify this if you want too.
-	 * Will never contain a null recipe
+	 * An immutable copy of currently registered recipes.
+	 *
+	 * Use the provided methods to actually modify the inscriber recipes.
```
thatsIch: you should notice, that this is quote expensive, since upon each call an immutable copy will be created.

also link the "provided" methods with something like `@see´

yueh: Just the list itself is immutable, not the containing objects. So we are mostly speaking about creating a list with maybe 4-5 references each time, which is not that expensive.

The most impact it could have is on `ContainerInscriber.isValidForSlot()` as well as `TileInscriber.getTask()`. Which then is mostly an implementation issue there. Iterating over every possible list to check if cobble is a valid item for a printed circuit is probably way worse.
Potentially even refactor it and move it to the `IInscriberRegistry` itself. Basically a method to check if 3 inputs match an existing recipe. Which also make a more sophisticated check more feasible as it is located in a central location.

The normal `Collections` would also provide a immutable list, but this is just a view on top of the original one. So fully affected by any modification. So potential CME and similar things.

PR #1609, opened 2015-06-16
Splitted channeldata into two ints

### LINKED ISSUE #1510: Disabled Channels Setting invalidates the network on too many channels
collection entry

yueh:
Previously it did encode the current and previous used channels into the
same as well as mask it with 0xFF. Which lead to an overflow every 256
gridnodes requiring a channel. This will not happen at > 2^31

Also removes the need to bitshift them for every access.

`usedChannels` is only used by the debug tool and did report the last used channels not the current ones. For whatever reason. It did not mask it with 0xFF, but as it depends on the masked LSB of the int, it will also never exceed 256.
The debug tool should be changed to either use the already existing methods for the last or current channels. The question is mostly which one.

Fixes #1510

### DIFF
--- a/src/main/java/appeng/me/GridNode.java
+++ b/src/main/java/appeng/me/GridNode.java
@@ -71,7 +71,8 @@ public class GridNode implements IGridNode, IPathItem
 	private Object visitorIterationNumber = null;
 	// connection criteria
 	private int compressedData = 0;
-	private int channelData = 0;
+	private int usedChannels = 0;
+	private int lastUsedChannels = 0;
 
 	public GridNode( IGridBlock what )
 	{
@@ -90,7 +91,7 @@ public class GridNode implements IGridNode, IPathItem
 
 	public int usedChannels()
 	{
-		return this.channelData >> 8;
+		return this.lastUsedChannels;
 	}
 
 	public Class<? extends IGridHost> getMachineClass()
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/me/GridNode.java
```diff
@@ -603,7 +604,7 @@ public void setControllerRoute( IPathItem fast, boolean zeroOut )
 	{
 		if( zeroOut )
 		{
-			this.channelData &= ~0xff;
+			this.usedChannels &= ~0xffffffff;
```
yueh: This is actually quite stupid. Previously it just changed the LSB to zero, which was necessary. Now just assigning 0 makes way more sense.

thatsIch: yes

thatsIch: I just noticed, that every call to `setControllerRoute` is having `zeroOut = true`

yueh:
Updated. I did not remove the `zeroOut` to keep the commit as small as possible (to pick it into rv2).

It is unnecessary, but actually not causing (known) bugs, butI do not know the pathing code well enough to make any assumption about the calls being maybe wrong.

thatsIch:
right, sounds fine to me.

PR #1617, opened 2015-06-17
Enables the pattern terminal to fetch patterns from the network

### LINKED ISSUE #119: Pattern terminal convenience
Have pattern terminals auto stock themselves with any available blank patterns in the AE system. minor feature, sorry if it was suggested already.

Originally Posted by ClockworkGod

This issue was migrated; For original details see http://ae-mod.info/Tracker/1102/

yueh:
Line 205-225 are the important ones.

Otherwise just cleanup like making fields private or variables final.

Closes #119

### DIFF
--- a/src/main/java/appeng/container/implementations/ContainerPatternTerm.java
+++ b/src/main/java/appeng/container/implementations/ContainerPatternTerm.java
@@ -71,13 +71,13 @@ public class ContainerPatternTerm extends ContainerMEMonitorable implements IAEA
 {
 
 	public final PartPatternTerminal ct;
-	final AppEngInternalInventory cOut = new AppEngInternalInventory( null, 1 );
-	final IInventory crafting;
-	final SlotFakeCraftingMatrix[] craftingSlots = new SlotFakeCraftingMatrix[9];
-	final OptionalSlotFake[] outputSlots = new OptionalSlotFake[3];
-	final SlotPatternTerm craftSlot;
-	final SlotRestrictedInput patternSlotIN;
-	final SlotRestrictedInput patternSlotOUT;
+	private final AppEngInternalInventory cOut = new AppEngInternalInventory( null, 1 );
+	private final IInventory crafting;
+	private final SlotFakeCraftingMatrix[] craftingSlots = new SlotFakeCraftingMatrix[9];
+	private final OptionalSlotFake[] outputSlots = new OptionalSlotFake[3];
+	private final SlotPatternTerm craftSlot;
+	private final SlotRestrictedInput patternSlotIN;
+	private final SlotRestrictedInput patternSlotOUT;
 	@GuiSync( 97 )
 	public boolean craftingMode = true;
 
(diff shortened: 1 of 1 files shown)


thatsIch:
Just tried this, and I do not think that the implementation is intuitive. 

I expected either

* that a new pattern is stored in the pre-configured slot, upon using the last one
* that the slot is removed and the GUI is pulling directly from the network displaying something like a counter of the ones stored in the network


Currently is it implemented, that it is using a new pattern from the network, upon process click (also this is not consistent with the crafting behavior, since the pattern does not refill the slots **after** they are empty, thus the recipe being invalid)

yueh:
Some better indication on how it works would certainly be an improvement. But the available space is pretty limited.

The first option with fetching a new pattern once the last is used still requires some manual work. Basically if the network runs out, the player needs to manually fetch a single pattern from the network and place it there. Or we need to handle it multiple, like on used last patter, pattern enters the network and pattern terminal is open or open the terminal.
Also it would require that same patterns are scatter over all existing terminals and not usable.

The second might be an idea. But it can break too much. For example players will not longer be able to access them once updated.
* Like under which conditions should the slot be disabled?
  Always when the terminal has access to a network with storage? Which also needs security and non 
extractable ones to be considered
* Once the network has some patterns in it? But what happens if the network is slowly refilling them and the player is continously creating new ones? Should the slot then start to flicker? Or what should happen, if someone crafts 2 stacks manually, puts the first stack into the slot and the rest into the terminal? Seeing the slot vanish with a full stack is probably not ideal.

It might work to show a greyed out icon of the network amount once the slot is empty.
But no idea how easy that is.

The pattern terminal itself is also pretty inconsistent. For example it does make a recipe invalid once there are no longer enough materials inside the network. In this case it would not be even possible to create patterns without having the exact amount of needed items and not just one like currently.
Also using it to craft will not always craft, at first it tries to extract the output from the network and only afterwards try to actually craft it. (And then still not make the recipe invalid without enough)

thatsIch:
I also have a third option to offer:

* Using patterns from network as first priority, before using the internal slot storage.

yueh:
Not sure about the network as first priority, it will not be that obvious what is happening and players will just yell "omg. it's duping patterns" :worried: 

Also completely removing it will also remove a feature. The terminal will happily encode patterns without any network, but still considers the player inventory. I am quite frequently using this to test setup, for example assemblers with patterns. Being forced to build a full network before would be a bit tedious.
Otherwise the implementation would be not that difficult. Terminals are already monitoring the network, so we could just check for patterns and without them disable the button and add a tooltip about missing patterns inside the network. Maybe also needs an indication when extracting them fails (security related issues)

I think the best option in terms of consistency, usability and backward compatibility would be (and without having some migration code laying around forever to say dump the old stacks into the network including dealing with all issues in regards to an inaccessible network)
* Keep the slot
* Prioritize the slot over the network
* Add a "fake item" icon with the amount of the network when empty + a tooltip stating they will be used from there
* Still allow patterns to be place into the slot once it's empty
* Also needs some feedback in regards of security stations.

thatsIch:
Suggestion sounds fine

thatsIch:
We could open up a poll to see which behaviour is desired. Maybe we are over-engineering it

thatsIch:
Requires rebase

PR #1624, opened 2015-06-20
You can now automate Vibration Chambers

TheCricket:
When attempting to automate with a lava bucket you get the bucket stuck in there, You can now automate the system using lava.

### DIFF
--- a/src/main/java/appeng/tile/misc/TileVibrationChamber.java
+++ b/src/main/java/appeng/tile/misc/TileVibrationChamber.java
@@ -21,6 +21,7 @@ package appeng.tile.misc;
 
 import io.netty.buffer.ByteBuf;
 
+import net.minecraft.init.Items;
 import net.minecraft.inventory.IInventory;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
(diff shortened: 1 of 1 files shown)


TheCricket:
:dancers:

yueh:
I am not really convinved, that this is a good chance.

First of all it will break consistency with vanilla furnaces, which do not allow fuel to be extracted.

Secondly it will break certain builds, these might not be used that much (if at all). But there is still the chance that someone uses it. For example if someone uses storage buses to keep them stacked with fuel and wants to be able to monitor it, but not allow the network to ever pull fuel from it.

FireBall1725:
I think if it was changed to only return true if the itemstack is an empty bucket would be fine.

TheCricket:
Ok I can do that :)

yueh:
Please squash the commits.

TheCricket:
so much easier to do this https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/1633

MyDeloreanWontStart:
Doesn't Extracells have a fluid vibration chamber?

TheCricket:
this isn't extracells is it

PR #1634, opened 2015-06-21
Fixes #1588: Charged Certus Quartz Ore was given a wrong name

### LINKED ISSUE #1588: Charged Certus Quartz Ore has same name as Certus Quartz Ore
Minecraft: 1.7.1
JABBA,1.2.1
AppliedEnergetics2 (rv2-beta-18)
Infinity 1.7 Mod Pack

OreDictionary Dump via NEI shows:
oreCertusQuartz,1xtile.appliedenergistics2.OreQuartz@0,appliedenergistics2:tile.OreQuartz,Certus Quartz Ore,false
oreCertusQuartz,1xtile.appliedenergistics2.OreQuartzCharged@0,appliedenergistics2:tile.OreQuartzCharged,Charged Certus Quartz Ore,false

Because of the same item name (oreCertusQuartz), sorting with other mods such as JABBA does not occur correctly.

thatsIch:


yueh:
As far is I remember, both shared to same oredict entry to be interchangable in some recipes.
Would this not break it?

thatsIch:
I thought this is actually the issue, **that** both share the same because they are actually not really interchangeable? (since different behaviour -> spawn charged certus quartz)

yueh:
I always considered them as interchangeable for most recipes. The charged ones are mostly to gate the progression, once found there are mostly equal. But without introducing another item which afterwards becomes useless or enable a situation where the player gets stuck  with only charged certus available and no way back.

thatsIch:
You can actually give them multiple ore dictionary entries

yueh:
Needs an update.

thatsIch:
not really required anymore, since the line is missing anyways.

thatsIch:
nvm, forgot how the new layout was

thatsIch:
updated

thatsIch:
Waiting for #1805

PR #1687, opened 2015-07-06
Fixes #1686: Let PlayerRegistry handle nonexistent players

### LINKED ISSUE #1686: Security station crashing on blank card
A security station will crash with a `NPE` on inserting a blank card. These use `null` when not assigned, but the `PlayerData` requires it to be not null.

yueh:
The `PlayerRegistry` now handles the case for null as player used for blank cards instead of passing it to the `PlayerData`, which should only handle existing players.


yueh:
Updated.

`PlayerData` will now only handle a valid `GameProfile` and not longer the special case for invalid or non existent ones as its job is to handle the saving and resolving the data.
`PlayerRegistry` is now handling these cases before forwarding the data

Issue #1010, opened 2015-03-11
AE network stops importing


henriksjostrom:
For some reason my AE network stops importing from my ender chest. It has storage and power so that is not the issue.

When the server starts it imports for a little while and then stops. This could be related to the fact that my ender chest has a rather massive item inflow, essentially everything I want to put into the network comes through it. It gets filled up almost instantly when the network stops. 

Is there some other requirement that I should be aware of when it comes to importing items?

thatsIch:
should not besides enough storage space
another method would be to use tesseracts and ME Interface

henriksjostrom:
Yes, having some pipes from the chest into an interface made it work again. Why the import buses fail I still have no clue.

BoboSims:
You might just be asking too much of that import bus. Maybe you could divide your incoming resource stream between two or more chests that each get processed by their own bus.

henriksjostrom:
It is not a single import bus, it is 12 with full acceleration upgrades and they did stop completely.

BoboSims:
Ah, your top post speaks of "ender chest", not "ender chest**s**", so I assumed 'one chest, probably one bus'. I still have trouble imagining 12 buses accessing one chest, though. So maybe there are more chests, too?

henriksjostrom:
Well I had 1 chest with 6 import buses, the issue arose so I upgraded by adding another 6. Did not help

thatsIch:
Are you by chance using an outdated version? There was a bug, when they had 4 upgrades installed, that they fail

Cisien:
Or maybe running out of power? Try slapping a dense energy cell on that network and see if the issue is still there

henriksjostrom:
I am using rv2-beta-8 (FTB infinity 1.2.1).

I had 230 dense energy cells in the network at last count storing 370 MAE and the network only draws roughly 10kAE. Power it should not be.

However I have noticed a few more things, when crafting on occasion it seems to not arrive at any crafting plan, nothing appears in the UI and I can only cancel. Trying a few seconds later generally solves this.

Furthermore the network can suddenly stop exporting to.

This might be crafting related as the network stopped exporting once I started a bunch of crafting jobs (crafting interfaces with storage interface on to make sure a certain amount of items are always available)

BoboSims:
To *me* that sounds like a processing/storage issue: too many tasks and not enough processors and/or storage units. The network is too busy solving the crafting issues, and no longer gets around to moving objects.

[...]

PR #1696, opened 2015-07-10
Adds a round robin and random mode to export buses.

### LINKED ISSUE #85: Round Robin upgrade for export buses
Pretty simple - If I have an export bus that has more than one item programmed into it, allow me to put it in 'round robin' mode (through upgrade or some other switch). This means, after inserting one of the programmed types, *if* the next programmed type is available in the network inventory, insert it next, followed by repeating for each type programmed in the export bus.

The use case is machines that make different things based on what you give them. Let's say I set up an ad-hoc network connected to something that can make widgets, woozles and wombats based on input of diamond, gold or iron - I'd like it to make these things consecutively, instead of making all the widgets, then woozles and finally wombats. 

Things that come to mind:

 - Sag mill
 - Molecular Transformer
 - AE inscribers (to some extent)
 - Pre-heated induction furnaces, canning machines, ore washers from IC2

I can't find a good use for this for import buses, or storage buses set on extraction only. First, unless it's an actual inventory, you can only take whatever is ready to take. 

If I am missing some other obvious way of doing this, I apologize, but I can't  seem to find any way of accomplishing this with any of the available upgrades.

Originally Posted by tinkertim

This issue was migrated; For original details see http://ae-mod.info/Tracker/0911/

yueh:
Adds a round robin and random export mode to export buses.

Setting will only be shown with at least 1 capacity card (else there is only 1 slot), but not requires any additional cards.

Something like a scheduling card might be an idea for the future, but currently it has a really limited use and would compete with the other cards for the export bus as well as also requiring at least an additional capacity card to work.

Resolves #85


yueh:
Updated and included some cleanup for the shared superclass of both import and export bus as well as some cleanup for the import bus.

Adds an additional check to determine, if the chunk of the target inventory is actually chunkloaded.
If this fails, the (import) bus is no longer put to sleep but just set to idle, so it can resume its operation once the chunk is loaded without needing a block update. This also means that the export bus is now put to sleep when not facing an inventory and will resume once one is placed.
This should fix one case of #1010

PR #1730, opened 2015-07-21
Fixes #1728 Mark the entity as dead before spawning overflow

### LINKED ISSUE #1728: Annihilation Plane destroys items, doesn't store them.
I'm using rv2-stable-10 with channels disabled in the config.

I have crystal growth accelerators set up with an annihilation plane and a storage bus on a subnetwork at the base. The storage bus is configure to only store pure crystals and fluix crystals.  Placing items in the water via any method causes the annihilation plane to instantly destroy any item and nothing is stored in the network, nor in any interface on the network.  The items are simply deleted.

yueh:
Fixes #1728

### DIFF
--- a/src/main/java/appeng/parts/automation/PartAnnihilationPlane.java
+++ b/src/main/java/appeng/parts/automation/PartAnnihilationPlane.java
@@ -274,8 +274,12 @@ public class PartAnnihilationPlane extends PartBasicState implements IGridTickab
 
 			if( capture )
 			{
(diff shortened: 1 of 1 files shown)


yueh:
Updated.
It will no longer respawn the overflow as new entity, instead it will now update the stacksize.
So it will no longer respawn the overflow as new entity and thus prevent exactly one stack from despawning forever.

Also a small fix to no longer player the animation without any actual imported items.

PR #1732, opened 2015-07-22
Use the correct resolver for crystal seeds

### LINKED ISSUE #1718: AE2 Seeds Problem with GregTech
I already posted the issue at Blood-Asp/GT5-Unofficial#91 but since it seems GregTech is not directly involved in it, I want to mention it here again:
When crafting Fluix or Netherquarz seeds they are displayed and recognized as a different item then the crafting logic expects and therefore the autocrafting logic breaks. Tested with only ae2-rv2-stable9 and gt-5.08.24 and IC2-2.2.746 and default configs.
As seen in the other issue thread, the crafting recipe doesn't have a nbt tag "progress" but for whatever reason the crafted item has (immediately). Certus Quarz Seeds apparently don't have this issue. 
If I can provide further Information please let me know!
ae2bug

yueh:
Use the Resolver for any seed and not just for certus crystals.

**This will break already existing patterns**

Fixes #1718

### DIFF
--- a/src/main/java/appeng/recipes/AEItemResolver.java
+++ b/src/main/java/appeng/recipes/AEItemResolver.java
@@ -109,11 +109,11 @@ public class AEItemResolver implements ISubItemResolver
 				}
 				if( itemName.equalsIgnoreCase( "ItemCrystalSeed.Nether" ) )
 				{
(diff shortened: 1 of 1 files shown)


yueh:
Test this please.

Issue #1552, opened 2015-06-05
"Shaped IC2 Crafting" not pulling items from storage buses when shift clicking recipe from NEI


016Nojr:
When I attempt to shift click a "Shaped IC2 Crafting" recipe from NEI it only places the IndustrialCraft 2 items in the crafting "window", nothing else.

Mod versions:
appliedenergistics2-rv2-stable-4
CodeChickenCore-1.7.10-1.0.6.43-universal
industrialcraft-2-2.2.730-experimental
Jabba-1.2.1a_1.7.10
NotEnoughItems-1.7.10-1.0.4.107-universal

![2015-06-05_19 31 11](https://cloud.githubusercontent.com/assets/8192699/8011552/daf1759e-0bb9-11e5-8098-8e2246e1df9a.png)
![2015-06-05_19 31 07](https://cloud.githubusercontent.com/assets/8192699/8011550/daed75f2-0bb9-11e5-8037-72c6c558b6fd.png)
![2015-06-05_19 31 21](https://cloud.githubusercontent.com/assets/8192699/8011551/daefd1e4-0bb9-11e5-8807-a17d40f51f49.png)

TheJulianJES:
@016Nojr Could you please try this in rv2 stable 5?

016Nojr:
Repeated same process in rv2 stable 5, not working.

oks1998:
rv2 stable 6 also not working.....and when auto crafting the solar distiller using ae2, it show me in the chat there: crafting tree construction failed

016Nojr:
That is happening to me with "Transformer Upgrades" aswell @oks1998

oks1998:
but if u create a pattern for the recipe it can craft it...it just cannot shift click the ? in nei and put all the stuff automatic on the crafting table

thatsIch:
probably like #1581

yueh:
It is pretty similar, but this is something we could fix on our side (without breaking stuff).

As everyone knows, using glass or iron ingots to hit zombies will slowly damage these items like any other weapon, thus the IC2 recipes are not asking for the undamaged item, but are using the Oredict wildcard to match any damaged itemstack of these undamagable items.

We could probably change the NEI handler to accept the wildcard on undamageable items.
I could imagine that one mod adds some damagable item with a matching oredict name and we might miss these then. But I currently cannot think of any mod doing it, to actually test it.

PR #1738, opened 2015-07-25
Fixes #1581: Uses 32767 as wildcard meta like IC2

### LINKED ISSUE #1581: Validate Item Metadata before entering system or notify user of invalid items in the system
can put it into the ae2 system but cannot take it out 
(shift click also cannot, right click also cannot, left click also cannot)
only can taken out by give 1 that item to me and put it into the crafting table in ae2 and take out the rest

Driphter:
Whenever an ingredient in `...\assets\ic2\config\shaped_recipes.ini` and `...\assets\ic2\config\shapless_recipes.ini` ends in a `@*` IC2 give the ingredient a meta of 32767. When IC2 compares `ItemStack`s it uses 32767 as a meta wildcard.

### DIFF
--- a/src/main/java/appeng/util/Platform.java
+++ b/src/main/java/appeng/util/Platform.java
@@ -1867,6 +1867,13 @@ public class Platform
 			{
 				return true;
 			}
+
+			// IC2 uses 32767 as a wildcard meta in its recipes
+			if( that.getItemDamage() == 32767 || other.getItemDamage() == 32767 )
+			{
+				return true;
+			}
+
 			return that.getItemDamage() == other.getItemDamage();
 		}
 		return false;


yueh:
This is a potentially fix for #1552 not #1581 (which is simply broken on their side).

The main issue here is the method is named `isSameItemType` not `isEqualItemType` and some callers require it to be the same type not just an equal. So changing this can potentially cause new bugs.

Also hardcoding the wildcard value is really bad, as this has already changed in the past.

Driphter:
My mistake on the issue number. I'm not sure how I screwed that up.

Other than that, I agree with you on both points. Would it be reasonable to only compare the metadata of items with `hasSubtypes`? [This is currently how NEI does it.](https://github.com/Chicken-Bones/CodeChickenLib/blob/master/src/codechicken/lib/inventory/InventoryUtils.java#L278)

yueh:
No issue with the number, just in case it gets mixed up somehow.

`hasSubtypes` might be an idea, but I cannot exclude any potientially issues with it just out of my head.
One of the actual lines affected by it is [Platform.java#L1835](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/3a9296f453c986c2dfc2de9133172f1a254bff4a/src/main/java/appeng/util/Platform.java#L1835), so it could also something to be handled in `sameOre()`.
Basically there are a numbers of potential fixes for it. There are even ways to handle it in the first try without needing the `checkFuzzy` branch, but these are really deep changes and I just have not enough time latey to wrap my head around it and which issues can appear from it.
When I need to choose between an existing (small) inconvenience and potentially introducing a dupe bug, I still favor the first option.

Driphter:
Perhaps we could be super-specific and have a check for `hasSubtypes`, `damage == 32767`, and `!isItemStackDamageable` when `output.getClass().getCanonicalName() == "ic2.core.item.ItemIC2"`.

Somewhat fragile but still specific enough to likely be safe. If the class name changes it will just stop working like it is right now. I know it introduces a bit of IC2 jank to the code, but at least it can be separated enough to be understandable.

yueh:
If we have to introduce mod specific code to work around their behaviour, I would honestly just say just ignore it.

Most likely not needed for this case. I just have no idea currently about what the best option is.

PR #1770, opened 2015-08-06
Update zh_CN.lang

bakaxyf:


yueh:
As a small hint for future translations. I would not consider the strings to be final during alpha, so it might be an idea to wait for the beta and save some of your time.

bakaxyf:
Thanks

PR #1772, opened 2015-08-06
Feature 1569 Typesafe Integration Modules

yueh:
`@Interface` will now require an `IntegrationType` as `iname` and not a string, thus not longer breaking on internal refactoring.

It might also be an idea to change `iface` to `Class<?>` in a further iteration and handle the potentially thrown exception, this would also allow to already fail at compile time in case of a nonexistent class.

Also moved the registration of layers from the registration into the appropriate `IIntegrationModule`.
As these are called during init and the final class is only assembled during postInit, this will not cause any issue, but better encapsulate it.
`IPartHelper.registerNewLayer` should potentially be changed to accept ´Class<?>´ and not `String`, but this would be a API breaking change.
To ensure typesafety it is possible to pass `Object.class.getName()` for now, the check for being enabled is already based on the existance of these classes and thus will catch any potential exception.

Moved `BaseModule.testClassExistence()` into a static helper as some integrations cannot extend `BaseModule` directly, for example the craftguide integration. This still allows to protect these ones in case of a breaking change of the related mod.
Also moved all checks into the constructor for a consistent approach in all integration modules.

The ASM integration is a bit messy as the annotations are now parsed as `String[]` and with the first element being the enum class and the second the acutal value.
Something similar will apply, when changing `iface` to `Class<?>`, but that is just messy ASM.

### DIFF
--- a/src/main/java/appeng/integration/BaseModule.java
+++ b/src/main/java/appeng/integration/BaseModule.java
@@ -21,12 +21,6 @@ package appeng.integration;
 
 public abstract class BaseModule implements IIntegrationModule
 {
-
-	protected void testClassExistence( Class<?> clz )
-	{
-		clz.isInstance( this );
-	}
-
 	@Override
 	public abstract void init() throws Throwable;
 
--- a/src/main/java/appeng/integration/IntegrationHelper.java
+++ b/src/main/java/appeng/integration/IntegrationHelper.java
@@ -1,6 +1,6 @@
 /*
  * This file is part of Applied Energistics 2.
- * Copyright (c) 2013 - 2014, AlgorithmX2, All rights reserved.
+ * Copyright (c) 2013 - 2015, AlgorithmX2, All rights reserved.
  *
  * Applied Energistics 2 is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
@@ -16,32 +16,14 @@
  * along with Applied Energistics 2.  If not, see <http://www.gnu.org/licenses/lgpl>.
  */
 
-package appeng.integration.modules;
+package appeng.integration;
 
 
-import appeng.helpers.Reflected;
-import appeng.integration.BaseModule;
-import appeng.integration.IntegrationHelper;
-
-
-public class RFItem extends BaseModule
+public class IntegrationHelper
 {
-	@Reflected
-	public static RFItem instance;
 
-	@Reflected
-	public RFItem()
+	public static void testClassExistence( Object o, Class<?> clz )
 	{
-		IntegrationHelper.testClassExistence( this, cofh.api.energy.IEnergyContainerItem.class );
-	}
-
-	@Override
-	public void init()
-	{
-	}
-
-	@Override
-	public void postInit()
-	{
+		clz.isInstance( o );
 	}
 }
--- a/src/main/java/appeng/integration/modules/CLApi.java
+++ b/src/main/java/appeng/integration/modules/CLApi.java
@@ -20,19 +20,26 @@ package appeng.integration.modules;
 
 
 import appeng.api.util.AEColor;
+import appeng.helpers.Reflected;
 import appeng.integration.BaseModule;
+import appeng.integration.IntegrationHelper;
 import appeng.integration.abstraction.ICLApi;
 
 
 public class CLApi extends BaseModule implements ICLApi
 {
-
+	@Reflected
 	public static CLApi instance;
+
+	@Reflected
+	public CLApi()
+	{
+		IntegrationHelper.testClassExistence( this, coloredlightscore.src.api.CLApi.class );
+	}
 
 	@Override
 	public void init() throws Throwable
 	{
-		this.testClassExistence( coloredlightscore.src.api.CLApi.class );
 	}
 
 	@Override
--- a/src/main/java/appeng/integration/modules/DSU.java
+++ b/src/main/java/appeng/integration/modules/DSU.java
@@ -25,7 +25,9 @@ import powercrystals.minefactoryreloaded.api.IDeepStorageUnit;
 
 import appeng.api.AEApi;
 import appeng.api.storage.IMEInventory;
+import appeng.helpers.Reflected;
 import appeng.integration.BaseModule;
+import appeng.integration.IntegrationHelper;
 import appeng.integration.abstraction.IDSU;
 import appeng.integration.modules.helpers.MFRDSUHandler;
 import appeng.integration.modules.helpers.MinefactoryReloadedDeepStorageUnit;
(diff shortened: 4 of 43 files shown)


src/main/java/appeng/integration/BaseModule.java
```diff
@@ -21,12 +21,6 @@
...
 public abstract class BaseModule implements IIntegrationModule
 {
-
-	protected void testClassExistence( Class<?> clz )
-	{
-		clz.isInstance( this );
-	}
-
 	@Override
 	public abstract void init() throws Throwable;
```
thatsIch: Any reason to keep them in a standalone module and not just put them into `IIntegrationModule`?

thatsIch:
Pointers:

* Maybe keep consistent of import vs full package path of class to test for existence (partially mixed)
* For the `registerNewLayer` you can introduce a class variant and `@deprecate` the other one
* comment (s.somewhere)

src/main/java/appeng/integration/BaseModule.java
yueh: The craftguide integration needs to extend `CraftGuideAPIObject` and thus extending `BaseModule` is simply not possible (except we would switch to scala).

We could move it into the `IIntegrationModule` interface, but then there is a chance that it will be missed when refactoring it later. Keeping it in a helper class will reduce this possibility.

I am also not positive about the current `testClassExistence` implemenation as is uses an reflective `instanceof`. `Class.forName(clz.getName())` might be a more fitting implementation and remove the need to pass this around. It really just needs to throw an exception.
Maybe `clz.getName()` could already throw one.

yueh:
Still ongoing currently.

1. Should really be consistent. For the sake of completeness I would favour a fqn.
2. There are a bunch of reflection uses all around, which could be removed. Like letting the `AEBaseBlock.getRenderer` return a renderer instance and not a `Class` and then call `.newInstance()`. It is really used as factory method.
3. See first, bit wasted to write documentation, when it can still change.

src/main/java/appeng/integration/BaseModule.java
thatsIch: I meant the `init()` and such methods, since they are defined as `public abstract`

The implementation is actually a no-brainer, since it is only used, if the definitions are existent. The semantic is non-existent, since the return value does not matter. You could actually remove the pointer to `this` everywhere and use some instance, like `clz.isInstance( clz )` also possible. We could make it explicit by defining the constructors to throw `NoClassBlubBla` exception or some some actual method like `validate() throws NoClassDefFoundException` since validation is not meant to be in the constructor(imo?)

thatsIch:
1. sounds fine
2. yes, *cries* 
3. not meant comments as in documentation, but a pointer to the comment I posted xD

thatsIch:
also squash

yueh:
Updated, squashed, changed everything to fqn and added some potentially missed checks.

yueh:
Test this please.

src/main/java/appeng/integration/modules/RC.java
```diff
@@ -49,13 +52,10 @@ public void rockCrusher( ItemStack input, ItemStack output )
 	@Override
 	public void init()
```
orod-org: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1186)

src/main/java/appeng/integration/modules/Waila.java
```diff
@@ -51,15 +64,11 @@ public static void register( IWailaRegistrar registrar )
...
 	public void init() throws Throwable
 	{
-		this.testClassExistence( IWailaDataProvider.class );
-		this.testClassExistence( IWailaRegistrar.class );
-
 		FMLInterModComms.sendMessage( "Waila", "register", this.getClass().getName() + ".register" );
 	}
 
 	@Override
 	public void postInit()
```
orod-org: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1186)

src/main/java/appeng/integration/modules/CLApi.java
```diff
@@ -20,19 +20,26 @@
...
 	public static CLApi instance;
 
+	@Reflected
+	public CLApi()
+	{
+		IntegrationHelper.testClassExistence( this, coloredlightscore.src.api.CLApi.class );
+	}
+
 	@Override
 	public void init() throws Throwable
```
orod-org: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1186)

src/main/java/appeng/integration/IntegrationHelper.java
```diff
@@ -0,0 +1,29 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Applied Energistics 2.  If not, see <http://www.gnu.org/licenses/lgpl>.
+ */
+
+package appeng.integration;
+
+
+public class IntegrationHelper
```
orod-org: MAJOR Add a private constructor to hide the implicit public one. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1118)

src/main/java/appeng/integration/modules/Mekanism.java
```diff
@@ -24,25 +24,31 @@
...
 	public static Mekanism instance;
 
+	@Reflected
+	public Mekanism()
+	{
+		IntegrationHelper.testClassExistence( this, mekanism.api.energy.IStrictEnergyAcceptor.class );
+	}
+
 	@Override
 	public void init() throws Throwable
```
orod-org: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1186)

src/main/java/appeng/integration/modules/DSU.java
```diff
@@ -50,7 +59,6 @@ public boolean isDSU( TileEntity te )
 	@Override
 	public void init()
```
orod-org: MAJOR Add a nested comment explaining why this method is empty, throw an UnsupportedOperationException or complete the implementation. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1186)

orod-org:
SonarQube analysis reported 8 issues:
* MAJOR 6 major
* MINOR 2 minor

* MINOR [Remove this unused import 'appeng.integration.IntegrationRegistry'.](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/f303c17ae0451d48ae6d1b20befb60f09ef5f90a/src/main/java/appeng/core/Registration.java#L79) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUselessImportCheck)
* MINOR [Remove this unused import 'appeng.integration.IntegrationType'.](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/f303c17ae0451d48ae6d1b20befb60f09ef5f90a/src/main/java/appeng/core/Registration.java#L80) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AUselessImportCheck)

Issue #397, opened 2014-11-06
Seed "Errors" (Too many meteors)


theriftwalker:
Apparently I'm being suggested to start AE. http://puu.sh/cFfQL/d26046eccd.jpg  http://puu.sh/cFfWf/7cca9d1d8c.jpg  http://puu.sh/cFg1B/755d4707d8.jpg (I'd hate to be that guys barber)  Seed is : 113963865144355269 with BoP installed, but not enabled.

thatsIch:
Somebody per chance still encountering this?

moto1239:
2016 and I'm experiencing it with a new world.
Running Feed The Beast Infinity Evolved v2.3.5

![2016-01-21_14 07 09](https://cloud.githubusercontent.com/assets/16826578/12496098/78c4538e-c048-11e5-9c0f-5b4ed677a4f2.png)
![2016-01-21_13 59 06](https://cloud.githubusercontent.com/assets/16826578/12496111/81166ad6-c048-11e5-9bf1-7eb92d903036.png)
First time using GitHub \o/
Let me know if more info/pics/world file is needed

wlhlm:
What are the coordinates and what is the seed? Are you running modpack unmodified?

moto1239:
Seed is -3468532045546355263 coordinates are around x: -550, z:800
iirc I haven't modified the pack, just updated it as a matter of fact.

thatsIch:
this is bullshit

thatsIch:
Is this on a server  @moto1239

moto1239:
No

Digitalsabre:
So, apparently I'm not the only one getting what's ostensibly entirely meteorite-based biomes...

gamecompiler:
Could it be that Biomes o plenty has a problem with mod rv2 ? Appearantly that "Bug" do not appear when using the default WorldGen.

I try to figure out why BOP-1.7.10-2.1.0.1394 and AE2-rv3-beta-1 work perfectly, but BOP... with AE2-rv2 not. 

So the "resolution" for this problem is to update to rv3-beta-1.

Sxw1212:
I experienced this bug without BOP. I had rv2, but I don't know if rv3 fixes everything or not.

mindforger:
why would you not update to the latest rv3.beta build 6 instead on rv3.beta build 1 ?!

thatsIch:
[...]

PR #1776, opened 2015-08-06
Closes #1283: Add custom Callable to prevent memory leaks

### LINKED ISSUE #1283: TickHandler / Callable memory leak

yueh:
Added `IWorldCallable` with `World` as parameter to replace the generic `Callable`.

Besides `BlockUpdate` and `MeteoriteSpawn` no other class actually needs to access the world passed as argument (or it is already available through being a `TileEntity`).

Closes #1283

### DIFF
--- a/src/main/java/appeng/hooks/TickHandler.java
+++ b/src/main/java/appeng/hooks/TickHandler.java
@@ -25,7 +25,6 @@ import java.util.Iterator;
 import java.util.LinkedList;
 import java.util.Queue;
 import java.util.WeakHashMap;
-import java.util.concurrent.Callable;
 import java.util.concurrent.TimeUnit;
 
 import com.google.common.base.Stopwatch;
@@ -55,6 +54,7 @@ import appeng.entity.EntityFloatingItem;
 import appeng.me.Grid;
 import appeng.me.NetworkList;
 import appeng.tile.AEBaseTile;
+import appeng.util.IWorldCallable;
 import appeng.util.Platform;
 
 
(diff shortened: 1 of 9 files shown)


yueh:
Might also be related to #397

src/main/java/appeng/hooks/TickHandler.java
```diff
@@ -88,11 +88,11 @@ public void addCallable( World w, Callable c )
 		}
 		else
 		{
-			Queue<Callable> queue = this.callQueue.get( w );
+			Queue<IWorldCallable<?>> queue = this.callQueue.get( w );
 
 			if( queue == null )
 			{
-				this.callQueue.put( w, queue = new LinkedList<Callable>() );
+				this.callQueue.put( w, queue = new LinkedList<IWorldCallable<?>>() );
```
thatsIch: maybe split in two lines

src/main/java/appeng/util/IWorldCallable.java
```diff
@@ -0,0 +1,49 @@
...
+	/**
+	 * Similar to {@link Callable#call()}
+	 *
+	 * @see Callable#call()
+	 * @param world
+	 * @return
+	 * @throws Exception
+	 */
+	@Nullable
+	T call( @Nullable World world ) throws Exception;
```
thatsIch: if the `world` is `nullable` you need to check them all over the place, else `NPE`s?

thatsIch:
reviewed, left 2 comments

src/main/java/appeng/hooks/TickHandler.java
yueh: Changed

src/main/java/appeng/util/IWorldCallable.java
yueh: There are actually just two `IWorldCallable` using the world (meteorspawn/blockupdates).
All in all there are 3 different types
1. Called inside the world specific queue and needing a world
2. Called inside the world specific queue (world can be null, but is always available)
3. Called inside the server queue (world = null, when adding the callback)

We could split 3. off, but currently the server queue is only used by the spatial transfer.

thatsIch:
Just wondering, did you removed the callable entirely? cause in some cases the call argument is not even used meaningfully.

yueh:
It is either that as all are currently passed through `TickHandler.addCallable()` or managing different types of callback queues per world.

PR #1787, opened 2015-08-10
Fixes #1786: Locale critical code now uses the english local. Fixes Turkish Problem.

### LINKED ISSUE #1786: Turkish locale convert IMCs
When using `toLowerCase` with Turkish Locale used, it will not convert the Strings as expected from an English Locale perspective, because it respects the `i18n` standard for localization.

[Diskussion Link](https://www.reddit.com/r/feedthebeast/comments/3gdxk1/psa_for_developers_turkish_locale_breaks_stuff/ctxhta2)

A fix would be to use an invariant locale like `new Locale(Empty, Empty, Empty)` or use the English locale as default I guess

thatsIch:
Applied English Locale where localization is not expected as in internal recipe handling and IMC handling, basically which interacts with public API where we either require to enforce the incoming text with regex ([a-z0-9]) or just expect proper usage of the API, but with just using upper cases in recipe files it would break in Turkish Locale like

ALIAS

another option would have been to use `equalsIgnoreCase` in some cases, but not all applicable

### DIFF
--- a/src/api/java/appeng/api/features/IRecipeHandlerRegistry.java
+++ b/src/api/java/appeng/api/features/IRecipeHandlerRegistry.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License (MIT)
  *
- * Copyright (c) 2013 AlgorithmX2
+ * Copyright (c) 2013 - 2015 AlgorithmX2
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy of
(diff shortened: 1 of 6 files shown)


yueh:
The reddit post also suggest using `Locale.ROOT` to have a neutral Locale and not have to deal with potential cases specific to english.

thatsIch:
This is to not use `LOCALE.US` which is just US, not normal English locale, but we are def using English in the API

PR #1805, opened 2015-08-15
Reorganizes the recipes in a more reasonable way.

### LINKED ISSUE #1791: Some recipes are not working with rv3
Some recipes are currently not working, for example this applies to the `ME Terminal`, `Network Tool` and nether quartz tools (but not certus ones)

I currently cannot see any pattern or issue causing it, but it seems to be related to oredict or group ingredients. For example the recipes for the quartz tools are identical, just use different oredict entries (for certus/nether quartz)

@thatsIch Could you maybe look into it? You should be a bit more familiar with the recipe handling.

yueh:
The recipes are now structured into multipe subfolder and split into more
distinct files, so the names are more appropriate and are better at
hinting which items the actually contain.

It also extends the RecipeResourceCopier to now handle the folder
recursively and extract all subdirectories and their files.

"import=" is currently requiring a relative path to the root directory
of the recipes. This would require a larger rewrite/refactoring, thus it
is kept for now until a potentially later changer.

This reverts splitting the oredict entries into their own directory and
moves them back into the recipes folder, as it currently is causing a
couple of issues like not being able to resolve the aliases or is not
working indev. But to keep it seperate it is now its own recipe file.

Fixes #1791
Reverts #1635


thatsIch:
* `misc.recipe` in `misc` folder is somewhat badly named
* `cables/misc.recipe` maybe into `parts`?

yueh:
I am not really happy with `misc` in general. It is really the place for leftover/uncategorized stuff.

I was thinking about mirroring the API with blocks/parts/materials/etc. But this is also not really fitting.

thatsIch:
the general layout looks fine to me,

* `misc.recipe` is just containing TNT so maybe just use it 
* `cables/misc.recipes` are anchors and such, which are technically no cables

yueh:
Small update to the `build.gradle` to no longer include the unused `oredict` folder.

thatsIch:
makes sense

thatsIch:
I noticed that some recipes were not working looking at the console

yueh:
Any specific ones? I went through all items ingame (hopefully not missed any) and there were no issues.

Also do not forget, that even before some recipes did fail to resolve in indev, as for example the wooden gear is disabled as BC already provides one. If I remember correclty, it were always 2 or 3 recipes failing due to having some other mods indev.

thatsIch:
oh right.. need better messages

but some did contain typos I think thus failing to resolve.

thatsIch:
posted in IRC

PR #1823, opened 2015-08-19
Fixes #1810: Removes a CompassReader once the world is unloaded.

### LINKED ISSUE #1810: CompassService memory leak
Apparently appeng.services.CompassService (temporarily?) leaks some (~5?) invalid world references through worldSet.

yueh:
This should no longer keep a reference to a World around and potentially
keep them loaded.
Also added a finalize() to CompassRegion to ensure the file is closed on a
GC.

Some cleanup regarding member order, final, etc
Fixes #1810 (should)

### DIFF
--- a/src/main/java/appeng/services/CompassService.java
+++ b/src/main/java/appeng/services/CompassService.java
@@ -35,11 +35,15 @@ import com.google.common.base.Preconditions;
 import net.minecraft.block.Block;
 import net.minecraft.world.World;
 import net.minecraft.world.chunk.Chunk;
+import net.minecraftforge.event.world.WorldEvent;
+
+import cpw.mods.fml.common.eventhandler.SubscribeEvent;
 
 import appeng.api.AEApi;
 import appeng.api.util.DimensionalCoord;
 import appeng.services.compass.CompassReader;
 import appeng.services.compass.ICompassCallback;
+import appeng.util.Platform;
 
 
 public final class CompassService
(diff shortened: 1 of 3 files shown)


src/main/java/appeng/services/CompassService.java
```diff
@@ -71,6 +75,22 @@ public CompassService( @Nonnull final File worldCompassFolder, @Nonnull final Th
...
+	 * Ensure the a compass service is removed once a world gets unloaded by forge.
+	 *
+	 * @param event the event containing the unloaded world.
+	 */
+	@SubscribeEvent
+	public void unloadWorld( WorldEvent.Unload event )
+	{
+		if( Platform.isServer() )
+		{
+			final CompassReader compassReader = this.worldSet.remove( event.world );
```
thatsIch: Is it 100% guaranteed, that the world is contained? else `.close()` will result in `NPE`

thatsIch:
reviewed, left comment

src/main/java/appeng/services/CompassService.java
yueh: Good point. There should be a small chance that the world is not contained. I will update it.

There is probably also a chance that `event.world` might be null. But that is not an issue for a `HashMap` and otherwise when sending an event with `null` as world, they deserve a crash. :grinning:

yueh:
Added a `Map.containsKey(event.world)` to the condition. Explicitly checking for null is really pointless.

PR #1839, opened 2015-08-28
Fixed the order craftinglinks are notified for the last item.

### LINKED ISSUE #1833: Interface with Crafting Card stops filling 1 craft short if there's no storage for the item
Telling an Interface to stock items and give it a Crafting Card, on a network that has no storage for the stocked item causes some strange behavior. (For this test, I was attempting to stock 32 Stone Bricks).
 - The interface gets 28 of the 32 Stone Bricks as expected. Then it stops filling.
 - The interface on the Molecular Assembler accumulates Stone Bricks indefinitely.

Expected behavior was that 32 Stone Bricks would be crafted and sent to the interface, and then crafting would stop.

This is in an SSP world using only AE2.
[Gallery of screenshots and other information](http://imgur.com/gallery/8tDRi)

yueh:
Changed the order in which the job is marked as complete and when the last items are inserted into the requesting machine.
See the corresponding issue for a detailed explanation, about why it happens.

`injectItems()` is probably a good candidate for a general refactoring, but the control flow is not trivial and I currently cannot ensure that this will not break it.

Also some general cleanup.

Fixes #1833

### DIFF
--- a/src/main/java/appeng/me/cache/CraftingGridCache.java
+++ b/src/main/java/appeng/me/cache/CraftingGridCache.java
@@ -314,9 +314,9 @@ public class CraftingGridCache implements ICraftingGrid, ICraftingProviderHelper
 			{
 				this.craftingCPUClusters.add( cluster );
 
-				if( cluster.myLastLink != null )
+				if( cluster.getLastCraftingLink() != null )
 				{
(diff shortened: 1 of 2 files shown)


ruifung:
"Not trivial", somehow I feel like that is understating it a bit. XD

PR #1851, opened 2015-09-01
Fixes #1850: Fixed support for second optional output of the AE2 Grindstone

### LINKED ISSUE #1850: NEI Handler doesn't support "SecondOptionalOutput" for the Grinder
If I add a recipe to the Grinder that has three outputs, which vanilla AE2 doesn't have, I discovered that the third output - or "SecondOptionalOutput" - [does not show](http://i.imgur.com/wctPels.png) in the NEI Handler [although the crafting itself works](http://i.imgur.com/sAXAfo5.png). Tested on both release and alpha.
You might also want to add a line of text for the third output chance as that is missing as well! :)

thatsIch:


src/main/java/appeng/core/localization/GuiText.java
```diff
@@ -55,9 +55,13 @@
 
 	inWorldFluix, inWorldPurificationCertus, inWorldPurificationNether,
 
-	inWorldPurificationFluix, inWorldSingularity, ChargedQuartz, OfSecondOutput,
+	inWorldPurificationFluix, inWorldSingularity, ChargedQuartz,
 
-	NoSecondOutput, Stores, Next, SelectAmount, Lumen, Empty,
+	NoSecondOutput,
+	OfSecondOutput,
+	MultipleOutputs,
```
yueh: Any reason for splitting this up into 1 value per line?

thatsIch: Just readability, some groups will end up gigantic.

yueh: Feels just a bit random with only one line changed.

thatsIch: Just wanted to keep the noise down, can make a full change if requested

yueh: Fine for now, just wanted to ask in case the autoformatter triggered or so. But we can then slowly split them into multiple lines when editing it.

PR #1859, opened 2015-09-07
zh-tw.lang   translate something and achievement

asd0210a:
gui.appliedenergistics2.Drive=ME硬碟櫃
gui.appliedenergistics2.Stores=存儲

add achievement


thatsIch:
please squash the commits so that it only is 1 change

PR #1866, opened 2015-09-08
Fixes #1865: Spawn overflow items into the world

### LINKED ISSUE #1865: Annihilation Plane not correctly spawning overflow items
The Annihilation Plane is not correctly spawning overflowing items into the world as last resort, instead these are currently voided.

yueh:
Fixes #1865

### DIFF
--- a/src/main/java/appeng/parts/automation/PartAnnihilationPlane.java
+++ b/src/main/java/appeng/parts/automation/PartAnnihilationPlane.java
@@ -87,7 +87,7 @@ public class PartAnnihilationPlane extends PartBasicState implements IGridTickab
 	}
 
 	@Override
-	public TickRateModulation call(World world) throws Exception
+	public TickRateModulation call( World world ) throws Exception
 	{
 		this.breaking = false;
 		return this.breakBlock( true );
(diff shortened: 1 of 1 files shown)


yueh:
Updated to spawn the entity in the center of the broken block and without any motion.

So it no longers spawns on the border of block and thus gets moved somewhere else as this space is occupied and no motion keeps it inside the same blockspace.

src/main/java/appeng/parts/automation/PartAnnihilationPlane.java
```diff
@@ -359,7 +359,36 @@ private boolean handleOverflow( EntityItem entityItem, IAEItemStack overflow )
...
+	 * @param y y coordinate
+	 * @param z coordinate
+	 * @param overflow the item to spawn
+	 */
+	private void spawnOverflow( IAEItemStack overflow )
+	{
+		if( overflow == null )
+		{
+			return;
+		}
```
thatsIch: not sure if this method is the place where to silently reject no overflow, should probably be checked in the loop

yueh: Probably fine for this case, I usually just prefer handling this stuff at least also on the lowest level and not have to wrap every call of `PartAnnihilationPlane.spawnOverflow` with a null check. Having one single null check everything needs to go through simply is a bit less code.

I could change it to a `PreCondition` or `assert`. But that then throw an exception under some rare conditions. Also I cannot use assertions in all cases. Like enabled assertions will cause Extra Utilities to not start indev, which I needed to test it this.

thatsIch:
tested on server

PR #1877, opened 2015-09-15
ItemList refactoring

yueh:
Splitted the ItemList and MeaningfulIterator into an item and fluid version.
Added an IdentityHashMap as additional item layer to the ItemList for faster access.
The FluidList is a copy of the old ItemList.

The general approach is to use a `IdentityHashMap` on the item layer to speed up the access to a single item type. As `Item` is using a flyweight pattern, we can switch a `HashMap` for the slightly faster `IdentityHashMap`.
This actually mimics our intention with the fuzzy sorting, as this is only require on a per `Item` level, not over the whole collection. Thus it replaces the `O(log n)` in a deep tree with an access with `O(1)` and `O(log n)` on a very shallow tree. This also applies to oredict based searching, as this is based on merging different distinct items together into a collection.

It does result in a clear improvement under normal condition (about 50% less cpu consumed). Very high load can still be an issue (moving about 1024 itemtyps/tick through a network). Maybe it results in actually more items moved per tick, but this is really hard to track down as long as it is CPU gated.

The main issue prevent better optimizations is still the `MeaningfulItemIterator`. Due to the change it will now put a bit more load into the system as it needs to handle a second inner collection. But overall it is still a solid improvement. Getting rid of the actual self cleaning and replace it with a more fine grained explicitly cleanup is the main requirement here. Which will most likely require an API break.

Special thanks to @sfPlayer1 for the idea of using a `IdentityHashMap` and split the collection.

### DIFF
--- a/src/main/java/appeng/util/item/FluidList.java
+++ b/src/main/java/appeng/util/item/FluidList.java
@@ -1,0 +1,201 @@
+/*
+ * This file is part of Applied Energistics 2.
+ * Copyright (c) 2013 - 2015, AlgorithmX2, All rights reserved.
+ *
+ * Applied Energistics 2 is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Applied Energistics 2 is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Applied Energistics 2.  If not, see <http://www.gnu.org/licenses/lgpl>.
+ */
+
+package appeng.util.item;
+
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.Map;
+
+import appeng.api.config.FuzzyMode;
+import appeng.api.storage.data.IAEFluidStack;
+import appeng.api.storage.data.IItemList;
+
+
+public final class FluidList implements IItemList<IAEFluidStack>
+{
+
+	private final Map<IAEFluidStack, IAEFluidStack> records = new HashMap<IAEFluidStack, IAEFluidStack>();
+
+	@Override
+	public void add( IAEFluidStack option )
+	{
+		if( option == null )
+		{
+			return;
+		}
+
+		final IAEFluidStack st = this.getFluidRecord( option );
+
+		if( st != null )
+		{
+			st.add( option );
+			return;
+		}
+
+		final IAEFluidStack opt = option.copy();
+
+		this.putFluidRecord( opt );
+	}
+
+	@Override
+	public IAEFluidStack findPrecise( IAEFluidStack fluidStack )
+	{
+		if( fluidStack == null )
+		{
+			return null;
+		}
+
+		return this.getFluidRecord( fluidStack );
+	}
+
+	@Override
+	public Collection<IAEFluidStack> findFuzzy( IAEFluidStack filter, FuzzyMode fuzzy )
+	{
+		if( filter == null )
+		{
+			return Collections.emptyList();
+		}
+
+		return Collections.singletonList( this.findPrecise( filter ) );
+	}
+
+	@Override
+	public boolean isEmpty()
+	{
+		return !this.iterator().hasNext();
+	}
+
+	@Override
+	public void addStorage( IAEFluidStack option )
+	{
+		if( option == null )
+		{
+			return;
+		}
+
+		final IAEFluidStack st = this.getFluidRecord( option );
+
+		if( st != null )
+		{
+			st.incStackSize( option.getStackSize() );
+			return;
+		}
+
+		final IAEFluidStack opt = option.copy();
+
+		this.putFluidRecord( opt );
+	}
+
+	/*
+	 * public synchronized void clean() { Iterator<StackType> i = iterator(); while (i.hasNext()) { StackType AEI =
+	 * i.next(); if ( !AEI.isMeaningful() ) i.remove(); } }
+	 */
+
+	@Override
+	public void addCrafting( IAEFluidStack option )
+	{
+		if( option == null )
+		{
+			return;
+		}
+
+		final IAEFluidStack st = this.getFluidRecord( option );
+
+		if( st != null )
+		{
+			st.setCraftable( true );
+			return;
+		}
+
+		final IAEFluidStack opt = option.copy();
+		opt.setStackSize( 0 );
+		opt.setCraftable( true );
+
+		this.putFluidRecord( opt );
+	}
+
+	@Override
+	public void addRequestable( IAEFluidStack option )
+	{
+		if( option == null )
+		{
+			return;
+		}
+
+		final IAEFluidStack st = this.getFluidRecord( option );
+
+		if( st != null )
+		{
+			st.setCountRequestable( st.getCountRequestable() + option.getCountRequestable() );
+			return;
+		}
+
+		final IAEFluidStack opt = option.copy();
+		opt.setStackSize( 0 );
+		opt.setCraftable( false );
+		opt.setCountRequestable( option.getCountRequestable() );
+
+		this.putFluidRecord( opt );
+	}
+
+	@Override
+	public IAEFluidStack getFirstItem()
+	{
+		for( IAEFluidStack stackType : this )
+		{
+			return stackType;
+		}
+
+		return null;
+	}
+
+	@Override
+	public int size()
+	{
+		return this.records.values().size();
+	}
+
+	@Override
+	public Iterator<IAEFluidStack> iterator()
+	{
+		return new MeaningfulFluidIterator<IAEFluidStack>( this.records.values().iterator() );
+	}
+
+	@Override
+	public void resetStatus()
+	{
+		for( IAEFluidStack i : this )
+		{
+			i.reset();
+		}
+	}
+
+	private IAEFluidStack getFluidRecord( IAEFluidStack fluid )
+	{
+		return this.records.get( fluid );
+	}
+
+	private IAEFluidStack putFluidRecord( IAEFluidStack fluid )
+	{
+		return this.records.put( fluid, fluid );
+	}
+}
--- a/src/main/java/appeng/util/item/ItemList.java
+++ b/src/main/java/appeng/util/item/ItemList.java
@@ -1,6 +1,6 @@
 /*
  * This file is part of Applied Energistics 2.
- * Copyright (c) 2013 - 2014, AlgorithmX2, All rights reserved.
+ * Copyright (c) 2013 - 2015, AlgorithmX2, All rights reserved.
  *
  * Applied Energistics 2 is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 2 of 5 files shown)


src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,280 @@
...
+
+public final class FluidList<StackType extends IAEStack> implements IItemList<StackType>
+{
+
+	private final NavigableMap<StackType, StackType> records = new ConcurrentSkipListMap<StackType, StackType>();
+	private final Class<? extends IAEStack> clz;
+
+	// private int currentPriority = Integer.MIN_VALUE;
+	public Throwable stacktrace;
+	int iteration = Integer.MIN_VALUE;
```
thatsIch: I think this is not used anywhere?

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,280 @@
...
+
+
+public final class FluidList<StackType extends IAEStack> implements IItemList<StackType>
+{
+
+	private final NavigableMap<StackType, StackType> records = new ConcurrentSkipListMap<StackType, StackType>();
+	private final Class<? extends IAEStack> clz;
+
+	// private int currentPriority = Integer.MIN_VALUE;
+	public Throwable stacktrace;
```
thatsIch: legacy code?

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,280 @@
...
+import net.minecraftforge.oredict.OreDictionary;
+
+import appeng.api.config.FuzzyMode;
+import appeng.api.storage.data.IAEFluidStack;
+import appeng.api.storage.data.IAEItemStack;
+import appeng.api.storage.data.IAEStack;
+import appeng.api.storage.data.IItemList;
+
+
+public final class FluidList<StackType extends IAEStack> implements IItemList<StackType>
```
thatsIch: I think the correct way is `StackType extends IAEStack<StackType>` although why not getting rid of it all together since you know it is a `IAEFluidStack`, so basically replace `StackType extends IAEStack` with `IAEFluidStack`?

thatsIch: erm, I meant `IItemList<StackType>` with `IItemList<IAEFluidStack>`

src/main/java/appeng/util/item/FluidList.java
yueh: Nope, but I simply did not want to mess with anything fluid related for now.

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -21,47 +21,54 @@
...
 
 import appeng.api.config.FuzzyMode;
-import appeng.api.storage.data.IAEFluidStack;
 import appeng.api.storage.data.IAEItemStack;
-import appeng.api.storage.data.IAEStack;
 import appeng.api.storage.data.IItemList;
 
 
-public final class ItemList<StackType extends IAEStack> implements IItemList<StackType>
+public final class ItemList<T extends IAEItemStack> implements IItemList<T>
```
thatsIch: Why use `StackType` in `FluidList` and renamed in `ItemList` it back to `T`

src/main/java/appeng/util/item/FluidList.java
yueh: Yeah. Removed it from the new ItemList, just will not appear as unused as it is public.

src/main/java/appeng/util/item/MeaningfulItemIterator.java
```diff
@@ -0,0 +1,89 @@
+/*
+ * This file is part of Applied Energistics 2.
+ * Copyright (c) 2013 - 2014, AlgorithmX2, All rights reserved.
```
thatsIch: outdated copyright notes?

src/main/java/appeng/util/item/FluidList.java
yueh: Yes, we can even go a bit further and remove the need for passing the `.class` to the constructor and let it handle by the compiler.

The whole fluidlist can now have a completely new implementation. Like do we actually need a fuzzy filtering? (are there even damagable fluids?) Can we just store every fluid into a `HashMap` or `IdentityHashMap` etc. Tons of possible optimizations here.
But for now it is just about the `ItemList`

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,219 @@
...
+	public Collection<IAEFluidStack> findFuzzy( IAEFluidStack filter, FuzzyMode fuzzy )
+	{
+		if( this.isStackNull( filter ) )
+		{
+			return Collections.emptyList();
+		}
+
+		final List<IAEFluidStack> result = Lists.newArrayList();
+
+		if( filter.equals( this ) )
```
orod-org: MAJOR Remove this call to "equals"; comparisons between unrelated types always return false. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2159)

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,219 @@
...
+
+	@Override
+	public IAEFluidStack findPrecise( IAEFluidStack i )
+	{
+		if( this.isStackNull( i ) )
+		{
+			return null;
+		}
+
+		final IAEFluidStack is = this.getFluidRecord( i );
```
orod-org: MINOR Immediately return this expression instead of assigning it to the temporary variable "is". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1488)

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,219 @@
...
+
+		return fluidrecord;
+	}
+
+	private IAEFluidStack putFluidRecord( IAEFluidStack fluid )
+	{
+		return this.records.put( fluid, fluid );
+	}
+
+	private boolean isStackNull( IAEFluidStack stack )
```
orod-org: MINOR Make "isStackNull" a "static" method. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2325)

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,219 @@
...
+			st.incStackSize( option.getStackSize() );
+			return;
+		}
+
+		final IAEFluidStack opt = option.copy();
+		this.putFluidRecord( opt );
+	}
+
+	/*
+	 * public synchronized void clean() { Iterator<StackType> i = iterator(); while (i.hasNext()) { StackType AEI =
```
orod-org: MAJOR This block of commented-out lines of code should be removed. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3ACommentedOutCodeLine)

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,219 @@
...
+	{
+		for( IAEFluidStack i : this )
+		{
+			i.reset();
+		}
+	}
+
+	private IAEFluidStack getFluidRecord( IAEFluidStack fluid )
+	{
+		final IAEFluidStack fluidrecord = this.records.get( fluid );
```
orod-org: MINOR Immediately return this expression instead of assigning it to the temporary variable "fluidrecord". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1488)

src/main/java/appeng/util/item/MeaningfulFluidIterator.java
```diff
@@ -24,13 +24,13 @@
 import appeng.api.storage.data.IAEStack;
 
 
-public class MeaningfulIterator<StackType extends IAEStack> implements Iterator<StackType>
+public class MeaningfulFluidIterator<StackType extends IAEStack> implements Iterator<StackType>
```
orod-org: MINOR Rename this generic name to match the regular expression '^[A-Z][0-9]?$'. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS00119)

src/main/java/appeng/util/item/MeaningfulItemIterator.java
```diff
@@ -0,0 +1,89 @@
...
+			{
+				this.innerIterater = this.parent.next().values().iterator();
+			}
+		}
+
+		return false;
+	}
+
+	@Override
+	public T next()
```
orod-org: MAJOR Add a "NoSuchElementException" for iteration beyond the end of the collection. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2272)

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -258,23 +201,61 @@ public synchronized StackType getFirstItem()
...
+
+	private Collection<IAEItemStack> findFuzzyDamage( AEItemStack filter, FuzzyMode fuzzy, boolean ignoreMeta )
+	{
+		final IAEItemStack low = filter.getLow( fuzzy, ignoreMeta );
+		final IAEItemStack high = filter.getHigh( fuzzy, ignoreMeta );
+
+		return this.getItemRecord( filter.getItem() ).subMap( low, true, high, true ).descendingMap().values();
+	}
+
+	private boolean isStackNull( IAEItemStack stack )
```
orod-org: MINOR Make "isStackNull" a "static" method. [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2325)

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -158,99 +117,83 @@ public boolean isEmpty()
...
 		if( st != null )
 		{
-			// st.setPriority( currentPriority );
 			st.setCraftable( true );
 			return;
 		}
 
-		StackType opt = (StackType) option.copy();
-		// opt.setPriority( currentPriority );
+		final IAEItemStack opt = (IAEItemStack) option.copy();
```
orod-org: MINOR Remove this unnecessary cast to "IAEItemStack". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1905)

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -158,99 +117,83 @@ public boolean isEmpty()
...
 			return;
 		}
 
-		StackType st = this.records.get( option );
+		final IAEItemStack st = this.getItemRecord( option.getItem() ).get( option );
 
 		if( st != null )
 		{
-			// st.setPriority( currentPriority );
 			( (IAEItemStack) st ).setCountRequestable( st.getCountRequestable() + option.getCountRequestable() );
```
orod-org: MINOR Remove this unnecessary cast to "IAEItemStack". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1905)

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -158,99 +117,83 @@ public boolean isEmpty()
...
 		{
-			// st.setPriority( currentPriority );
 			st.incStackSize( option.getStackSize() );
 			return;
 		}
 
-		StackType opt = (StackType) option.copy();
-		// opt.setPriority( currentPriority );
-		this.records.put( opt, opt );
+		final IAEItemStack opt = (IAEItemStack) option.copy();
```
orod-org: MINOR Remove this unnecessary cast to "IAEItemStack". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1905)

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -158,99 +117,83 @@ public boolean isEmpty()
...
 		if( st != null )
 		{
-			// st.setPriority( currentPriority );
 			( (IAEItemStack) st ).setCountRequestable( st.getCountRequestable() + option.getCountRequestable() );
 			return;
 		}
 
-		StackType opt = (StackType) option.copy();
-		// opt.setPriority( currentPriority );
+		final IAEItemStack opt = (IAEItemStack) option.copy();
```
orod-org: MINOR Remove this unnecessary cast to "IAEItemStack". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1905)

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -19,126 +19,85 @@
...
 		{
 			return null;
 		}
 
-		StackType is = this.records.get( i );
-		if( is != null )
-		{
-			return is;
-		}
+		final IAEItemStack is = this.getItemRecord( i.getItem() ).get( i );
```
orod-org: MINOR Immediately return this expression instead of assigning it to the temporary variable "is". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1488)

orod-org:
SonarQube analysis reported 14 issues:
* MAJOR 4 major
* MINOR 10 minor

* MAJOR [Add a "NoSuchElementException" for iteration beyond the end of the collection.](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/1121b11249aaece1aaa0e81009d01ed6c6ce52d4/src/main/java/appeng/util/item/MeaningfulFluidIterator.java#L58) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS2272)

src/main/java/appeng/util/item/FluidList.java
thatsIch: I would disable the static method stuff, static is evil

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,219 @@
...
+
+import appeng.api.config.FuzzyMode;
+import appeng.api.storage.data.IAEFluidStack;
+import appeng.api.storage.data.IItemList;
+
+
+public final class FluidList implements IItemList<IAEFluidStack>
+{
+
+	private final Map<IAEFluidStack, IAEFluidStack> records = new HashMap<IAEFluidStack, IAEFluidStack>();
```
thatsIch: Just wondering why a `HashMap`? I feel like a `Bag` would do the same Job here, as it seems the key and value are the same and thus even duplicated. Are these also unique fluids? or can there be duplicates.

src/main/java/appeng/util/item/FluidList.java
```diff
@@ -0,0 +1,219 @@
...
+		if( st != null )
+		{
+			st.setCountRequestable( st.getCountRequestable() + option.getCountRequestable() );
+			return;
+		}
+
+		final IAEFluidStack opt = option.copy();
+		opt.setStackSize( 0 );
+		opt.setCraftable( false );
+		opt.setCountRequestable( opt.getCountRequestable() );
```
thatsIch: `opt.getCountRequestable()` and `option.getCountRequestable()` are the same? if yes, can outsource variable before `if`

src/main/java/appeng/util/item/FluidList.java
yueh: Already considered it. 

But then they have a point about their description.
It clearly indicates this method will not mutate the current state similar to `final` for variables.
Also only limited to `private`

src/main/java/appeng/util/item/ItemList.java
thatsIch: agreed, should be same type since you do not use the `StackType` anymore

src/main/java/appeng/util/item/ItemList.java
thatsIch: same.

src/main/java/appeng/util/item/ItemList.java
thatsIch: s.o.

src/main/java/appeng/util/item/MeaningfulItemIterator.java
```diff
@@ -0,0 +1,89 @@
...
+package appeng.util.item;
+
+
+import java.util.Iterator;
+import java.util.NavigableMap;
+
+import appeng.api.storage.data.IAEItemStack;
+
+
+public class MeaningfulItemIterator<T extends IAEItemStack> implements Iterator<T>
```
thatsIch: How does this work for the `FluidList`? Should this not be `IAEStack<T>`?

src/main/java/appeng/util/item/FluidList.java
yueh: Fluids should be unique, so a `Set` would even fit better. But they do not have a `get(T o)`, so we can only use `contains(T o)`. But this uses `equals()`, thus we cannot get the actual stored stack easily. The same more or less applies to itemstacks to.

Then Fluids are not guaranteed to be singletons. They are designed as them, but Forge simply fails at providing a reasonable design, which prevents creating multiple instances of the same fluid (as usually...)

src/main/java/appeng/util/item/FluidList.java
yueh: No change here, so it really is a potential bug from the beginning. Should most likely be `option` not `opt`

src/main/java/appeng/util/item/ItemList.java
yueh: Intentionally left to test the tool. :grinning:

src/main/java/appeng/util/item/MeaningfulItemIterator.java
yueh: `FluidList` has its own iterator as they are no longer share the same implementation. Thus `IAEItemStack` is perfectly valid.

src/main/java/appeng/parts/automation/PartImportBus.java
```diff
@@ -284,6 +285,31 @@ private boolean importStuff( InventoryAdaptor myAdaptor, IAEItemStack whatToImpo
 		return false;
 	}
 
+	private int calculateMaximumAmountToImport( InventoryAdaptor myAdaptor, IAEItemStack whatToImport, IMEMonitor<IAEItemStack> inv, IEnergySource energy, FuzzyMode fzMode )
```
orod-org: MAJOR Remove the unused method parameter(s) "energy". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1172)

orod-org:
SonarQube analysis reported 5 issues:
* MAJOR 3 major
* MINOR 2 minor

orod-org:
SonarQube analysis reported 4 issues:
* MAJOR 2 major
* MINOR 2 minor

src/main/java/appeng/util/item/FluidList.java
thatsIch: would remove this method all together. It does not handle with complicated booleans or so, just checking for `null`

src/main/java/appeng/util/item/MeaningfulItemIterator.java
thatsIch: oh `ItemIterator`

reading and stuff

src/main/java/appeng/util/item/FluidList.java
yueh: Probably. With refactoring the generics etc, the method simply got shorter each time.
I wish we had the non local returns of scala. So we could just call `isStackNull()` wherever it is needed and not wrap everything into an if condition..

thatsIch: hail refactoring I guess

orod-org:
SonarQube analysis reported 2 issues:
* MAJOR 2 major

orod-org:
* MAJOR 1 major

yueh:
Change the `FluidList.findFuzzy` to return a collection containing the precise fluid instead of allways an empty collection.

@ruifung Do you know if ExtraCells actually does use this functionality or if something like oredict fluids are even a thing?

ruifung:
Is fuzzy even relevant for fluids? The FluidRegistry only allows one Fluid to be associated with a fluid name after all, and unlike the oredictionary does not let you convert between different Fluids with the same name. 

And EC2 does not appear to use anything more then AE2's API. findFuzzy isn't defined in the interface after all.

yueh:
That is the question. I have not really looked into fluids besides just the most basic things. So asking someone who is a bit more familiar with it is probably a good idea. Especially in regards to the bug regarding "fuel".

`FluidList` will be added with this PR, but this is only the internal implementation provide by the API.
As long as you are using the API to create a new `IItemList<IAEFluidStack>`, you are perfectly fine.
The reason for it is that we want to improve the implementation for the item storage and fluids really do not fit well, thus we splitting the internal implementation.

Especially for fluids we do not need a sorted collection as backend and are able to replace it with a simple `HashMap`, thus replacing `O(log n)` with `O(1)` for most operations. There are not that many fluids to really make a huge impact. But for example iterating over the whole collection is now `O(n)` instead of `O(n log n)`.

For items it is sadly not that easy as we still need the fuzzy operations and a `NavigableMap` simply has to many advantages over other solutions as the are sorted once per add and not with every single get.

ruifung:
Bug regarding fuel as in that issue on the ec2 github? Yeah, from what I can tell forge assumes only one Fluid will have any given name at a time. But some mods might use their own Fluid classes and cause stuff to break when FluidA with name fuel and FluidB with name fuel doesn't match. Heck, from what I can see, Fluids look like they are supposed to be singletons. And probably should also be compared by their names instead of just the Fluid objects.

yueh:
Everything around fluids is designed as them being singletons. The only thing which do not know this are fluids themselves, thus allowing them to break everything.

orod-org:
* MAJOR 1 major

orod-org:
* MAJOR 1 major

PR #1879, opened 2015-09-16
Fixes #1878: Now saves the dimensionlist correctly

### LINKED ISSUE #1878: DimensionData not saved correctly
The `DimensionData` mapping ids to spatial dimension is not storing these inside the config file.

yueh:
Fixes #1878

### DIFF
--- a/src/main/java/appeng/core/worlddata/DimensionData.java
+++ b/src/main/java/appeng/core/worlddata/DimensionData.java
@@ -19,7 +19,6 @@
 package appeng.core.worlddata;
 
 
-import java.io.File;
 import java.util.List;
 
 import javax.annotation.Nonnull;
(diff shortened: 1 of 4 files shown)


yueh:
No idea why the first commit even worked...

Nevertheless `Configuration` really does not like using a shared file. Sometimes it works, sometimes not. Thus it now uses a single shared instance.

src/main/java/appeng/core/worlddata/DimensionData.java
```diff
@@ -120,7 +117,7 @@ public void addStorageCell( int newStorageCellID )
 
 		String[] values = new String[this.storageCellDimensionIDs.size()];
 
-		for( int x = 0; x < values.length; x++ )
+		for( int x = 0; x < this.storageCellDimensionIDs.size(); x++ )
```
thatsIch: super important change xD

yueh: The first useless one. I should change it back...

For whatever reason it worked the first time... Even then no idea why it worked.
But switching back between both version was reliably working with the change but not the old one....

PR #1881, opened 2015-09-17
Update fr_FR.lang

Mazdallier:


yueh:
I have no idea about the french grammar, but is there any special reason to write quartz as lower case? As `Certus Quartz` is the proper name for it and not something similar to `quartz of type Certus`.

Changing `Quartz du Nether` to `quartz du Tréfonds` could also add some confusion as the first option is the official translation. I am certain that most players will understand it, but breaking with the official one should really be avoided.

Pyeroh:
I'd write ```Certus Quartz``` in french, and ```Quartz du Tréfonds``` probably comes from canadian french (Quebec). In France, it doesn't matter that much to use some anglicisms, if it helps understanding. ```Nether``` is of course an english word, but I can guarantee you that no one uses ```Tréfonds``` in "France" french

Mazdallier:
Quartz is like Iron, so lower case this commun name is normal.
Certus is like Nether, this is a proper name. We allway upper case this.
Certus existe in French but not Nether, the current translation in the french wiki (http://minecraft-fr.gamepedia.com/Le_Nether) is Tréfonds or Nexus. Tréfonds is the most explicite of both.

thatsIch:
How does Certus exists in French? It does not have the same meaning as in French I assume. The original name is derived from Latin

http://mymemory.translated.net/de/Latein/Franz%C3%B6sisch/certus

I strongly believe that Certus and Nether are considered own Names here, see

http://minecraft-fr.gamepedia.com/Quartz_du_Nether_(gemme)

Mazdallier:
Certus is also a trademark

PR #1893, opened 2015-09-22
Fixes incorrect handling of prioritized inventories

### LINKED ISSUE #1892: Import bus not respecting the storable amount when importing.
(Yes, I really wrote import bus)

I made a processor setup using inscribers and storage buses, where each inscriber has a storage bus with a filter as the input, and an import bus as the output. Pretty simple. However, from time to time, it gets jammed because one of the printed circuits ends up back in the inscriber's middle slot, which shouldn't happen because the storage bus is configured to only accept the correct items (ie. silicon, gold, pure certus or diamonds) and the only other thing connected to the inscriber is an import bus.

storage-bus
bus-config

After further investigation, it appears that:

* The last inscriber on the line (the one that assembles printed circuits into processors) needs to be running for the issue to present itself.
* Removing the storage buses on the first 4 inscribers doesn't make the issue go away.
* Adding a controller and making it all one network also fixes the issue.

I suspect the import bus pulls the items from the inscriber, tries to put it into the other network through the storage bus + interface combo, fails to do so because, and then tries to push it back into the inscriber. Since the sides of the inscriber puts items in the middle slot and not the output slot, the printed circuit ends back in the middle slot instead of the output slot where it comes from. Additionally, if there's items in the middle slot, the printed circuit is destroyed and lost in the process.

Happens in both rv2.stable build 10 and rv3.alpha build 6 downloaded less than an hour ago. AE is the only mod loaded, using the latest recommended forge (10.13.4.1448). Reproduced both in single player and multiplayer.

Video (on server): http://d.max-p.me/temp/irc/ftb/inscriber.mp4
Video (import bus alone): http://d.max-p.me/temp/irc/ftb/inscriber-2.mp4
World save exposing the issue: http://d.max-p.me/temp/irc/ftb/Inscriber.zip

yueh:
Prioritized inventories are not longer used twice for storing leftover
items and thus finally reporting twice the amount of storable items when
they are the only possible option to store something.

Also fixes import buses now respecting the amount of storable items inside
the network instead of trying to place the exported items back and failing
on any restricted inventory, potentially voiding the overflow.

Fixes #1892

### DIFF
--- a/src/main/java/appeng/parts/automation/PartImportBus.java
+++ b/src/main/java/appeng/parts/automation/PartImportBus.java
@@ -54,6 +54,7 @@ import appeng.me.GridAccessException;
 import appeng.util.InventoryAdaptor;
 import appeng.util.Platform;
 import appeng.util.inv.IInventoryDestination;
+import appeng.util.item.AEItemStack;
 
 
 public class PartImportBus extends PartSharedItemBus implements IInventoryDestination
@@ -238,8 +239,8 @@ public class PartImportBus extends PartSharedItemBus implements IInventoryDestin
 
 	private boolean importStuff( InventoryAdaptor myAdaptor, IAEItemStack whatToImport, IMEMonitor<IAEItemStack> inv, IEnergySource energy, FuzzyMode fzMode )
 	{
-		final int toSend = Math.min( this.itemToSend, 64 );
-		ItemStack newItems;
+		final int toSend = this.calculateMaximumAmountToImport( myAdaptor, whatToImport, inv, fzMode );
+		final ItemStack newItems;
 
 		if( this.getInstalledUpgrades( Upgrades.FUZZY ) > 0 )
 		{
@@ -264,8 +265,8 @@ public class PartImportBus extends PartSharedItemBus implements IInventoryDestin
 				this.lastItemChecked.setStackSize( newItems.stackSize );
 			}
 
-			IAEItemStack failed = Platform.poweredInsert( energy, this.destination, this.lastItemChecked, this.source );
-			// destination.injectItems( lastItemChecked, Actionable.MODULATE );
+			final IAEItemStack failed = Platform.poweredInsert( energy, this.destination, this.lastItemChecked, this.source );
+
 			if( failed != null )
 			{
 				myAdaptor.addItems( failed.getItemStack() );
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/parts/automation/PartImportBus.java
```diff
@@ -284,6 +285,31 @@ private boolean importStuff( InventoryAdaptor myAdaptor, IAEItemStack whatToImpo
 		return false;
 	}
 
+	private int calculateMaximumAmountToImport( InventoryAdaptor myAdaptor, IAEItemStack whatToImport, IMEMonitor<IAEItemStack> inv, IEnergySource energy, FuzzyMode fzMode )
```
orod-org: MAJOR Remove the unused method parameter(s) "energy". [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1172)

orod-org:
* MAJOR 1 major

src/main/java/appeng/parts/automation/PartImportBus.java
```diff
@@ -284,6 +285,31 @@ private boolean importStuff( InventoryAdaptor myAdaptor, IAEItemStack whatToImpo
 		return false;
 	}
 
+	private int calculateMaximumAmountToImport( InventoryAdaptor myAdaptor, IAEItemStack whatToImport, IMEMonitor<IAEItemStack> inv, FuzzyMode fzMode )
+	{
+		final int toSend = Math.min( this.itemToSend, 64 );
```
thatsIch: wondering, why 64? always though that the max is 96

src/main/java/appeng/parts/automation/PartImportBus.java
```diff
@@ -284,6 +285,31 @@ private boolean importStuff( InventoryAdaptor myAdaptor, IAEItemStack whatToImpo
...
+		final ItemStack simResult;
+		final IAEItemStack itemAmountNotStorable;
+
+		if( this.getInstalledUpgrades( Upgrades.FUZZY ) > 0 )
+		{
+			simResult = myAdaptor.simulateSimilarRemove( toSend, whatToImport == null ? null : whatToImport.getItemStack(), fzMode, this.configDestination( inv ) );
```
thatsIch: arguments of both cases can be extracted to the front of the `if`clause, 
would also improve readability

src/main/java/appeng/parts/automation/PartImportBus.java
yueh: It is still working with normal itemstacks, not their AE2 version. So 64 is the maximum for them.
Also I would not be surprised, if some inventories could crash if you ask to import a stack with a size of 96.

PR #1902, opened 2015-09-23
Fixes 1896: Only update a monitor if the text would change

### LINKED ISSUE #1896: Storage Montors are possibly rendering too often
The storage monitors are using the TESR upon list change which makes sense. 
But there is a possibility that the amount is not taken into consideration before marking a storage monitor to update. (Only need to re-render if the amount hits a specific threshold for the amount renderer to render something different? e.g. 10k are still rendered the same even with 10001 items)

Did not investigate this nor did I look into the code. Just had a hunch when hearing somebody saying that too many storage monitors is lagging for him in a large storage solution.

yueh:
Fixes #1896

### DIFF
--- a/src/main/java/appeng/parts/reporting/AbstractPartMonitor.java
+++ b/src/main/java/appeng/parts/reporting/AbstractPartMonitor.java
@@ -79,6 +79,7 @@ public abstract class AbstractPartMonitor extends AbstractPartDisplay implements
 {
 	private static final IWideReadableNumberConverter NUMBER_CONVERTER = ReadableNumberConverter.INSTANCE;
 	private IAEItemStack configuredItem;
+	private String lastHumanReadableText;
 	private boolean isLocked;
 	private IStackWatcher myWatcher;
 	@SideOnly( Side.CLIENT )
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/parts/reporting/AbstractPartMonitor.java
```diff
@@ -79,6 +79,7 @@
 {
 	private static final IWideReadableNumberConverter NUMBER_CONVERTER = ReadableNumberConverter.INSTANCE;
 	private IAEItemStack configuredItem;
+	private String itemStackSize;
```
thatsIch: I think the naming of the variable is quite bad since it is not the item stack size being stored (which would natively be an integer) but a render cache or something along the line. Maybe just `textRenderCache` or so.

yueh: Yeah, except that this is serverside and thus containing `render` is also not the best idea.

thatsIch: what about `lastTextDisplay` or just `lastText` and add a comment on top of the variable. Else this just sounds like a HÄÄÄ moment few months later.

yueh: Changed to `lastHumanReadableText`

PR #1912, opened 2015-09-25
Replace with array initializer

thatsIch:
this can be done through http://docs.oracle.com/javase/specs/jls/se6/html/arrays.html#10.6


yueh:
I remember something faintly about them...

PR #1921, opened 2015-09-27
Nudge Better Gradle Version

RlonRyan:
I am on a mission to get mods to update to a better version of Gradle.

If you don't wish to merge simply run <code>gradlew.bat wrapper
--gradle-version 2.7</code>


yueh:
Can you guarantee that any existing plugin will work with this? Especially the latest gradle versions are extremely sensitive about the order inside a `build.gradle` and can easily break.

I also do not see any particular improvement in terms of runspeed. The current version builds more or less as fast as this one and we especially start with a clean cache each time to ensure nothing old or broken affects the build (which happen a couple of times in the past without it)

RlonRyan:
No guarantees... just saying it might be worth the effort, as there are some interesting features to the latest Gradle.

yueh:
Based on the changelogs there is absolutely no improvement except when using non public repositories or the play framework and a potential break for any eclipse user.

thatsIch:
@yueh your server failed for some unknown reason

yueh:
I tried a manual build vs an automatic one to test the potential speedup (which does actually not exist).
But in this case the sonar plugin fails due to a syntax error. I did not have time to verify it is an issue with gradle, a potentially error with the jenkins plugin or simply gradle 2.7 not compatible with it.

Also I cannot remember, if Algo's server is using gradle directly or the wrapper. So this is also a potential issue, which we cannot really test.

So to summarize, it might break our current build setup, does not achieve any noticable speedup except that it claims to fix a performance regression of 2.6 (which we do not use) and no new features which would directly benefit us.

RlonRyan:
Fascinating... It does seem like the version you have is a better version in regards to your usage cases, my apologies for the pr.

PR #1934, opened 2015-10-05
Fixes #1932: Better VersionChecker exception handling

### LINKED ISSUE #1932: VersionChecker unhandled exceptions
My friend has that weird freeze issue. From what he sees, there's some weird exception in AE2 VersionChecker which freezes his game dead stop.
https://dl.dropboxusercontent.com/u/77108893/Crash%20Reports/AE2%20log.txt

yueh:
ModVersionFetcher will now return a MissingVersion in case of an exception
instead of letting it propagate upwards.

Also added a generic try/catch to the VersionChecker itself, just in case
any unchecked exception might be triggered inside the thread and at least
not logged correctly.

### DIFF
--- a/src/main/java/appeng/services/VersionChecker.java
+++ b/src/main/java/appeng/services/VersionChecker.java
@@ -21,6 +21,10 @@ package appeng.services;
 
 import java.util.Date;
 
+import javax.annotation.Nonnull;
+
+import com.google.common.base.Preconditions;
+
 import net.minecraft.nbt.NBTTagCompound;
 
 import cpw.mods.fml.common.Loader;
@@ -66,30 +70,40 @@ public final class VersionChecker implements Runnable
 	private static final int MS_TO_SEC = 1000;
 	private final VersionCheckerConfig config;
 
-	public VersionChecker( final VersionCheckerConfig config )
+	public VersionChecker( @Nonnull final VersionCheckerConfig config )
 	{
+		Preconditions.checkNotNull( config );
+
 		this.config = config;
 	}
 
 	@Override
 	public void run()
 	{
-		Thread.yield();
+		try
+		{
+			Thread.yield();
 
-		// persist the config
-		this.config.save();
+			// persist the config
+			this.config.save();
 
-		// retrieve data
-		final String rawLastCheck = this.config.lastCheck();
+			// retrieve data
+			final String rawLastCheck = this.config.lastCheck();
 
-		// process data
-		final long lastCheck = Long.parseLong( rawLastCheck );
-		final Date now = new Date();
-		final long nowInMs = now.getTime();
-		final long intervalInMs = this.config.interval() * SEC_TO_HOUR * MS_TO_SEC;
-		final long lastAfterInterval = lastCheck + intervalInMs;
+			// process data
+			final long lastCheck = Long.parseLong( rawLastCheck );
+			final Date now = new Date();
+			final long nowInMs = now.getTime();
+			final long intervalInMs = this.config.interval() * SEC_TO_HOUR * MS_TO_SEC;
+			final long lastAfterInterval = lastCheck + intervalInMs;
 
-		this.processInterval( nowInMs, lastAfterInterval );
+			this.processInterval( nowInMs, lastAfterInterval );
+		}
+		catch( Exception exception )
+		{
+			// Log any unhandled exception to prevent the JVM from reporting them as unhandled.
+			AELog.error( exception );
+		}
 
 		AELog.info( "Stopping AE2 VersionChecker" );
 	}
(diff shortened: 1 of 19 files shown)


orod-org:
* BLOCKER 1 blocker

* BLOCKER [Add or update the header of this file.](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/af33ac74eb830612a5793c387673b567238ddae5/src/main/java/appeng/services/version/ModVersionFetcher.java) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=squid%3AS1451)

src/main/java/appeng/services/VersionChecker.java
```diff
@@ -76,16 +76,24 @@ public void run()
...
+			final long nowInMs = now.getTime();
+			final long intervalInMs = this.config.interval() * SEC_TO_HOUR * MS_TO_SEC;
+			final long lastAfterInterval = lastCheck + intervalInMs;
+
+			this.processInterval( nowInMs, lastAfterInterval );
 
-		this.processInterval( nowInMs, lastAfterInterval );
+			AELog.info( "Stopping AE2 VersionChecker" );
+		}
+		catch( Exception e )
```
thatsIch: just `Exception` pretty sure, you can specify it

src/main/java/appeng/services/version/ModVersionFetcher.java
```diff
@@ -30,11 +53,18 @@ public Version get()
...
 		}
-		else
+
+		try
 		{
 			final Version version = this.parser.parse( this.rawModVersion );
 
 			return version;
 		}
+		catch( final Exception e )
```
thatsIch: should be more specified, what kind of exception can even be thrown here. The only one I did add, was an `AssertionError`

src/main/java/appeng/services/VersionChecker.java
yueh: More intented as a catch all to handle any unexpected (unchecked) exception bubbling up and prevent them from reaching the main thread and reporting something about an unhandled exception.

src/main/java/appeng/services/version/ModVersionFetcher.java
yueh: Nearly no user will ever encounter an `AssertionError` as like nobody starts their client with `-ea`, thus unit tests are more or less the only place they are thrown.

In the related issues an `IllegalArgumentException` bubbles up, due to the disabled assertions.

If the intention for the assertions is that they should always be triggered, then `Preconditions.checkArgument()` is probably a better alternative.

src/main/java/appeng/services/VersionChecker.java
thatsIch: This will never reach the main thread. unchecked exceptions are bugs, why catching them unless you do not want them being reported. But I guess here is fine since its `AELog.error()`

src/main/java/appeng/services/version/ModVersionFetcher.java
thatsIch: Should be probably exported to a new class handling that. Thus can also make it testable.

src/main/java/appeng/services/VersionChecker.java
yueh: main thread as in log "Exception in thread ?" as well as log the exception to `stdout`.
This is just to prevent the jvm from reporting any unhandled exception, but still log the exception in the right context (and not simply `stdout`)

src/main/java/appeng/services/version/ModVersionFetcher.java
yueh: Considering something similar already. Basically replacing all assertion through checked exceptions, but I wanted to hear your feedback first as you implemented the version checker.

thatsIch: Makes sense to me

orod-org:
* CRITICAL 1 critical

* CRITICAL [Redundant nullcheck of appeng.services.version.BaseVersion.channel, which is known to be non-null in appeng.services.version.BaseVersion.hashCode()](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/be10b86d172e566fbd3fd1cb4841735f710b7e83/src/main/java/appeng/services/version/BaseVersion.java#L88) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=findbugs%3ARCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE)

orod-org:
* CRITICAL 1 critical

* CRITICAL [Redundant nullcheck of appeng.services.version.BaseVersion.channel, which is known to be non-null in appeng.services.version.BaseVersion.hashCode()](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/f45b86231680951e53117f3847aea9865e36f817/src/main/java/appeng/services/version/BaseVersion.java#L88) [rule](http://nemo.sonarqube.org/coding_rules#rule_key=findbugs%3ARCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE)

thatsIch:
Seems the null check is really redundant

PR #1940, opened 2015-10-09
Fixes #1938: UnsortedList not respecting meaningful entries.

### LINKED ISSUE #1938: Items Access seems broken
we are running on a small server and using the newest rv3a8 and a lot of other mods as we noticed 2 odd things

first ist the listing of items in the system while taking out an item
best way to reproduce was by dropping different kinds of potions in the system
when you search for potion, you'll see them all, but when you take a potion out, the list halfs or sometimes also gets empty until you search again
(it's probably related to us havin still extracells installed, but i'll test this again without later, as we do not want to fragment or break our world i need to make a new testserver)

the second bug is related to MFR Plasticsheets:
when i have several stacks of plastic sheets in my system and left click and draw them out i get a stack of 128 plastic sheets for some unkown reason (same for rightclick, where i get 64 instead of 32)... i noticed an odd behaviour with tinkerers construct and plastic sheets too, so maybe it's not your fault at all

the funniest is when i drop the stack in my inventory i am forced to split it to regular 64 items-stacks, but when i drop the drawn stack to the ground and then loot it, it becomes overstacked
(same goes for nonstackable but stacked items i get from villagers (bee frames for example), so i believe this also is not your fault, more like forge/minecraft mechanic)

yueh:
Self cleaning via meaningful records is a bit too hacky...

Fixes #1938

### DIFF
--- a/src/main/java/appeng/util/item/UnsortedItemList.java
+++ b/src/main/java/appeng/util/item/UnsortedItemList.java
@@ -94,7 +94,7 @@ public final class UnsortedItemList implements IItemList<IAEItemStack>
 	@Override
 	public boolean isEmpty()
 	{
-		return this.records.isEmpty();
+		return !this.iterator().hasNext();
 	}
 
 	@Override


thatsIch:
yuck

PR #1954, opened 2015-10-16
Implement toString() for AppEngCore; fixes #1953

### LINKED ISSUE #1953: AppEngCore.toString() does not return ModId
071ae6d6808a701afd95f22762a9a39e8cede3d5 changed `AppEngCore.md`
into `AppEngCore.metadata` This broke the class' `toString()` method; it now returns the FML default string `"Dummy Container ("+label+") @" + System.identityHashCode(this)`

This appears in the modlist as returned by `Loader.instance().getIndexedModList()`
For example, [ServerSync](https://github.com/superzanti/ServerSync) uses this string to compare the mods installed on the client and server. Because the result of `toString` is different for each launch of Minecraft, the comparison always fails.  
In my opinion, since the metadata for the coremod is defined, `AppEngCore.toString()` should return the modID. See #1954

Nazfib:


yueh:
See the related issues for reasons.

PR #1996, opened 2015-11-11
Fixes #1970: Lighting was not re-enabled when looking at an inscriber

### LINKED ISSUE #1970: Rendering of a press in inscriber leave the GL engine in an improper state
When displaying a press, the inscriber leave the GL engine in an improper state causing ambient occlusion to not render at some angles.

Look at toolrack render differences (top-right of pictures)
Without press in inscriber:

With press in inscriber:

See also: 
https://github.com/ForestryMC/ForestryMC/issues/919

thatsIch:


yueh:
Moving the `glEnable/glDisable` one method up clearly makes more sense then my fix for the rendering cleanup in the local method.

PR #1997, opened 2015-11-11
Fixes #1995: Allows creation of version checker config file after start up

### LINKED ISSUE #1995: VersionChecker fails to start up if the config has yet to be created
`Preconditions.checkState( file.isFile() );` will fail if there is no config yet

thatsIch:


yueh:
Additionally checking for `!file.isDirectory()` does make sense for the extremely rare case that someone created a directory with the same name, thus preventing the config file from being created.

But it feels a bit hacky as it will implicitly test for `if (!file.exists() || file.isFile())` by hiding it behind `if (!file.isDirectory)`

thatsIch:
the problem is that `isFile` is not checking if it was generated as a file, but it uses `exists()` as a precondition for that. The best choice is to check if its not a directory.

yueh:
Yes, it clearly works. But my intention says something is still missing.

For example the previous check also implicitly validates that the parent directory does exist. This will no longer be the case with the change. Thus the `Configuration` itself might later fail due to not recursively creating all necessary directories.

Maybe it should be something like this?
```
Preconditions.checkNotNull( file );
Preconditions.checkState( file.getParentFile().isDirectory() );
Preconditions.checkState( !file.isDirectory() );
```

So it does still verify that the directory actually exists and a new configuration can be created later, but should for whatever reason a folder with the same name as the configuration file be present, it will still fail.

thatsIch:
I will look into that

yueh:
Probably fine for now. The parent directory can only be missing, if someone deletes it immediately after its creation and before the version checker is constructed. Should this ever occur, we can still change it.

PR #2006, opened 2015-11-16
Update fr_FR.lang

Mazdallier:


thatsIch:
Please squash them into 1 commit if possible

PR #2015, opened 2015-11-27
Update hu_HU.lang

XFRGod:


yueh:
Could you please squash it into a single commit?

XFRGod:
Sure

PR #2035, opened 2015-12-06
Update hu_HU.lang

XFRGod:


yueh:
Thanks.

XFRGod:
No problem since it really wasn't difficult. I'll probably commit again because the language file needs a little more work in areas but it's 85-90% done.

yueh:
Would be highly appreciate.

PR #2037, opened 2015-12-06
Update hu_HU.lang

XFRGod:
Okay now it's absolutely 100% done and I'm officially dead.
Onto EnderIO and Botania then!


yueh:
Thanks again for your work.

XFRGod:
Oh come on there is no need to thank me. Sure it took time but it was worth it and as far as I can see the mod will be more popular and easier to understand for those who can't speak english and that would make this worth it so much more^^

PR #2032, opened 2015-12-03
Changed NetworkMonitor to update its cache.

yueh:
**Seems to work fine**

No longer throws the full cache away with every single inventory change.
Should ensures that updates are updates are send exactly once per network. No longer duplicate ones  due to feedback from subnetworks.

For now it no longer extends `MEMonitorHandler` provided by the API. Mostly to have a clean base and not be influenced by it. Mabye it is necessary to backport some features like the `semaphore`.
But most issues seem to be specific to the `NetworkMonitor` and how it posts the updates recursively.

Looks surprisingly stable for now. I have not found any issue with wrongly updated terminals, monitors or level emitters.
But I am not completely convinced that it is working perfectly for now.

Performance improves would be pretty impressive. My testworld sees about a 95% improvment in cpu time related to inventory operations. Not much less for every AE2 related ticking.
Should also fix issues like the performance drop with mass shift clicking in crafting terminals.

NPS snapshots:
[cached networkmonitor](https://www.dropbox.com/s/k4rjymwjqvssk4e/cached-networkmonitor.nps?dl=0)
[uncached networkmonitor](https://www.dropbox.com/s/evnkfatp4kyk3qb/uncached-networkmonitor.nps?dl=0)

PR #2075, opened 2015-12-27
Fixes #2043: Storage buses updates should happen in the correct order.

### LINKED ISSUE #2043: Storage Bus disturbed by ME Level Emitter after restart
I recently came across a bug in rv2-stable-10 on my server.
However I can reproduce the bug with only AE2 installed in Single Player modus.
The problem only occurs after a reboot or new session in Single Player.

The problem is, that an ME Storage Bus does not seems to be working, when there is an ME Level Emitter with a Fuzzy Card in it installed on the same network.

This is my setup:
The ME network comprises the Storage Bus, an Energy Cell, an Terminal, and the ME Level Emitter.

A Chest with an adjacent ME Storage Bus in default settings.

In the Chest is an diamond pickaxe.

And the ME Level Emitter is configured to emit a signal if there is below 1 diamond pickaxes present, with a Fuzzy Card present.

So in theory, the level emitter shouldn't emit a redstone signal and the terminal should show the pickaxe. But after a restart (for single player, go into main menu), the terminal don't show the pickaxe and there is a redstone signal.
However, after a slight change to the Network, e.g., take away the energ cell, it works again like intended.

*edit: I forgot to mention the forge Versions:
Forge Version on my server: 1448
Forge Version on Client: 1291, 1448 and 1403, all three are affected

### LINKED ISSUE #2403: Crafting Dialog shows Red Zero as Stack Size
This may be a side-effect of the stacksize=0 hack to get our own stack size to be drawn:

Intended behavior is not showing the number at all, even if the recipe produces more than 1 output.

yueh:
This is loosely based on #2032 and should prevent any duplicated events,
but still doing a refresh of the cache due to potentially issues with some inventories.

It also ensures that storage buses will not announce a CellArrayUpdate before they are fully initialized, thus no longer causing them to return null during an update and not being added as inventory.

Fixes #2403

### DIFF
--- a/src/main/java/appeng/me/cache/NetworkMonitor.java
+++ b/src/main/java/appeng/me/cache/NetworkMonitor.java
@@ -21,34 +21,295 @@ package appeng.me.cache;
 
 import java.util.Collection;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.Iterator;
-import java.util.LinkedList;
+import java.util.Map;
 import java.util.Map.Entry;
 
+import javax.annotation.Nonnegative;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Lists;
+
+import appeng.api.config.AccessRestriction;
+import appeng.api.config.Actionable;
 import appeng.api.networking.events.MENetworkStorageEvent;
 import appeng.api.networking.security.BaseActionSource;
 import appeng.api.storage.IMEInventoryHandler;
+import appeng.api.storage.IMEMonitor;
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/me/cache/NetworkMonitor.java
```diff
@@ -21,99 +21,252 @@
 
 import java.util.Collection;
 import java.util.Deque;
+import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedList;
```
orod-org: MINOR Remove this unused import 'java.util.LinkedList'. [rule](http://sonar.orod.org/coding_rules#rule_key=squid%3AUselessImportCheck)

orod-org:
* MINOR 1 minor

PR #2079, opened 2015-12-29
Japanese translation files

Honon:
Hello!!
As we have created a Japanese translation files , and you sure you want to got to add when it is good?


thatsIch:
Does anybody wants to proof-read this? I do not know any other Japanese people personally which I can ask.

Honon:
I'm sorry...
Erase by Tier.

PR #2256, opened 2016-04-29
Applied Energistics 2 1.7.10 All blocks are invisible and xray glitched

Auroxiel:
Hello I'm having an issue with applied energistics 2, all of the blocks are invisible and xray glitching through the world. I'm not really sure how to fix it here are the mods I'm using. 
Please help me, I really wanna use this mod

capture


Cisien:
Why is this a PR? I don't understand what this is?

Issue #21, opened 2014-09-27
Logistics Pipes Integration


AlgorithmX2:
Was in AE1.

Originally Posted by AlgorithmX2

This issue was migrated; For original details see http://ae-mod.info/Tracker/0192/

Ioandar:
LP -> AE2 does actualy work fine via ME Interface...

I would very much like to use AE2 again but without LP integration I find it hard to peek/migrate (I am an old LP dog and having some trouble learning all the cool new tricks AE2 has to offer)

If you could find some time to make the request-pipe work again on the ME-Interface it would be great!

cheers and tyvm

Drethic:
Has this integration made it onto the Roadmap and/or had an estimated release date assigned to it?

thatsIch:
All features have no high priority, before not most critical bugs are not fixed. Adding new features, while other bugs are not fixed, will only lead to more inconsistent behavior and harder to track.

If you feel the need, to have this feature earlier, you are free to make a PR and suggest it to us.

Drethic:
Are there any developer docs that would point me in the right direction to start the integration?  I found what looks like the AE 1 integration comment in Applied-Energistics-2/src/main/java/appeng/transformer/asm/ASMIntegration.java on line 60.  However, I don't see where you've integrated with mods like Jabba with the storage bus.

thatsIch:
JABBA is an instance of `IDeepStorageUnit` of MFR

NEXUS2345:
Would really like to see this added sooner or later, at least the ability to request from the ME system

SharkWipf:
Took me a while to figure out it didn't work because of a bug. Guess I'll have to make room for a quantum network bridge after all.
Would love to see this (re)implemented sooner or later though.

ruifung:
Any progress on this?

loordgek:
lp is skipping 1.8
and i need it because ae2 has stuff that lp does not have and the other way around

thatsIch:
API for that

https://github.com/RS485/LogisticsPipes/blob/b616f29b749c2391bca8cf1b4a0be989bd432b18/common/logisticspipes/api/ILPPipeTile.java

since LP pipes are no BC pipes anymore, need to make own new stuff to wrap it up

shadoxxhd:
any time frame for when this will be fixed?

thatsIch:
had a somewhat working branch for that,
[...]

PR #2401, opened 2016-10-02
(Issue #21) Implement Logistics pipes integration

hg-0:
So this is a pretty big patch, based on the ideas for changing ME interfaces from @thatsIch's branch

- LP item requests act like crafting requests, but bypasses the crafting CPU unless the item is an input to an AE2 crafting recipe.
  (This should be trivial to change to require a crafting cpu for all LP requests if decided)
- Most of the external storage handler code for the interface is based on storage bus code, except it supports multiple handlers per block side
- Export busses and ME interfaces require crafting cards to export items from LP
- Partial LP requests aren't allowed, so if an export bus has acceleration cards, it will only export if it can fulfill the request
- Nothing involving LP power was integrated

Code changes that probably need special attention:
- MEMonitorIInventory changes, especially involving the onTick code to detect item changes for requestables/craftables
- DualityInterface supporting all sides for inventory handlers. Doesn't seem like an issue unless multiple large LP networks were connected to one interface. Could be
  easily changed to only support one per ME interface though
- CraftingTree/Grid/CPU additions, tried to avoid touching anything important since I didn't fully understand how everything worked, but definitely needs thorough review
- AEStack::isMeaningful change for requestable items. I didn't look into the impact this would have, since countRequestable isn't used much and it's really only for a minor reason
  (so the requestable item tooltips update in terminals for decreasing items)

Known issues
- Provider pipes could cause an infinite loop of provided->requestable->provided, but luckily LP only reports items with stacksizes right now
- Provider pipes could result in items being shown as requestable after all of an item that only existed in ae2 storage is pulled out,
  closing and re-opening the terminal refreshes that though
- LP requestables are visible on subnets, but not requestable (LP craftables are not visible)

I'm completely unfamiliar with gradle and forge mod dependencies, so I copied the logistics pipes dependecy info from @thatsIch's branch in order to build cleanly

Obviously this will need thorough testing with large scale builds, but all of my little test setups work
My testing was done with:
Forge 1.7.10-10.13.4.1448-1.7.10, Buildcraft 7.0.9, Logistics Pipes 0.9.3.126
Forge 1.7.10-10.13.4.1614-1.7.10, Buildcraft 7.1.16, Logistics Pipes 0.9.3.126

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -42,6 +42,7 @@ pneumaticcraft_version=1.9.15-105
 #########################################################
 mekansim_version=8.0.1.198
 rotarycraft_version=V6e
+logisticspipes_version=0.9-stable
 
 
 #########################################################
--- a/gradle/scripts/dependencies.gradle
+++ b/gradle/scripts/dependencies.gradle
@@ -126,6 +126,7 @@ dependencies {
     compile "appeng:RotaryCraft:${rotarycraft_version}:api"
     compile "appeng:mekanism:${minecraft_version}-${mekansim_version}:api"
     compile "appeng:InventoryTweaks:${invtweaks_version}:api"
+    compile "appeng:logistics-pipes-api:${logisticspipes_version}"
 
     // self compiled stubs
     compile(group: 'api', name: 'coloredlightscore', version: "${api_coloredlightscore_version}")
--- a/src/api/java/appeng/api/networking/crafting/ICraftingGrid.java
+++ b/src/api/java/appeng/api/networking/crafting/ICraftingGrid.java
@@ -96,6 +96,13 @@ public interface ICraftingGrid extends IGridCache
 	boolean canEmitFor( IAEItemStack what );
 
 	/**
+	 * @param what to be requested item
+	 *
+	 * @return true if the item can be requested via an interface with external crafting or storage
+	 */
+	boolean canRequestFor( IAEItemStack what );
+
+	/**
 	 * is this item being crafted?
 	 *
 	 * @param aeStackInSlot item being crafted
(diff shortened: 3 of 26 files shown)


yueh:
Thanks for your effort. But the 1.7.10 version is pretty much frozen and will not see any further work.
So any LP integration is most likely postponed until the 1.10 (or 1.11?) port is finished and LP is available for this version.

At least I want to give you some feedback based a quick glance and the potential issues I noticed. 

In general it is an API break, which means it potentially break any AE2 addon around, without any guarantee to ever be fixed with the current state of 1.7.10. This alone is pretty risky and can easily destroy the world of anyone still playing 1.7.10.

The changes to `DualityInterface` are also a bit worrying. Ticking an `MEMonitor` multiple times is certainly a bad idea. Best case it just a major performance issue to a system already plagued by them. Worst case I cannot even estimate currently.

Changing `AEStack::isMeaningful()` might also be a larger issue that it might appear. We certainly use negative stacksize already, but these should never reach anything where `isMeaningful` is important and has to be used to remove any stacks with a negative stacksize.

Crafting seems also to be massively changed and the LP integration feels way too tightly coupled to LP. Currently it would fully support using LP as inventory and use requests to account for the time delay. Which works perfectly fine with other mods already. Requesting crafting and not only storage should maybe even work with emitable items. This might stale the crafting cpu, without enough resources to fullfil the request. But certainly a better option then putting them into an infinite loop.

So for now I sadly have to reject it. Mainly because we simply do not want to spend the limited amount of time we have on adding/testing features for 1.7.10, which we then have to remove again. Instead we simply prefer to put that time towards the 1.10 port.

Regardless of this decision. We would highly appreciate any help to add LP integration once the 1.10 port is done and should you are still interested in helping us out. Or even for any other mod.

hg-0:
Thanks for the feedback. If LP integration is still desired after 1.10 is done I'll definitely check back on this.

shartte:
Do you know of the status for LP on 1.10, @hg-0 ?

hg-0:
I don't, sorry. I'm not affiliated with the project

KamranMackey:
@shartte It's still only for 1.7.10 unfortunately.

PR #2469, opened 2016-10-16
Added a comment for the client configuration.

KamranMackey:
Also made some misc other changes such as converting the Client configuration part in AEConfig to a forEach statement. I checked to make sure everything worked ok with this change. Everything worked fine.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,4 +1,4 @@
-#Fri Sep 16 14:56:32 CEST 2016
+#Sat Oct 15 22:59:17 MDT 2016
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
--- a/src/main/java/appeng/core/AEConfig.java
+++ b/src/main/java/appeng/core/AEConfig.java
@@ -392,15 +392,13 @@ public final class AEConfig extends Configuration implements IConfigurableObject
 	@Override
 	public void updateSetting( final IConfigManager manager, final Enum setting, final Enum newValue )
 	{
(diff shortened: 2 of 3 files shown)


shartte:
Wait, what is this for actually? Is this for the "Mod options" feature in Forge?

If so, we'd need to make the description translateable using the current .lang files.

KamranMackey:
@yueh it's not a "generic comment". And you guys need to fix your config so that way it's easier for people to understand. You just can't name categories without adding a tooltip comment. That's bad coding, and makes it really hard for people to understand what each category contains. Heck, for all we know one of the categories could contain a setting which destroys your entire Minecraft installation (impossible, but just an example). 

And oops, I forgot that that's what forEach does. It was a inspection done by IDEA which said it could be converted to a forEach statement. My bad!

KamranMackey:
@yueh And I originally went with a comment that said "Configure settings which affect AE2's functionality client-side" but that seemed too silly so I changed it.

yueh:
That you considered the correct comment as silly is actually a good thing. Because that is the reason why the comment is useless.

If you are already editing a **configuration file** to **configure** something and have a **client section** to **configure** the **client**, then adding a generic comment about `# configure the client` is one of the best examples for bad programming (even if that is not programming at all). Someone who cannot make the trivial connection between configuration file + client section, will not really find a comment stating the obvious very beneficial. Or worst case even messing it up regardless of how many lines of documentation it contains. Like if it states `only 0 or 1`, you can expect them to use `-1` as option.

A good comment has to document non-obvious behaviour. Say the range of possible values and what happens when the value exceeds this range. Like does it use `min()/max()`, falls back to the default value or will simply crash.
Or information about `these options can also be changed ingame without restarting the client`, potential side effects it causes and so on.

PR #2511, opened 2016-10-26
Made Memory card messages easier to understand for newcomers.

DeadSix27:
The old ones were a little too generalized and a bit hard to
understand for people that never played with AE.


src/main/resources/assets/appliedenergistics2/lang/de_DE.lang
```diff
@@ -73,17 +73,17 @@ tile.appliedenergistics2.sky_stone_slab.name=Himmelssteinstufe
 
 // Chat Messages
 chat.appliedenergistics2.ChestCannotReadStorageCell=ME-Truhe kann die Speicherzelle nicht einlesen.
-chat.appliedenergistics2.SettingCleared=Einstellung gelöscht.
+chat.appliedenergistics2.SettingCleared=Speicherkarte geleert.
 chat.appliedenergistics2.OutOfRange=Drahtloser Zugangspunkt außer Reichweite.
 chat.appliedenergistics2.InvalidMachine=Ungültige Maschine.
-chat.appliedenergistics2.LoadedSettings=Einstellungen erfolgreich geladen.
+chat.appliedenergistics2.LoadedSettings=Geräte einstellungen von Speicherkarte geladen.
```
Skaronator: *Geräteeinstellungen

DeadSix27: My German isn't that good (anymore) :) Thanks

shartte:
I am not really satisfied with the memory card feedback in general.

If you think about it, linking P2P tunnels together gives you no usable feedback at all, and calling it "settings" is also super misleading.

DeadSix27:
Indeed, would require code changes to give feedback as to what has been linked and saved to memory card. E.g "P2P Bus (ME) configuration saved to memory card".. "Created P2P Link (ME) between Red Cable at (x,y,z) and (x,y,z) on Red Cable Line"  (or so) would be the best, but a simple translation change for the beginning still seems like an easy thing to add until further development has been done. I could even take a look into how that could be done, altho as I mentioned way back I never worked with Forge plugins (but with Java) and I'd have to first read myself into AE code in general anyway.

shartte:
Yeah, it's probably easy to change on the code side, the question is how it should work in the first place.
In general as can be seen in #2522, we have some difficulty with the current way of even setting the type of P2P tunnels, so I am pondering a bit on how to proceed here. 
Especially interesting is the fact that memory cards also set the tunnel-type on the target tunnel, which somehow makes it appear as "just a setting" which in my eyes shouldn make it unnecessary to rightclick it with a specific item to set the type. 

Regarding your PR:

> Loaded device configuration of memory card.

Should probably be Loaded device configuration *from* memory card.

Also in the english translation you changed, you might correct the capitalization of Changed :)

DeadSix27:
:) will do I'll also play around with some ideas in the memory card code later/tomorrow and see what I come up with, maybe you can work upon what I can put together and enhance it.

shartte:
The code you want is in PartP2PTunnel, methods onPartActivate for right-click and onPartShiftActivate for shift+right click.

DeadSix27:
Thanks, saves me some searching :P Also, what about the capitalization? Couldn't find any usage of capitalized "Changed" in the lang file however "Cleared" had a capital C, even in the original lang file (prior to my edits), if that is whats you meant.

leagris:
It also raise the following question:
Why have multiple tunnel types or expose this to the player in the first place?
In reality a tunnel should be neutral and tunnel whatever is at input to whatever is at output, be it ME network, whatever energy, light, redstone…
The most useful and missing information that player needs is the location and dimension of the input side, and probably the network tool should let player list all input and outputs while probing any tunnel part like EnderIO list all connexions to a conduit.

shartte:
Err yes, I meant cleared. 

@leagris: AE takes a tax on energy connections though 💰

leagris:
> @leagris: AE takes a tax on energy connections though 💰 
@shartte get the tax paid in AE energy (as it has always been). Tax calculation may be Dependant on the tunnel content/usage but this does not need to be exposed to the Player/End-User. It would be the internal job of AE to check what is tunneled and tax appropriately.

DeadSix27:
Another idea for the far far future could be a configuration tool, ditching the memory card, like a tool to click the bus that opens the Config UI which allows you to set the type and see both P2P tunnel's in it, whether they're actually connected via a cable or not etc. 
But then again that could possibly just be done by WAILA showing the 2 Linked P2P buses and their location in the WAILA overlay and making the memory card to the same as before and add a right click small config GUI to P2p busses to set their type

But that sounds like a big overhaul in the way later stages.

(3 commits to change 4 lines, I better stick to coding instead of language files ;) )

DeadSix27:
@shartte Is there any existing lang-file formatting functions in AE? E.g 
chat.appliedenergistics2.test=foo {bar2} bar
and then simply string replacing the placeholders later on

yueh:
We already considered that. But that might actually be difficult here. Currently the server sends the message as the client simply never not know, if it is successful. But that will make it difficult to translate it.

DeadSix27:
since I came to that problem now, I made a basic linking information code that just gives feedback that you linked P2p bus x,y,z to bus x,y,z but ofc its only english.

PR #2517, opened 2016-10-27
RU: Update Translate!

xMrVizzy:


yueh:
Thanks. Looks like the search&replace does not look unicode.

PR #2588, opened 2016-11-06
Gradle Source Set Changes

shartte:
Introduced separate source sets for "api" and "apis" (in preparation for source only 3rd party APIs).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -55,6 +55,9 @@ jar {
         attributes 'FMLCorePluginContainsFMLMod': 'true'
     }
 
+    from sourceSets.api.output
+    dependsOn apiClasses
+
     // specify which files are really included, can control which APIs should be in
     include "appeng/**"
     include "assets/**"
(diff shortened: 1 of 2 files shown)


shartte:
@yueh I've separated out the api folder as a real source set, and added an "apis" source set to include stuff like the COFH RF API in. What's your take?

yueh:
The changes to `main` and `api` should be fine. The `resources` might be useful, if I recall correctly, the build script will use these to assemble the final jar. Which gives us more control over it. Especially with curse whining about accidentally bundling a gradle wrapper or probably stuff like the `Thumbs.db` from windows...

The `apis` one is really hard. The easy to fix problem is the name. `api` and `apis` is way too easy to mix up. If it should be something like `3rdparty`. But my main issue is with having to manually deal with dependencies. It is not just a change version number inside the `gradle.properties`, it means fetching the source, probably having to strip the api classes from it. So it could be pretty tedious. Some mods might not even release the source files, just but a random jar somewhere, so that would require yet another approach for these ones. Maybe even having to assemble an api jar by hand, because there is only the full mod, but the author also complaining about hosting it somewhere else for easy integration into the build system...

Getting a free CI and maven repo these days is so easy. In most cases there are even copy&paste scripts for gradle available to deploy it automatically.

Also there are a couple of maven proxies around, which combine some of the maven repos for minecraft mods. Iirc there should also be a jar for RF around.

shartte:
Regarding the resources assembly: Standard packaging is to take everything from src/main/resources. 
I think it's a trade off between being intransparent about what is being included and what is not. I'd rather follow the general convention that gradle uses here.

apis -> 3rdparty is fine by me.

I tried finding a good jar for the RF API, although I didn't really search that much. Looking at mods like Ender IO and others, it seems pretty common to just use the source from the RF API GitHub.

Issue #2509, opened 2016-10-26
SpongeForge API Crash 


YOYOK9:
#### Description
So not sure whether to report this to you or sponge powered (Forge) but it seems that there is an incompatibility with the 2 mods. 
From what it looks like when I add AE2 something loads prematurely and the server false to start. (Takes about 10 seconds to crash)

#### Environment
multiplayer server. 
Using spongepowered API for plugins. 
Running latests of SpongeForge, Forge, and AE2 nightly build
http://pastebin.com/eNiLEPit

- Sponge Forge: spongeforge-1.10.2-2107-6.0.0-BETA-1842
- Minecraft Version: 1.10.2
- AE2 Version: -rv4-nightly-923517
- Forge Version: 12.18.2.2114

Please let me know.

yueh:
Known, but will not be fixed before reaching a more stable state.

PR #2483, opened 2016-10-17
Split DummyModContainer from IFMLLoadingPlugin

gabizou:
This is the necessary (and tested) changes to fix #2482.

What is happening otherwise is that `ModMetadata` is being class loaded before other `IFMLLoadingPlugin`s are able to register their necessary transformers. This specifically fixes SpongePowered/SpongeForge#966.

PR #2604, opened 2016-11-12
Split-off coremod and mod classes

Prototik:
Split-off coremod and mod classes due to classloading errors with SpongeForge.

Issue reported in #2509, first revision of PR (#2483) was rejected due to naming conventions, so I tried to keep old coremod name and give `appeng.transformer.AppEng` to mod class.


src/main/java/appeng/transformer/AppEng.java
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with Applied Energistics 2.  If not, see <http://www.gnu.org/licenses/lgpl>.
+ */
+
+package appeng.transformer;
+
+
+import appeng.core.AEConfig;
+import com.google.common.eventbus.EventBus;
```
orod-org: MINOR Wrong order for 'com.google.common.eventbus.EventBus' import. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck)

src/main/java/appeng/transformer/AppEng.java
```diff
@@ -0,0 +1,82 @@
...
+import net.minecraftforge.fml.common.DummyModContainer;
```
orod-org: MINOR 'net.minecraftforge.fml.common.DummyModContainer' should be separated from previous imports. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck)

orod-org:
SonarQube analysis reported 2 issues
* MINOR 2 minor

src/main/java/appeng/transformer/AppEng.java
```diff
@@ -0,0 +1,83 @@
...
+ * along with Applied Energistics 2.  If not, see <http://www.gnu.org/licenses/lgpl>.
+ */
+
+package appeng.transformer;
+
+import com.google.common.eventbus.EventBus;
+
+import appeng.core.AEConfig;
+
+import net.minecraftforge.fml.common.DummyModContainer;
```
orod-org: MINOR Wrong order for 'net.minecraftforge.fml.common.DummyModContainer' import. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck)

orod-org:
SonarQube analysis reported 1 issue
* MINOR 1 minor

Prototik:
@yueh any comments why it's closed?

Prototik:
@yueh @shartte anyone?

Cloudhunter:
I've been using the auto build from the PR for two days now - appears to working well. I'd much love for this to be merged into the mainline, so that I don't have to use an unofficial build just to keep Sponge!

Prototik:
So heated discussion...

shartte:
See #2636 
If you want to contribute, you can help review that...

KasperFranz:
@shartte but it didn't get implemented with that, and making that still a problem.

@Prototik  tried to fix it with #2670 

You guys are really making it hard for everyone :/, since we have to use a custom fork which includes this or similar fixes - since not running sponge is not really an option when having a large server.

yueh:
It is still just a bandaid not a fix. Forge is even considering getting rid of coremods completely. So we might have to replace it completely. It should be doable to dynamically generate bytecode for things like `TileEntity` as they have a factory method. But that might not apply to every use case we need. So worst case we have to use `LaunchWrapper` or inject our own classloader somehow, etc.

So now is your chance to actually provide some constructive feedback about how we can remove the coremod completely, while still maintaining the same feature set. Instead of just complaining that you have to put a few minutes of effort into applying a patch and run `./gradlew build`.

Because once we find and choose an approach, we will most likely stick to it as it will probably be not a trivial effort to get it working. Which can also mean it might no longer be compatible with sponge at all. Not just some random loading of forge classes (which can be trivially fixed).

At that point it will be more or less "The issue was open for a long time. You did not raise any concerns  about whatever solution or suggest an approach. We consider that as tacit approval. Now deal with it."

I doubt anyone of you will appreciate a solution, which pretty much cannot be reverted anytime and completely break it for you (neither do we). But applying the band-aid now means you can ignore it and have to face the actual solution in a few weeks or months. Whatever it will be.

mcsnetworks:
Correct me if I am wrong but a duct tape fix is better than telling your server user base to use a custom fork or recompile it yourself isn't it?

It's my understanding that the patch the SpongeTeam recommended didn't interfere with any subsystem.  Feel free to correct me if I am making the wrong assumption.

PR #2651, opened 2016-11-22
Create en_NO.lang

016Nojr:
WIP


atakle:
Shouldn't the language code be nb_NO? Also, this pull request has _a lot_ of errors, and pretty much looks like it was machine translated. If I tried playing with this, I would turn mad within minutes.

PR #2840, opened 2017-03-18
Only send filter to jei when it's changed.

Randomfish132:
Current JEI synchronization resets the search and page anytime the monitor is updated. This changes that to act more like the old NEI sync.

[Video of current functionality.](https://youtu.be/Nyy3LAbKE0k)


Randomfish132:
http://files.minecraftforge.net/ was down when these ran, don't know how to have them re-run.

yueh:
I see the issue, but I think we should reject this approach in favor of an two way synchronized approach.

Not perfectly ideal as we cannot observe the JEI filter for an actual change, but copying the JEI filter each tick back into our field once they differ might be enough. Or maybe request a new event from JEI if the input changes. But not sure if 1.10 will see new features.

PR #2876, opened 2017-04-16
Check for contained items instead of removability

### LINKED ISSUE #681: Ender IO Alloy smelter does not follow Interface blocking rules
The Interface block will continue to add items to the Ender IO Alloy Smelter's input even if the blocking mode is set to "Do not push crafting items if inventory contains items" and there are already items present. The Interface block should be unable to add items to the input if this mode is set and there are items present. This issue is crossposted to the Ender IO github.

Using AE2 rv2 beta 1 and Ender IO 2.2.5.311 with Minecraft 1.7.10 and Forge 10.13.2.1230. Issue occurs on a multiplayer server.

TomyLobo:
Fixes #681

### DIFF
--- a/src/main/java/appeng/helpers/DualityInterface.java
+++ b/src/main/java/appeng/helpers/DualityInterface.java
@@ -947,7 +947,7 @@ public class DualityInterface implements IGridTickable, IStorageMonitorable, IIn
 			{
 				if( this.isBlocking() )
 				{
-					if( ad.simulateRemove( 1, null, null ) != null )
+					if( ad.containsItems() )
 					{
 						continue;
 					}


orod-org:
SonarQube analysis reported 16 issues
* MAJOR 14 major
* INFO 2 info

#### Top 10 extra issues

Note: The following issues were found on lines that were not modified in the pull request. Because these issues can't be reported as line comments, they are summarized here:

1. MAJOR [DualityInterface.java#L335](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L335): Method appeng.helpers.DualityInterface.updateCraftingList() needlessly boxes a boolean constant [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3ANAB_NEEDLESS_BOOLEAN_CONSTANT_CONVERSION)
1. MAJOR [DualityInterface.java#L335](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L335): Method appeng.helpers.DualityInterface.updateCraftingList() appears to call the same method on the same object redundantly [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3APRMC_POSSIBLY_REDUNDANT_METHOD_CALLS)
1. MAJOR [DualityInterface.java#L484](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L484): Method appeng.helpers.DualityInterface.addToCraftingList(ItemStack) appears to call the same method on the same object redundantly [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3APRMC_POSSIBLY_REDUNDANT_METHOD_CALLS)
1. MAJOR [DualityInterface.java#L534](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L534): Method appeng.helpers.DualityInterface.gridChanged() appears to call the same method on the same object redundantly [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3APRMC_POSSIBLY_REDUNDANT_METHOD_CALLS)
1. MAJOR [DualityInterface.java#L545](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L545): Remove this unused method parameter "dir". [rule](https://sonar.orod.org/coding_rules#rule_key=squid%3AS1172)
1. MAJOR [DualityInterface.java#L568](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L568): Remove this unused method parameter "side". [rule](https://sonar.orod.org/coding_rules#rule_key=squid%3AS1172)
1. MAJOR [DualityInterface.java#L768](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L768): Method appeng.helpers.DualityInterface.handleCrafting(int, InventoryAdaptor, IAEItemStack) orders expressions in a conditional in a sub optimal way [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3ASEO_SUBOPTIMAL_EXPRESSION_ORDER)
1. MAJOR [DualityInterface.java#L905](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L905): Class appeng.helpers.DualityInterface defines List based fields but uses them like Sets [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3ADLC_DUBIOUS_LIST_COLLECTION)
1. MAJOR [DualityInterface.java#L1047](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L1047): Method appeng.helpers.DualityInterface.provideCrafting(ICraftingProviderHelper) orders expressions in a conditional in a sub optimal way [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3ASEO_SUBOPTIMAL_EXPRESSION_ORDER)
1. MAJOR [DualityInterface.java#L1103](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/98a4f1304cc0a9cf60140c57c66058a4949ddc39/src/main/java/appeng/helpers/DualityInterface.java#L1103): Method appeng.helpers.DualityInterface.getHost() appears to call the same method on the same object redundantly [rule](https://sonar.orod.org/coding_rules#rule_key=fb-contrib%3APRMC_POSSIBLY_REDUNDANT_METHOD_CALLS)

TomyLobo:
Those sonar messages aren't in my code. I only changed line 950.

yueh:
Sorry, but this will completely break blocking mode and change it from not working for a few exceptions to not working with pretty much everything.

TomyLobo:
This patch fixed #681 for us.

We tested, in blocking mode:
* vanilla chests (for our empowerer automation)
  * patched behaviour: works
  * rv4-alpha-11 behaviour: works
* Ender IO alloy smelters
  * patched behaviour: works
    * Also works with Octadic Capacitors
  * rv4-alpha-11 behaviour: same as non-blocking mode
* Ender IO sag mills *without* flint/dark steel ball
  * patched behaviour: works
    * Also works with Octadic Capacitors
  * rv4-alpha-11 behaviour: same as non-blocking mode
* Ender IO sag mills *with* flint/dark steel ball
  * patched behaviour: Blocks, until I remove the flint/dark steel ball
  * rv4-alpha-11 behaviour: same as non-blocking mode
* ExtraUtils2 Resonator
* IC2 block cutting machine
  * Blocks, until I remove the sawblade
  * rv4-alpha-11 behaviour: same as non-blocking mode
* IC2 machines with overclocker upgrades
  * Blocks, until I remove the overclockers
  * rv4-alpha-11 behaviour: same as non-blocking mode

One could regard these as regressions, but I would argue that having any kind of blocking functionality is preferable to having two modes that essentially do the same for everything but chests.
There's also a simple workaround to get the rv4-alpha-11 behaviour: change to non-blocking mode :)

TomyLobo:
> Sorry, but this will completely break blocking mode and change it from not working for a few exceptions to not working with pretty much everything.

Can you name an example of "pretty much everything" that is not some kind of storage container, like a chest?

TomyLobo:
@yueh do you even read this?
I want to adjust my fix to encompass the machines you're referring to, but your comment is too vague to do that.
I have an experimental patch that I'm currently testing against all the previous test cases, but if you don't tell me which mods I supposedly broke, I can neither verify that nor fix the issue.

TomyLobo:
Alright, tests conclude that for IC2, EIO and EU2, behaviour is definitely improved.
For vanilla, behaviour is as before.
That's 3:0 for my patch, now i'd like to see your examples :)

yueh:
It does not need any further example. You already showed that it is a major regression for a majority of your tested machines.

Ommina:
I guess I'm not understanding.

The assertion was made that this change would result in blocking mode "not working with pretty much everything".

Actually testing suggests otherwise, and, indeed, seem to show blocking mode working with a greater number of machines than before.  This strikes me as a step forward toward making the interface work "as advertised".  What am I missing?

yueh:
You cannot pick simply pick the one nice case and ignore every other broken one completely.

If you sumarize the behaviour including the unfavourable ones, it is as follows:

* vanilla chests (for our empowerer automation)
  * no change
* vanilla furnaces (no fuel in the pattern, ME interface on top extraction pipe on bottom)
  * always blocking, thus never crafting
* Ender IO alloy smelters
  * might be changed
* Ender IO sag mills
  * always blocking, thus never crafting
* ExtraUtils2 Resonator
  * might be changed
* IC2 machines  
  * always blocking, thus never crafting

Neither approach is perfect. The current one is simply with vanilla as "gold standard" in mind. Thus furnaces and chests and they work perfectly. You can argue however you want, but vanilla behaviour is pretty much the only standard we have. Regardless of considering it broken to be able to extract from an input. (Maybe just because you designed your pipes in a way that it otherwise breaks down horribly.)

Also it fails gracefully and while not working at best efficiency at least it ensures it is always working. Not completely breaking down your crafting setup due to an apparently unrelated change like finally being able to use some machine upgrades.

TomyLobo:
Let me push that updated patch real quick, now that it's burned in a bit on our server...

yueh:
Changing `InventoryAdaptor` in this way is pretty much obsolete as it is already sided. There could be a bug with `WrapperMCISidedInventory` but the whole case is pretty much unused. The whole `IInventory` path should go away.

In other words it changes the signature of `InventoryAdaptor` in a way indicating a completely different behaviour as it is actually, while still not fixing anything because nearly no machine uses `IInventory` or `ISidedInventory` anymore.

TomyLobo:
Let me start out by correcting a few things you misread in my post:
1. sag mills work nicely if you dont use a grinding ball (or make the grinding ball part of the recipe)
2. vanilla furnaces work fine with all 3 versions, no caveats that i'm aware of.

To reply to your most recent post, I've been told IItemHandler is what enderio uses, so I was essentially doing nothing with regards to the alloy smelter with my 2nd change :)
I still think that change to the ISidedInventory wrapper is needed, though, in order to support legacy machines in a side-aware way.

To address the "graceful fallback" argument, I've had that in mind even when I first posted this, but seeing as it's a major improvement on everything but a few broken mods, I think all in all it's a step forward.
It's better to hit an issue early on and have a way to work around it than to have certain machines that you just cannot automate without potentially building as many of them as there are recipes.

For example, right now the alloy smelter cannot be easily automated with an ME interface directly on the machine, due to recipes intersecting with one another.
You'll get your machines randomly stuck in ways that are hard to diagnose.
And once you figure out what's wrong, you basically have to play a game of recipe sudoku to avoid collisions.

With a chest in the middle (as suggested on #681) the situation seems to be improved, but there are still possible collisions, since the ME interface still doesn't know whether the alloy smelter is empty or not and will continue to push possibly colliding ingredients.

This situation needs a fix and it's either the AE2 and EnderIO teams who needs to find it.
Your teams consist of very active and very diligent people from what I gather, but I have yet to see evidence of you two getting together to address the problem, just blame ping-pong.

There are a number of ways this could be addressed:
1. The change outlined in the [improved version of this PR](https://github.com/AppliedEnergistics/Applied-Energistics-2/compare/master...TomyLobo:fix-blocking-mode)
2. A 3rd mode for the ME Interface which works like 1.
    * Current text: Do not push crafting items if inventory contains items.
    * New text for the current mode: Do not push crafting items if inventory contains _requestable_ items.
    * Text for the new mode: Do not push crafting items if inventory contains _any_ items.
3. Maybe someone can find a good heuristic to tell which slots are inputs slots
    * I have no idea what that would look like, seeing as those are just numbers and maybe some pull/push behaviours...
4. Maybe you could implement a Forge proposal for this
    * People on IRC suggested that this is probably outside the scope of Forge
    * It would also not help with existing mods
5. A library mod that authors
    * I feel that this is waaaay beyond my capabilities to correctly implement and promote
    * As the previous option, this would also not help with existing mods

I think option 2 is the most realistic one here.
Since this would be a significant amount of work, I'd like to know in advance, if you'd even consider such a solution.

One final note, having your mod depend on other mods reproducing a quirk in vanilla furnaces (which were written with only hoppers in mind) can't be the solution and as you can see.
If all the Minecraft mod community did was refer back to vanilla, we probably wouldn't even have had BuildCraft pipes.

yueh:
> sag mills work nicely if you dont use a grinding ball (or make the grinding ball part of the recipe)

The problem here is, we simply cannot say you cannot use any machine upgrade at all. Otherwise it breaks horribly. Autocrafting has to work, always. Making it more efficient for some cases on top is just a bonus.

> vanilla furnaces work fine with all 3 versions, no caveats that i'm aware of.

Furnaces usually expose their fuel slot as non extractable slot (except if it is an empty bucket or so). Thus it will always be blocking. And no "make the fuel part of the pattern" is not a solution.

> EnderIO stuff

The main problem is that enderio machines are extremly broken compared to any other mod. Like they eat the ingredients immediately without a chance for a player to correct a mistake. Which also means their are always empty in this case. Which is caused by their pipes, because they are in and out by default and would otherwise just steal the materials before it could craft anything. Same with the input and output on the same side, which nearly no other mod does (at least to that extend).

Thus enderio will never be fully compatible to AE2 as it follows completely different design ideas, which are somewhat mutual.


> 1. The change outlined in the improved version of this PR

As already said, that is simply not possible due to breaking the behaviour in a very random way.

> 2. A 3rd mode for the ME Interface which works like 1.

Will be a nigthmare to actually explain as it is really opaque how machines expose their slots.

> 3. Maybe someone can find a good heuristic to tell which slots are inputs slots

Impossible. Pretty much every other mod will work in the exact opposite way. It will essentially mean we need a unique handler for every single mod out there.

> 4. Maybe you could implement a Forge proposal for this

This is actually the only correct solution. Forge needs to either patch tileentities to provide something like a `isBusy()/isIdle()` or a capability based approach. Of course a bit tricky for passive inventories. But it could use something like `te.isBusy() || !inv.isEmpty()`

> 5. A library mod that authors

Would actually be pretty easy. Like 4 just a cap to signal a machine is busy or idle. Very straightforward but it would shove AE2 as a depencendy into nearly every mod and force their devs into spending time to just get it to work. Not a good decision.

TomyLobo:
Odd, I would have thought that you understand why alloy smelters don't work with blocking mod.
Let me explain it to you again (numbering unrelated to your post):
1. alloy smelters do *not* always present their inventory as empty. They merely eat the first set of ingredients immediately. If there's any left over, it *will* be visible to containsItems.
    All that does for automation with blocking mode is that you have a backlog of one further item if you cancel the job. Not a big deal in my opinion.
    As for humans, you're correct, the item is gone as soon as you start processing it. Makes sense if you think about it. Loss of items being processed is not an issue either, since unlike many machines, enderio machines store their state when you pickaxe them.
2. Like most machines, EnderIO machines will not allow their input slots to be pulled from.
     * Not even your molecular assembler allows input extraction
     * The inscriber seems to be completely broken in that regard:
        It doesn't obey your "gold standard" (since only the top input can be pulled from)
        It doesn't obey the quasi-standard I've come to expect from a mod (since the top input slot where the press usually is can still be pulled from, making automation with pipes impossible without a filter)
     * Before you ask: I tried all of this from all sides.
3. EnderIO item conduits are *not* in+out by default. By default they will be extract-only, activated by redstone, meaning they will not do *anything* unless you either reconfigure them or apply redstone.

I hope this clears up some misconceptions on your side.

Now to your replies to my points:
1. I understand that you feel that way, Let's look for an alternative solution.
2. I provided new explanation text for both blocking modes. Can you explain how my explanations don't explain the modes to be explained?
3. I completely agree.
4. Sounds like we're making progress :)
An explicit cap would probably be the best choice.
Otherwise you'll get regressions everywhere.
I'm thinking of something like ``te.hasCap(BUSY_BEAVER) ? te.isBusy() : currentImplementation()``
Note, I have no idea how these cap things work, so this is pseudocode :)

nodar86:
Is there any update regarding to this issue?

PR #2933, opened 2017-06-18
Inscriber ISidedInventory fix

fscan:
Had problems with Thermal Dynamics item ducts sending to many items and the overflow getting dropped on the floor. This PR should fix this.

### DIFF
--- a/src/main/java/appeng/tile/misc/TileInscriber.java
+++ b/src/main/java/appeng/tile/misc/TileInscriber.java
@@ -294,6 +294,20 @@ public class TileInscriber extends AENetworkPowerTile implements IGridTickable, 
 	}
 
 	@Override
+	public boolean canInsertItem( final int slotIndex, final ItemStack insertingItem, final EnumFacing side )
+	{
+		if( !this.isItemValidForSlot( slotIndex, insertingItem ) )
+		{
+			return false;
+		}
+		if( getStackInSlot( slotIndex ) != null )
+		{
+			return false;
+		}
+		return true;
+	}
+	
+	@Override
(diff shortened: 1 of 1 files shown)


yueh:
This is not really a fix, more of a hack.

The main issue is that most tiles still share an `ItemHandler` as well as an old `I(Sided)Iventory`. While the later one should really be removed (an actual fix). As it might not be limited to inscribers but pretty much every other inventory in some way as they all share the same inventory code with through one of the superclasses (chargers, interfaces, etc)

fscan:
I don't know enough of the internals of AE2 to tell if the ISidedInventory is needed internally somewhere. To be honest the inheritance tree is a bit of a mess at first glance. 
So is your plan to remove the interface from the blocks? If not, a working implementation is better than a defect one, IMHO.
BTW, TD fixed ducts/retrievers to use IItemHandler, but vanilla hoppers seem to use the ISidedInventory interface if present.

yueh:
Vanilla hoppers worked fine in 1.7.10 via `IInventory`. So either there is something broken with them. As it is forge I would not be surprised, should they prefer `IInventory` for scanning the inventory, `IItemHandler` for simulate and `IInventory` again for the actual extraction. Due to a forgotten patch or a patch applied to the wrong position, whatever.

Or something with `I(Sided)Inventory` changed between 1.7.10 and 1.10.2. If so, then it applies to more than just the inscribers. Or inscribers breaking the normal inventory behaviour somhow (just guessing here).

But it is still a quick'n'dirty hack, which probably solves it for now. Until the real bug bites back and eats half of the items inserted into any interface...

Bandaids are never a solution, they only cause you to ignore and forget the real bug until it resurfaces in a more dramatic way.

PR #2935, opened 2017-06-23
Improve behaviour of stacks that have 0 stackSize and non-zero requestable count

thiakil:
**Preface:** I'm not 100% sure if this is how requestable stacks are supposed to work; there did not seem to be much that actually uses requestable stacks - something to do with power usage? In my addon mod I've been using them for items that don't technically "exist" until they are created (on extraction); they are items that should not (for example) be transferred by the IO Port to/from Item Cells.

Without these changes such stacks appear as craftable in the ME Terminal, but are unable to proceed past the Craft Confirm GUI, and also unable to be used in autocrafting.

With these changes the ME Terminal will not display Craft text when requestable count > 0. Stacks with 0 stack size and requestable count > 0 will display no number in the slot. Crafting calculation can successfully use the requestable stacks as if they existed.

### DIFF
--- a/src/main/java/appeng/client/gui/AEBaseGui.java
+++ b/src/main/java/appeng/client/gui/AEBaseGui.java
@@ -453,7 +453,7 @@ public abstract class AEBaseGui extends GuiContainer
 					action = ( mouseButton == 1 ) ? InventoryAction.SPLIT_OR_PLACE_SINGLE : InventoryAction.PICKUP_OR_SET_DOWN;
 					stack = ( (SlotME) slot ).getAEStack();
 
-					if( stack != null && action == InventoryAction.PICKUP_OR_SET_DOWN && stack.getStackSize() == 0 && player.inventory.getItemStack() == null )
+					if( stack != null && action == InventoryAction.PICKUP_OR_SET_DOWN && stack.getStackSize() == 0 && player.inventory.getItemStack() == null && stack.getCountRequestable() == 0 )
 					{
 						action = InventoryAction.AUTO_CRAFT;
 					}
--- a/src/main/java/appeng/client/render/StackSizeRenderer.java
+++ b/src/main/java/appeng/client/render/StackSizeRenderer.java
@@ -53,7 +53,7 @@ public class StackSizeRenderer
 			final boolean unicodeFlag = fontRenderer.getUnicodeFlag();
 			fontRenderer.setUnicodeFlag( false );
 
-			if( is.stackSize == 0 )
+			if( is.stackSize == 0 && ( aeStack == null || aeStack.getCountRequestable() == 0 ) )
 			{
 				final String craftLabelText = AEConfig.instance().useTerminalUseLargeFont() ? GuiText.LargeFontCraft.getLocal() : GuiText.SmallFontCraft.getLocal();
 				GlStateManager.disableLighting();
--- a/src/main/java/appeng/crafting/MECraftingInventory.java
+++ b/src/main/java/appeng/crafting/MECraftingInventory.java
@@ -95,6 +95,13 @@ public class MECraftingInventory implements IMEInventory<IAEItemStack>
 		}
 
 		this.localCache = this.target.getAvailableItems( new ItemListIgnoreCrafting<>( AEApi.instance().storage().createItemList() ) );
+		for( IAEItemStack is : this.localCache )
+		{
+			if( is.getCountRequestable() > 0 )
+			{
+				is.incStackSize( is.getCountRequestable() );
+			}
+		}
 
 		this.par = parent;
 	}
@@ -134,8 +141,13 @@ public class MECraftingInventory implements IMEInventory<IAEItemStack>
 		}
 
 		this.localCache = new ItemListIgnoreCrafting<>( AEApi.instance().storage().createItemList() );
-		for( final IAEItemStack is : target.getStorageList() )
+		for( IAEItemStack is : target.getStorageList() )
 		{
+			if( is.getCountRequestable() > 0 )
+			{
+				is = is.copy();
+				is.incStackSize( is.getCountRequestable() );
+			}
 			this.localCache.add( target.extractItems( is, Actionable.SIMULATE, src ) );
 		}
 
@@ -177,6 +189,13 @@ public class MECraftingInventory implements IMEInventory<IAEItemStack>
 		}
 
 		this.localCache = target.getAvailableItems( AEApi.instance().storage().createItemList() );
+		for( IAEItemStack is : this.localCache )
+		{
+			if(is.getCountRequestable() > 0)
+			{
+				is.incStackSize( is.getCountRequestable() );
+			}
+		}
 		this.par = null;
 	}
 


yueh:
No, requestable items have to exist or at least the count is guaranteed to be available when being requested, but actually are crafted. Which of course also includes any ingredient to not be advertised to the network at the same time. Say you cannot have 1 log requestable and at the same time make smelting it to charcoal and craft it to planks requestable.

The main idea is mostly for something like logistic pipes. Stuff is stored somewhere, but not available immediately. (Of course LP never really added the support for it). Iirc pneumaticcraft was pretty much the only one using it. As reference: https://www.youtube.com/watch?v=3m9QT00X4MA

So no, you cannot use it for items which are generated out of thin air. At least not without previously knowing the amount you can provide. Which of course is always a finite amount and once a player request a larger amount then available, it a cannot use it for autocrafting and changing it breaks the intended functionality.

thiakil:
I'm not creating items out of thin air really, it is a conversion from another stack type that so happens to be inside the same ME network, just the crafting system does not know how to handle it on its own. 

Using the pneumaticcraft example, does that supply a stack that has a stack size equal to the requestable count and just returns nothing on extractItems? (until crafted? How does that part work?) 

Doing that could work for my purposes, but seems it may still be extracted by the IO Port, which is undesired as the actual stacks are handled separately.

thiakil:
Also my changes don't allow extraction more than what is actually available, that is only violated if you supply a stack that has a non-zero size and non-zero requestable, and then don't check that in the extraction process

yueh:
They are not extractable. That is the whole point of being requestable. So something like `don't allow extraction more than what is actually available` is completely void. It is not about that "prevent dupes" is enough. The whole point of extracting them is wrong. Otherwise the whole requestable could be removed.

Requestables are a bit of an hybrid between actual autocrafting and real recipes. While AE knows how much is available and can use it for the crafting calculation, it has no access the the actual item. It still goes through the crafting system to actually request them. The autocrafting is more or less abused as request system because it is essentially the same. It is then up to whatever announces a requestable item to fullfill the crafting/request once issued by a CPU.

There is simply no way around it. Once you start duping them, which you are actually doing by wanting to extract them, you are duping them. Say you request 1 diamond, but directly extract 1 but still issue the drone/robot/logisticpipe to send 1 diamond, you'll end up with 2. Or the the drone/robot/logisticpipe does not recieve the update at all and cannot destroy the actual item.

thiakil:
> The whole point of extracting them is wrong. Otherwise the whole requestable could be removed.

Except if the assumption is that the extraction call *is* the request - which is why I mentioned what my assumption was in the initial post.

> It is then up to whatever announces a requestable item to fullfill the crafting/request once issued by a CPU.

Did you read my initial post completely? Or did you stop when you realised my assumption was wrong?

To be explicit I've added emphasis:
> Without these changes such stacks appear as craftable in the ME Terminal, **but are unable to proceed past the Craft Confirm GUI, and also unable to be used in autocrafting**.

So either setting requestCount alone is not enough (and you're not being helpful enough to actually mention that fact), or the relevant code is missing in the 1.10 branch.

> There is simply no way around it. Once you start duping them, which you are actually doing by wanting to extract them, you are duping them.

In your example sure, where I assume the Drone doesn't extract from the chest until it reaches it. How I'm actually using it is (using the Drone to aid your understanding) that the "Logistics network' / drone removes the item from the chest **inside** the extract call, and the animation would be purely aesthetic. This works in my case because **the operation is entirely inside the ME Grid**

yueh:
The problem here is that I simply do not know what you want to achieve. There are like 3 or 4 systems inside AE which could fit. But each has their own requirements and downside, which might or might not work for you. `they are items that should not (for example) be transferred by the IO Port to/from Item Cells` is simply not sufficient to say use this or that system. All fit the requirement or it could even be something AE was never designed for and therefore intentionally does not work.

On the other hand, this is a PR about an actual (breaking) change and not really a question about "I want to do this and that, what would you recommend?". So please bear with me as I simply do not want to go off-topic from the PR.

thiakil:
The pneumaticraft system is essentially how I thought it was supposed to work initially; something that's not sitting in the AEItemStack storage network as an item, but can be transformed into such a thing. A crafting step without an AE-visible input that is handled by some object attached to the Grid somewhere.

That being why I asked `How does that part work?`; you mentioned a solution.

Requiring a solution to a problem is the purpose of the PR, so I think it is on topic to talk about what system best suits.

yueh:
To give a quick summary.

Everything **`appeng.crafting`**

Pretty much only used for the initial crafting calculation. It should take requestable and the other cases into account. There is pretty much no change since 1.7.10. But it does not handle autocrafting itself. Just creates the job tree and hands it off to the CPU, which then actually request all tasks (which also means no change here will actually cause them to extract requestable items at all)

**Any normal item (via extract/insert)**

* Known amount
* Cannot fake any amount (think compacting drawers advertising nuggets, ingots and blocks at basically 3x times as actually is available)
* Time critical, extract needs to happen immediately including any cleanup of external inventory)

Common use case: Items stored in the network or external inventories via storage bus

Only case affected by IO Ports

**Craftable Items**

* Pattern with input + output must be known 
* Output will appear as craftable itemstack
* Input/Output are usually precise ItemStack, no fuzzy damage value or different NBT

Common use case: Sure, autocrafting

**Requestable Items**

* Mostly for existing items outside the network
* Not time critical (like it could take hours to fullfil a request)
* Require a known amount
  * Thus no infinite or unknown
* Could be used for crafted items, but must have exclusive inputs
  * Thus you cannot announce 1 charcoal as requestable should the network have 1 log.
  * Autocrafting cannot determine that it needs to reserve 1 log for it. So crafting torches from a single log would not work and get stuck (but probably with 2 logs as long as <= 4 torches are requested)
* Fullfilling the request is never handled by AE2 directly
  * iirc `ICraftingWatcherHost#onRequestChange()` is called by a CPU to actually request the item, which then has to be inserted into the network. Like placing into an interface.

Common use case: Any external inventory, which cannot provide immediate extraction. Drones, Logistic Pipes, etc

Crafting from an exclusive input might work, like keeping 64 log inside a furnace at all time and smelt it to charcoal should something request it. But normal autocrafting via interface/pattern or maybe level emitter is clearly the superior option.

**Emitable Items**

* Special case of crafted items
* Must be an infinite amount
  * Not being able to provide an infinite amount will stall the crafting.
* Not time critical
* (Cannot remember the class out of my head. But searching for emitable should only provide like 2-3 classes with the level emitter as example)

Common case: Cobblestone generator activated by a redstone signal (level emitter)


So yeah, the last 3 things fit the "not affected by IO Port". But that's it.
I still have no idea what you want to achieve. For whatever reason this "I have an idea, but I cannot talk about it. But that is the solution. Do you think it is the correct one for my idea?" is really common and wastes so much time by going back and forth until the actual idea is somewhat visible.

thiakil:
Thank you for the explanation. I explained my case in general terms that should have been enough in my second post. Thing which AE Crafting knows nothing about can be represented as item, it's just gotta be able to ask for it. It shouldn't really matter WHAT it actually is, implementation is external. Only the how to supply ItemStack should need to matter.

yueh:
Ultimately the CPUS just work their job tree and either `ICraftingMedium#pushPattern()` directly or `ICraftingWatcherHost#onRequestChange()` handles delivering the request item to the network. Everything else is more about giving the calculation more hints. Like infinitly emitable items are a really special edge case and need to be handle a bit differently.

But it is all about ItemStacks. The part "Thing which AE Crafting knows nothing about can be represented as item" is usally indicating something like fluid crafting or anything similar. And that will simply not work at all. It is pretty much guaranteed to either not work at all or simply stall the autocrafting at some point.

It pretty much requires either `insert itemstack, get itemstack back` or `guaranteed infinite amout emitable". Something like have 10 buckets of water requestable by calculating that a tank with 10000 mB water will produce 10 water buckets does not work. Something like `1 Bucket + 1 Block of Ice => 1 Bucket of Water` will work perfectly fine.
For anything like real liquid crafting, you pretty much have to implement your own terminal + grid + autocrafting + patterns + machines + whatever else.

thiakil:
Say you have `x` farts (thing that exists, can be anything, mana, liquids, gass). Farts are stored in a Fart Jar which is connected to the network. A crafting pattern exists for `1 fartItem + mud = poop` connected to an Assembler/Interface. `fartItem` conceptually is a `fart` tweaked by AE energy physics to be semi-stable to be crafted with. Basically what Water is to a Water Bucket

This `fartItem` is what needs to be exposed.

PR #2940, opened 2017-07-01
Fixes #2679: Correct UVs for East, North and Down.

MoreThanHidden:
(#2679) Fix for facades rendering in the wrong direction.

Changed default UVs in CubeBuilder to match original textures. Tested all directions of facades and other items that are using the CubeBuilder.

PR #2941, opened 2017-07-02
Fixes #2679: Correct UVs for East, North and Down.

### LINKED ISSUE #2679: Facades facing East and North are rendering the wrong block face
#### Environment
- Mods: [list.txt](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/629596/list.txt)
- Minecraft Version: 1.10.2
- AE2 Version: alpha 8
- Forge Version: 2135

MoreThanHidden:
(#2679) Fix for facades rendering in the wrong direction.

Changed default UVs in CubeBuilder to match original textures. Tested all directions of facades and other items that are using the CubeBuilder. Moved from fix in 1.10 branch (#2940) to master and tested.

### DIFF
--- a/src/main/java/appeng/client/render/cablebus/CubeBuilder.java
+++ b/src/main/java/appeng/client/render/cablebus/CubeBuilder.java
@@ -212,9 +212,9 @@ public class CubeBuilder
 		{
 			case DOWN:
 				uv.u1 = texture.getInterpolatedU( x1 * 16 );
-				uv.v1 = texture.getInterpolatedV( z1 * 16 );
+				uv.v1 = texture.getInterpolatedV( 16 - z1 * 16 );
 				uv.u2 = texture.getInterpolatedU( x2 * 16 );
-				uv.v2 = texture.getInterpolatedV( z2 * 16 );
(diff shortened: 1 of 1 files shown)


MoreThanHidden:
Before:
image
After:
image

yueh:
Still does not cover that it is a broken fix.

MoreThanHidden:
@yueh whats broken?

yueh:
Flipped textures on other models. Which is the simple reason it is not fixed already (we actually reverted the same during porting to 1.10)

MoreThanHidden:
No worries I'll have a look, do you have an example of something that was flipping?

yueh:
Mostly cables. But only in some parts of them. Pretty much any change to facades will break them.

MoreThanHidden:
I've thoroughly tested dense, smart, covered and regular cable textures and haven't been able to find any issues :S, also tested the displays and other cable bus items. 

I did try a bunch of different fixes earlier that did break things like this, but can't find any issues with changing it here.

MoreThanHidden:
@yueh if you could test when you get a chance as I haven't found any issues. Was planning to fix a few more issues when I have the time, is it better to make individual pull requests per issue or just branch and send though a large pull request at the end?

PR #2963, opened 2017-07-21
Fixes PatternTerminal clearing crafting slot on tried crafting

tyra314:
The ME Pattern Terminal recipe output gets cleared when you can't craft
the item. To fix that, I call getAndUpdateOutput() in the
ContainerPatternTerm::onSlotChange, when it seems appropriate.

This is a continuation and split of https://github.com/GuntherDW/Applied-Energistics-2/pull/10

This is the part, which is "probably safe"


yueh:
test this please.

PR #2966, opened 2017-07-22
Started P2P rework

yueh:
Rebase of #2692 onto rv5-1.12

PR #2974, opened 2017-07-23
Fixes some of the orod-org comments in #2966

tyra314:
This fixes some of the orod-org comments and renamed the `twoBits` variable.

The other comments of orod-org I tend to disagree.

### DIFF
--- a/src/api/java/appeng/api/features/IP2PTunnelRegistry.java
+++ b/src/api/java/appeng/api/features/IP2PTunnelRegistry.java
@@ -55,6 +55,5 @@ public interface IP2PTunnelRegistry
 	 *
 	 * @return null if no attunement can be found or attunement
 	 */
-	@Nonnull
 	TunnelType getTunnelTypeByItem( ItemStack trigger );
 }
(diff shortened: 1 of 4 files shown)


yueh:
Thanks, but the branch is still highly tentative and subject to change. It could easily never be merged. Therefore there is no need to do a simple cleanup pass at this point.

Also just removing `@Nonnull` to suppress an issue is not the best solution. I currently cannot comment on it, but should it return null, it is more likely that it should actually throw an expection or at least log it and return a default case or similar.

tyra314:
The only class implementing the `IP2PTunnelRegistry` is [P2PTunnelRegistry.java#L214](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/rv5-1.12/src/main/java/appeng/core/features/registries/P2PTunnelRegistry.java#L214). But in this class, the method is tagged as `@Nullable`. @GuntherDW changed it for the interface in eb11845a2f0b2f1c06e3399951f9acdc6a6f115e from `@Nullable`. Looking at the other changes, I assume it was just an oversight.

Issue #2978, opened 2017-07-25
Memory Card QoL Rework


yueh:
# Goal

Improving the memory card in terms of QoL/UX.
This is directly or indirectly related to other issues/FR and can needs these as prerequisite

## Make use of the proposed color coding in #2674 (prerequisite)

The minimum has to be displaying the 4 color codes as text within the tooltip. 

Additionally formatting them by the actual color might be an idea. But that can be a visibility issue. (black on black). 

Important here is that can be used as alternative for colorblindness. Even if black on black is not that much of an issue, it should not make it harder otherwise. This also means no squares with a color.
16 different shapes mapped to each color might be an alternative, main question would be the be distinct enough.

**WAILA/Hwyla/TOP** 
Similar to the item tooltip. But additionally it can use different background colors

## Improve the interaction parts by using the proposal in #1480 (prerequisite)

Currently we are limited to 2 actions per part. Right click and shift + right click. This limits us to each a single read and write action for the card. In terms of the p2p tunnels this includes also reseting the frequency and changing it to an input, when reading the settings. Without it, there would be no way to ever create a new input or even a new tunnel.

With the proposed changes in #1480 we would have at least 4 different actions available. Which could translate something like

| action | effect |
| --- | --- |
| Left Click | Read settings, no change (analog to extract item) |
| Shift + Left Click | Reset and read (new frequency/input/etc) |
| Right Click | Write settings (analog to insert item) |
| Shift + Right Click | Could also insert cards like proposed in #1382 (Or should we ever allow N:N tunnels to set input) |

## GUI 

To start a GUI directly for the memory card is pretty much a no-go. It has to cover every block and part of AE2 itself, which very different and also has to do the same for every addon. Which makes it quite complex. Further it can easily create issues when applying incorrect settings to parts. Like make more than 1 input P2P. As well as the obvious balance issues.

An option might be something like a P2P terminal. Which could contain some basic informations like the frequency (colorcode), amount of outputs and the type. Certainly not detailed infos like the amount of used channels inside the tunnel. Smart cables cover it already. Main feature would be to copy the settings of a tunnel to a memory card. Still highly tentative, as organising the network should be a challenge. Not something the network does for the player on top of autocrafting and a massive storage.

Issue #2674, opened 2016-12-01
P2P QoL Rework


yueh:
# Goal

Improve the QoL for P2P tunnels without having to resort to 3rd party modules.

## Change tunnel frequency from long to short.

The main idea is to allow some actual QoL improvements.

This will introduce a regression and reduce the amount of p2p links to 65k per network (not actual tunnel parts). Considering that the maximum amount of channels should be around 20-22k for the largest controller, there is still no way to actually hit the potential limit of 65k tunnels of a single network. Non useless setups will be even limited to around 10-11k tunnels in total.

We could further create a combined id of the type and frequency which would expand it to `types * 65k` possible links, should it ever be a problem.

One of the few real issue would be joining two networks using the same frequency for their links as this will cause a collision by introducing 2 input tunnels. But that should be really unlikely to happen, if we assign random frequencies with smaller networks and joining large networks will certainly have more issue than some P2P conflicts. A possible solution would be using using a global frequency generated by an LFSR or similar, but that would limit it to 65k links for everyone, which might be hit on very large servers. Or we could simply downgrade a random input tunnel to an ouput, when necessary.

## Colour code tunnels

The back of each tunnel currently provides 4 2x2 pixel areas, which could be used to map a frequency to a colour code. 

This is the main reason to change the frequency to shorts as this allows us to encode them with 4 minecraft colours. ( `16^4 = 2^16` )
Each spot could represent one colour, which would give it a more visible area for each colour, but could more easily be obstructed at certain angles. Or as alternative each spot could contain all 4 colours at the cost of having smaller areas per colour and thus less visible from further away.

Mockup:
![colour coded p2p tunnel](http://i.imgur.com/znCvckh.png)

Linked state and output/input could also potentially be encoded by colours similar to the state indicator on parts. With the dark colour for everything fine, light for missing channel and off for no power.

Tunnels could use the dark colours for linked input tunnels, medium for linked output and the light one for unlinked ones (of course losing input/output here) and off for no power.

Another option would be to keep longs as value and encode them as 16 colours, It would still fit the available area with `4 * (2*2)` pixels. While 4 colours are certainly easy to remember and compare, this will not really be the case for 16 ones. As well as each colour being limited to 1 pixel and thus being more easily obstructed and less visible from farther away.

## Waila/OneProbe/etc

[...]

PR #3009, opened 2017-08-06
Color code tooltip for memory cards (Part of #2978 and #2674)

tyra314:
This adds a simple tool tip representing the stored P2P frequency of a memory card.
This is part of #2674 and #2978.

What I haven't dealt with yet, is a colorblind mode. In #2674, it was suggested to use a distinct symbol for each color. According to [this](https://gaming.stackexchange.com/questions/103304/allowed-symbols-characters-in-minecraft-chat), there aren't that much supported characters, so we may simply use the RBG hexadecimal representation. However, a colorblind mode should be separated from the normal mode, as it might be hard to read symbols in colors too.

My current solution uses two `TextFormatting.OBFUSCATED` characters per color and looks like thi:.

Though, some colors are hard to see. But I don't think it is a big issue, as the only information conveyed is the color itself.

Note: This adds a new `I18n` entry, which has to be translated in other languages as well.

### DIFF
--- a/src/main/java/appeng/integration/modules/waila/part/P2PStateWailaDataProvider.java
+++ b/src/main/java/appeng/integration/modules/waila/part/P2PStateWailaDataProvider.java
@@ -21,12 +21,16 @@ package appeng.integration.modules.waila.part;
 
 import java.util.List;
 
+import appeng.api.util.AEColor;
+import appeng.util.Platform;
 import com.google.common.collect.Iterators;
 
 import net.minecraft.entity.player.EntityPlayerMP;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.text.TextFormatting;
+import net.minecraft.util.text.translation.I18n;
 import net.minecraft.world.World;
 
 import mcp.mobius.waila.api.IWailaConfigHandler;
@@ -87,6 +91,8 @@ public final class P2PStateWailaDataProvider extends BasePartWailaDataProvider
 					}
 				}
 			}
+
+			currentToolTip.add( getFrequency( (PartP2PTunnel) part ) );
 		}
 
 		return currentToolTip;
@@ -153,6 +159,22 @@ public final class P2PStateWailaDataProvider extends BasePartWailaDataProvider
 		{
 			return String.format( WailaText.P2PInputManyOutputs.getLocal(), outputs );
 		}
+	}
+
+	private static String getFrequency( PartP2PTunnel tunnel )
+	{
+		AEColor[] colors = Platform.p2p().toColors( tunnel.getFrequency() );
+
+		StringBuilder result = new StringBuilder();
+
+		result.append( I18n.translateToLocalFormatted( "gui.tooltips.appliedenergistics2.P2PFrequency", result.toString() ) );
+
+		for( int i = 0; i < 4; i++ )
+		{
+			result.append( TextFormatting.fromColorIndex( colors[i].dye.getDyeDamage() ) ).append(String.format("%X", colors[i].ordinal()));
+		}
+
+		return result.toString();
 	}
 
 }
--- a/src/main/java/appeng/items/tools/ToolMemoryCard.java
+++ b/src/main/java/appeng/items/tools/ToolMemoryCard.java
@@ -73,7 +73,7 @@ public class ToolMemoryCard extends AEBaseItem implements IMemoryCard
 
 			for( int i = 0; i < 4; i++ )
 			{
-				result.append( TextFormatting.fromColorIndex( colors[i].dye.getDyeDamage() ) ).append( TextFormatting.OBFUSCATED ).append( "##" );
+				result.append( TextFormatting.fromColorIndex( colors[i].dye.getDyeDamage() ) ).append(String.format("%X", colors[i].ordinal()));
 			}
 
 			lines.add( I18n.translateToLocalFormatted( "gui.tooltips.appliedenergistics2.P2PFrequency", result.toString() ) );
--- a/src/main/java/appeng/items/tools/ToolMemoryCard.java
+++ b/src/main/java/appeng/items/tools/ToolMemoryCard.java
@@ -29,6 +29,7 @@ import net.minecraft.util.EnumActionResult;
 import net.minecraft.util.EnumFacing;
 import net.minecraft.util.EnumHand;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.text.TextFormatting;
 import net.minecraft.util.text.translation.I18n;
 import net.minecraft.world.IBlockAccess;
 import net.minecraft.world.World;
@@ -37,6 +38,7 @@ import net.minecraftforge.fml.relauncher.SideOnly;
 
 import appeng.api.implementations.items.IMemoryCard;
 import appeng.api.implementations.items.MemoryCardMessages;
+import appeng.api.util.AEColor;
 import appeng.core.localization.GuiText;
 import appeng.core.localization.PlayerMessages;
 import appeng.items.AEBaseItem;
@@ -60,6 +62,21 @@ public class ToolMemoryCard extends AEBaseItem implements IMemoryCard
 		if( data.hasKey( "tooltip" ) )
 		{
 			lines.add( I18n.translateToLocal( this.getLocalizedName( data.getString( "tooltip" ) + ".name", data.getString( "tooltip" ) ) ) );
+		}
+
+		if( data.hasKey( "freq" ) )
+		{
+			final short freq = data.getShort( "freq" );
+			AEColor[] colors = Platform.p2p().toColors( freq );
+
+			StringBuilder result = new StringBuilder();
+
+			for( int i = 0; i < 4; i++ )
+			{
+				result.append( TextFormatting.fromColorIndex( colors[i].dye.getDyeDamage() ) ).append( TextFormatting.OBFUSCATED ).append( "##" );
+			}
+
+			lines.add( I18n.translateToLocalFormatted( "gui.tooltips.appliedenergistics2.P2PFrequency", result.toString() ) );
 		}
 	}
 
--- a/src/main/resources/assets/appliedenergistics2/lang/de_de.lang
+++ b/src/main/resources/assets/appliedenergistics2/lang/de_de.lang
@@ -323,6 +323,7 @@ gui.tooltips.appliedenergistics2.SchedulingModeRoundRobin=Exportiere im Round-Ro
 gui.tooltips.appliedenergistics2.SchedulingModeRandom=Exportiere im Zufallsmodus.
 gui.tooltips.appliedenergistics2.ItemsStored=Items gespeichert: %s
 gui.tooltips.appliedenergistics2.ItemsRequestable=Anfragbare Items: %s
+gui.tooltips.appliedenergistics2.P2PFrequency=Frequenz: %s
 
 // Units
 gui.appliedenergistics2.units.appliedenergstics=AE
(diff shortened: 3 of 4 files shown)


yueh:
Obfuscated is really bad here. It is essentially the same as showing `[ Red, Lime, Orange, Black ]` and color it `blue, pink, white, red`. It is more confusing than any additional help.

In terms of symbols, minecraft should support full unicode, I have no idea about the range it supports, but it is way more than the usual ansi/ascii font. My idea was using something like dot, triangle, square, diamond etc and map them to a color. But 16 symbols are way too much and using every simple polygon is also not helping.

Further displaying it as part of the memory card has serious drawbacks. As it is the only option to display them currently, it will encourage players to copy an existing tunnel to show it, but this will reset the tunnel completely and form a new one, break the existing links and force the player to reassign everything.

It pretty much is the last step. After the model itself shows it, there is a way to copy an existing tunnel frequency without resetting and converting it into an input, maybe some additional waila integration (only after it is usable without waila), etc.

tyra314:
I see this PR as a visual testing field for the implementation and have no intent to merge it without the missing part of #2674.

What I'd try next, is using [geometric shapes](https://en.wikipedia.org/wiki/Geometric_Shapes) for a colorblind mode, but still only color for non-colorblind mode.

tyra314:
Well, you're right, Minecraft can render more characters than the ASCII range. But, it does look like shit and is barely recognizable. I think just printing the frequency as a number, either decimal or hex would be better.

yueh:
Hex is certainly the best choice. It exactly matches the 4 colors and we have 1 characters per color.

yueh:
Sorry for the double, but just in case you miss as edit.

As already said, using hex would be a very good choice. It limits it to `0000-FFFF`, maybe one less depending on how we encode unlinked. With 4 digits, it is also not hard to remember or you can remember the color combination.

We can directly use the ordinal to map a hex digit to a color in a fixed way. As this is a bijective mapping, colorblind players can easily switch to the text representation. Still mapping 1 to RED is not the best way as we are using text and not symbols. But it is way better than having a text "blue" with lime as color. (ignore me, if this is the incorrect ordinal).

We can also reuse the exact same behaviour for WAILA/HYWLA/TOP/whatever.
Nevertheless we still have the requirement to not depend on something like WAILA for displaying additional data. There has to be a way in AE2 itself before we can add it. (This is more to annoy the devs and that "But you can use WAILA" is not a valid excuse for ignoring every other solution).
WAILA&co is certainly something nice for various cases, e.g. provide an additional text representation for colorblind players and so on.

In terms of mapping a color frequency to hex, the helper should probably extended. We might even be able to reuse it for WAILA/TOP. (This is not a task for you @tyra314, more of a TODO for myself. I hope to change it later today or over this weekend)

I still fear that many players will try to shift rightclick a tunnel to copy the frequency to the card to look it up, say without having the color dots on the back. But doing so means reseting the tunnel and having to reassign every output again.

Reworking the whole part interaction is certainly a long time to go. An intermediate solution might be ok, while it certainly trains a behaviour, which will get invalid later.
E.g. it might be an idea to make the card a two pass process:

1. Shift + rightclick on an existing tunnel will copy it to the card, no reset (this can then also copy from an output)
2. Shift + rightclick on an unlinked tunnel will initialise it and assign a new frequency.
3. Shift + rightclick on a tunnel with the same frequency as stored will reset it, change it to an input and assign a new frequeceny (Essentially double rightclick)

Of course this needs better chat messages. Mostly to distinguish between `Settings copied to the card` and `Assigned new frequency to the tunnel and changed it to an input`.

Btw. I am not perfectly convinced about the term frequency. It is more like a ID for an encapsulated packet or some broad/multicast address. Just to mention it. Maybe someone has a better idea.

tyra314:
> As already said, using hex would be a very good choice. It limits it to 0000-FFFF, maybe one less depending on how we encode unlinked. With 4 digits, it is also not hard to remember or you can remember the color combination.

Yeah, maybe this should simply be the default, because I just realized, there is an issue with #2674: 
What is the first color and which sequence do they have on the tunnel? The problem about the first comes apparent, when you place the tunnel at the ceiling or the floor.

> I still fear that many players will try to shift rightclick a tunnel to copy the frequency to the card to look it up, say without having the color dots on the back. But doing so means reseting the tunnel and having to reassign every output again.

What about a simple solution: Just don't reset tunnels, unless requested by the user with copying the settings onto the tunnel? Tbh. I wasn't even aware of the fact that copying data to the memory card resets tunnels and if I were, I'd had it reported as bug ages ago.

Therefore I'd suggest:

1. Shift + rightclick on unlinked tunnel will initialize it and assign a new freqency
2. Shift + rightclick on a linked tunnel will copy settings onto the memory card, no changes to tunnel
3. Rightclick on unlinked tunnel will apply settings
4. Rightclick on a linked tunnel with the same settings as on the card is a NOOP.
5. Rightclick on a linked tunnel with different settings will apply those new settings

yueh:
> Yeah, maybe this should simply be the default, because I just realized, there is an issue with #2674:
What is the first color and which sequence do they have on the tunnel? The problem about the first comes apparent, when you place the tunnel at the ceiling or the floor.

They indicate the channels on 2 fixed sides. So you already know one of two direction. So just one additional indicator for "this side up" would be sufficient.

> What about a simple solution: Just don't reset tunnels, unless requested by the user with copying the settings onto the tunnel? Tbh. I wasn't even aware of the fact that copying data to the memory card resets tunnels and if I were, I'd had it reported as bug ages ago.

Then we have no way to create a new tunnel without disassembling it. Should we ever decide to keep the settings, this means you cannot even reuse a tunnel ever for another tunnel. Except some workarounds like put it into a crafting table and craft it to a new one...

3, 4, and 5 a the same. There is no reason to distinguish between any of them. It is always "apply settings". We just need a way to initialise a tunnel as input, copy an existing freq to the card and apply that to another tunnel to change it to an output.

tyra314:
Just to let you know, I still want to work on this issue, but I ran out of time right now.

> Then we have no way to create a new tunnel without disassembling it.

I think that is better than the alternatives:
a) having different behaviours for consecutive copy-to-memory-card operations - I'm one hundred percent sure that I would always end up copying twice by accident, which would reset my tunnels.
b) having the tunnel reset on every single copy operation - which would reset my tunnels every time ;)

> Should we ever decide to keep the settings, this means you cannot even reuse a tunnel ever for another tunnel. Except some workarounds like put it into a crafting table and craft it to a new one...

IMHO, we should keep that in mind and discuss it, if we actually implement keeping settings. Or are you planning ahead for this feature?

> 3, 4, and 5 a the same.

Yes, I just wanted to list every single possible operation from a player's point of view.

yueh:
No need to hurry. If it takes time, it takes time.

> a) having different behaviours for consecutive copy-to-memory-card operations - I'm one hundred percent sure that I would always end up copying twice by accident, which would reset my tunnels.

That is actually a valid point. But it also opens up other question. Like attuning a tunnel to a different type also resets it. And that can also happen by just accidentally clicking it with a random block/item because you wanted to point at it for various reasons. Without any real solution for it. Sure we can pop up a modal dialog about "Are you sure to reset it?". Or some artifical time gating like "Please click again after 10 secs, if you want to reset it.". In both cases some will just click yes or spam clicking it, without reading it and then complain about it having reset the tunnel. While everyone else complains about an artifical timesink and why do they have to wait 10 secs before they can reset a tunnel.

It is a balance act. But I certainly have no issues with punishing the player a bit for "stupid" actions. If one screws up by double clicking it (and they get enough immediate feedback via chat/etc), it might happen once or twice and after they will have learnt that double clicking everything is bad. If not, any other solution is probably not improving it, just making it worse for anyone else.

> IMHO, we should keep that in mind and discuss it, if we actually implement keeping settings. Or are you planning ahead for this feature?

I have not really made my mind up about it. It can easily flood a players inventory, if suddenly nothing stacks anymore due to NBT being preservered. There might be other solutions to it like allowing the memory card to store multiple profiles. But selecting them is pretty annoying. Anther reason was also that the old recipe system never supported NBT, so we could not ignore certain tags for using items with additional NBT as crafting ingredient. That is certainly easier now, but then allows something like "OMG. It eat my export buses with important configs instead of crafting new ones"

src/main/java/appeng/integration/modules/waila/part/P2PStateWailaDataProvider.java
```diff
@@ -21,12 +21,16 @@
 
 import java.util.List;
 
+import appeng.api.util.AEColor;
+import appeng.util.Platform;
 import com.google.common.collect.Iterators;
```
orod-org: MINOR Wrong order for 'com.google.common.collect.Iterators' import. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck)

src/main/java/appeng/integration/modules/waila/part/P2PStateWailaDataProvider.java
```diff
@@ -155,4 +161,20 @@ private static String getOutputText( int outputs )
...
+	{
+		AEColor[] colors = Platform.p2p().toColors( tunnel.getFrequency() );
+
+		StringBuilder result = new StringBuilder();
+
+		result.append( I18n.translateToLocalFormatted( "gui.tooltips.appliedenergistics2.P2PFrequency", result.toString() ) );
+
+		for( int i = 0; i < 4; i++ )
+		{
+			result.append( TextFormatting.fromColorIndex( colors[i].dye.getDyeDamage() ) ).append(String.format("%X", colors[i].ordinal()));
```
orod-org: MINOR '(' is not followed by whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)
MINOR '(' is not followed by whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)
MINOR ')' is not preceded with whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)
MINOR ')' is not preceded with whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)

src/main/java/appeng/items/tools/ToolMemoryCard.java
```diff
@@ -61,6 +63,21 @@ public void addCheckedInformation( final ItemStack stack, final World world, fin
...
+		if( data.hasKey( "freq" ) )
+		{
+			final short freq = data.getShort( "freq" );
+			AEColor[] colors = Platform.p2p().toColors( freq );
+
+			StringBuilder result = new StringBuilder();
+
+			for( int i = 0; i < 4; i++ )
+			{
+				result.append( TextFormatting.fromColorIndex( colors[i].dye.getDyeDamage() ) ).append(String.format("%X", colors[i].ordinal()));
```
orod-org: MINOR '(' is not followed by whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)
MINOR '(' is not followed by whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)
MINOR ')' is not preceded with whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)
MINOR ')' is not preceded with whitespace. [rule](https://sonar.orod.org/coding_rules#rule_key=checkstyle%3Acom.puppycrawl.tools.checkstyle.checks.whitespace.ParenPadCheck)

orod-org:
SonarQube analysis reported 9 issues
* ![MINOR][MINOR] 9 minor

[MINOR]: https://sonarsource.github.io/sonar-github/severity-minor.png 'Severity: MINOR'

tyra314:
Sorry again for the long delay. It took me a while to get back in the mood :(

However, I changed the tooltip to a colored hexadecimal number. Still, black and dark blue, both are hard to read but distinguishable. 

I started working on the WailaProvider, so this could be at least be somewhat useful even without the other rendering done. However, the IPart you have access to in the WailaDataProvider only returns 0 for the frequency, regardless of link state. I'm not sure what is going on there.

PR #3025, opened 2017-08-11
Fill existing stacks first when adding to player inventory.

fscan:
I also removed the .copy() in AddItems because i think it's not needed. IItemHandler insertItem only returns the same stack if it hasn't changed (disregarding buggy implementations). Am i missing something here?

Obviously needs a lot of testing.


yueh:
Buggy implementations are sadly quite common. But there is also the chance that we do not know the source of an itemstack, so they might expect that we do not modify the itemstack similar the an itemhandler, which must not changed to passed one at all.

Also copying an itemstack is not that bad in controlled cases. It only gets out of hand when moving a couple of 10k item/tick around. And it usually needs very good optimitized inventories to even allow it.

Therefore I would err on the side of caution and keep it.

PR #3033, opened 2017-08-15
Fix AppEngSlot crash

fscan:
do not call super methods


mindforger:
is that one possibly related to the autocrafting crash happening when you have wireless crafting window mod installed and trying to autocraft while having items in particular slots?

fscan:
this is strictly 1.12 as it it's a bug from the IItemHandler conversion

PR #3106, opened 2017-09-21
Fixes #3105: Dupe bug due to forge default itemhandler implementation.

### LINKED ISSUE #3105: Dupe bug using only AE2 items
- Minecraft Version: 1.12.1
- AE2 Version: rv5-alpha-4
  -- Noticed this in the log but cannot find that version for 1.12.1
       [AE2 VersionChecker/INFO] [AE2:C]: Newer version is available: rv4-stable-1 (found) > rv5-alpha-4 (current)
- Forge Version: 14.22.1.2485
- No other mods installed

Found this while looking at another dupe bug over at the [Applied Llamagistics github](https://github.com/thiakil/Applied-Energistics-2/issues/65)  

image1

The image above shows the minimum needed to replicate. This was done in a custom testpack using only AE2 items with only ae2 installed (see hotbar of screenshot). No priorities were even set (only needed if using other storage).   

Just set the above system up and then place 64 then 1 of any item in the chest to start the duping.

yueh:


fscan:
wouldn't be a copy at the assignment to gathered be enough?

PR #3177, opened 2017-10-25
Fix texture error

p455w0rd:
Fixes the `textures/parts/cable/dense_covered/0.png` missing texture error


yueh:
Already fixed locally.

PR #3336, opened 2018-01-16
Update ru_RU.lang

Foxgreat:
Hello, please update Russian lang file.

Updated phrases. Corrected code in association in one row and added missed words in russian. Also some grammar updates.


yueh:
Thanks.

Issue #3476, opened 2018-05-02
Crash with Name Inscriber Press and Inscriber


sajhino:
[Crashlog](https://pastebin.com/A7QZ3aAy)

Modpack: FTB Revelations 1.8.0
Minecraft Version: 1.12.2
Forge Version: 14.23.2.2627
AE Version: rv5-stable-11

Steps to replicate crash:
1. Make 2 Name Inscriber Press.
2. Place them into an Inscriber, one on the top slot and one on the bottom slot.
3. Place an ME Interface into the middle slot.
4. Remove the top Name Inscriber Press as the ME Interface is being renamed.
5. Game crashes.

Can't even load the world anymore after the crash, have to use an older backup.

mindforger:
hope you don't mind me slapping the label on it @yueh, this one seems confirmed, a more clean report is over in the linked issue i just closed because of duplication

https://pastebin.com/PnQrSya4

from @bkbarbee for reference

PR #3494, opened 2018-05-16
Make sure there are item stacks before calling isEmpty() (#3476)

kylev:
It was possible to cause a crash by removing something from one of the
slots during a crafting operation.


yueh:
These checks are essential.

kylev:
This is actually a crash fix for #3476. The old code throws a NPE when you remove something from an optional input slot.

The new `hasOptional()` helper method returns true if the necessary optional ingredients are preset, but also does null checks before calling isEmpty().

yueh:
In return it introduces new NPEs. Invalidates some assumptions necessary for planned changes, therefore having a good chance that it is already buggy right from thr start or needing special handling and more code to maintain, which this requirement wants to avoid. And so on.

PR #3495, opened 2018-05-17
Eliminate several javadoc warnings.

kylev:


yueh:
Sorry, but these changes are still incorrect in some places or simply do not add anything valuable to them.

kylev:
This isn't intended to "add", it just corrects some of the documentation so that it doesn't throw compiler warnings. I'm just trying to help with maintenance so the codebase is more pleasurable for other developers to participate. 😄

yueh:
I understand this. Yet I doubt that many potential modders will even notice compiler warnings, otherwise there would already be an issue about it. Also incorrect docs do not make it more pleasurable for other devs as well as it does not really save me any time as I would have to correct them and deleting a few lines again is the least amount of effort then.

PR #3570, opened 2018-06-28
update zh_cn.lang

xsun2001:
Update zh_cn.lang


Metric-Void:
Well, they’re not closing it. Maybe they just haven’t seen this. (For this long)

yueh:
Basically it's not ideal to merge localizations before the strings are really stable. It just causes translators constantly having to update it in case something needs a change.

Also please do no translate comments, they are mostly for us devs to quickly locate a certain section. If really needed, at least preserve the normal one and just append the translation. 
E.g. something like `// foobar (translation)`

yueh:
Thanks

Issue #2842, opened 2017-03-22
[Suggestion] Reimplement OC P2P


XFactHD:
It would be great if someone who has the time and the knowledge to do it could reimplement the OC P2P Tunnel for 1.10.2. I would do it myself but I currently have very little free time because of school.

CraftMinerKyle:
Can't you just copy and edit the original code, or is it really that much work to re-implement the P2P Tunnels?

Notawallplan:
In the meantime, work around it using OC's wireless networking. You can actually do a lot with it if you understand it.

TheMuso:
It is for 1.12.2, but I've worked on this. #3720 is the pull request. If you know how to build mods, feel free to test.

PR #3720, opened 2018-09-07
(#2842) Re-add support for the OpenComputers network P2P Tunnel

TheMuso:
The OpenComputers P2P tunnel was disabled during the port to 1.8. This
commit re-adds the OpenComputers P2P tunnel. The implementation is based
on the OpenComputers P2P tunnel code from AE2 in 1.7.10, updated and
refactored based on the IC2 P2P tunnel and general integration module
design.

Tested and working on Minecraft 1.12.2, Forge version 14.23.4.2750,
with OpenComputers version 1.7.2.67, connecting a RAID block, tier 3
screen and keyboard to both a tier 3 computer and a tier 3 server.
Also successfully tested AE2 with the OpenComputers mod removed.

Also happy to split into separate commits if preferred, wasn't entirely sure about this from contributing guide, which mentions splitting into logical units, so not sure if that meant separate commits for changes to different areas of the codebase, or if it meant separate commits for each piece of overall functionality change.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -27,6 +27,7 @@ cofhcore_version=1.12.2-4.5.2.19
 crafttweaker_version=4.1.8.9
 inventorytweaks_version=1.63
 ctm_version=0.3.0.15
+opencomputers_version=MC1.12.2-1.7.+
 
 #########################################################
 # Deployment                                            #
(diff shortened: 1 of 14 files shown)


yueh:
Thanks, but the reason the old implementation was not readded is simply that it was always somewhat unstable and stopped working randomly or even crashing. As well as making P2P tunnels a bit less static before adding more ones again.

PR #3750, opened 2018-09-30
Allow JEI to differentiate between different cable facades

mezz:
I occasionally get issues related to hiding AE cable facades in JEI, because JEI does not understand how to differentiate between different facades.

This PR lets JEI understand how to tell the difference between facades.
This does not impact looking up the recipes for facades in JEI.
I also bumped the dev version of JEI to the latest version, but it is not necessary for this.

Related JEI issues:
https://github.com/mezz/JustEnoughItems/issues/1256
https://github.com/mezz/JustEnoughItems/issues/1291
https://github.com/mezz/JustEnoughItems/issues/1346


covers1624:
Failure seems to only be related to SonarQue bot.

yueh:
Yes, travis does not set the required envvars for an external PR. A bit annoying as I would have to disable them (if possible), but that also makes it somewhat useless.

PR #3819, opened 2018-11-11
Fixed sources jar to include api sources

BrockWS:
The sources jar will now include both main sources/resources and api sources.


yueh:
Closing in favour of the newer PR.

PR #3896, opened 2019-01-02
Color Applicator enhancements

### LINKED ISSUE #88: Color applicator (easy color switching)
If you have 16 different dyes in your color applicator it can be a nuisance to find color you want, a way to switch quickly would be nice. Something like shift+middle click on already colored cable would switch the color applicator to color of that cable.

Originally Posted by auboli77

This issue was migrated; For original details see http://ae-mod.info/Tracker/0918/

### LINKED ISSUE #3854: Color applicator offhand cable placement
**Describe the feature**

The ideia is to select a color in the color applicator, put it in the offhand and when a cable is placed it will place a cable with the selected color in the applicator.

When playing I saw a lot of situations that this would help. Not having to craft a specific colored cable every time or having to place all the cables and go back coloring them. Also I see an improvement when placing cables next to each other when you don't want them to connect since placing a standard one would cause the network to connect and refresh the connections until you go and color it.  This would go nice with #88  but I suggest getting the color with only holding shift and scrolling while aiming at the colored cable.

### LINKED ISSUE #3895: Better indication when the Color Applicator runs out of the selected dye color
**Describe the feature**
The Color Applicator should switch to the "Empty" rendering when the currently selected dye is completely exhausted.

This functionality was suggested in another issue as something that should be considered, but no item for it was opened in the issue tracker.

**Additional Context**
I am implementing enhancements for #88 and #3854 and felt this should be included with those fixes, but wanted an issue in the tracker for this additional change.

Mordenkainen:
Enhancements to the Color Applicator:
- Improve the indication that a dye is empty. Fixes #3895 
- Select a color by clicking on a colorable block. Fixes #88
- Place properly colored cables when holding Color Applicator in the other hand. Fixes #3854

### DIFF
--- a/src/main/java/appeng/client/ClientHelper.java
+++ b/src/main/java/appeng/client/ClientHelper.java
@@ -311,7 +311,7 @@ public class ClientHelper extends ServerHelper
 	@SubscribeEvent
 	public void wheelEvent( final MouseEvent me )
 	{
-		if( me.getDwheel() == 0 )
+		if( me.getDwheel() == 0 && me.getButton() != 2 )
 		{
 			return;
 		}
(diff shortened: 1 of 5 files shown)


shartte:
Finally reimplemented this for 1.18 (#6013)

PR #3904, opened 2019-01-10
Added support for fluids in pt-BR lang file

InterPlay02:
Hello!

I've updated the pt-BR.lang file to add support for the new fluids feature.
I hope you can use it!

Bye!


InterPlay02:
Why these checks are failing?
I don't understand.

Mordenkainen:
Looking at the build it looks like something is up with the SonarQube scanner. Don't think it has anything to do with your changes.

InterPlay02:
Well... I will wait then.

Let's see if anyone will merge this.

Issue #3915, opened 2019-01-27
ConcurrentModificationException at EnergyGridCache.java:316


fluffle:
Hi! This is a dupe of #3800, which is locked, but please don't close this without reading it! I managed to trigger this exception repeatedly in a very short space of time, so I think I can provide some useful information on a way to reliably reproduce the crash and thus narrow down the root cause.

### Background

I was scaling out my AE system by adding a large number of quantum bridges. I had what I thought was a neat solution to the "both sides of the bridge must be powered" problem: I put an AE energy cell on the remote side of each bridge link. This was charged while the link was active, and contained enough of a buffer to kick the bridge back into life when the world was reloaded. It would then recharge itself.

### Problem

Quantum bridges are power hungry. The crashes started when the number of bridges on my network exceeded the power I was supplying (5400 RF/t from a nuclearcraft reactor, FYI), causing parts of the network to switch on and off rapidly. If we look at what the code is doing:

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/rv6-1.12/src/main/java/appeng/me/cache/EnergyGridCache.java#L316

Line 316 is in the middle of `extractProviderPower`. It's iterating over the LinkedHashSet of energy providers in the grid cache, and removing these providers from the grid cache if they don't have enough power. The set of providers is modified in `storagePowerChangeHandler`, `removeNode` and `addNode`:

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/rv6-1.12/src/main/java/appeng/me/cache/EnergyGridCache.java#L151
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/rv6-1.12/src/main/java/appeng/me/cache/EnergyGridCache.java#L509
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/rv6-1.12/src/main/java/appeng/me/cache/EnergyGridCache.java#L557

And iterated over (read-only) here in `refreshPower`:

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/rv6-1.12/src/main/java/appeng/me/cache/EnergyGridCache.java#L287

[...]

Issue #3800, opened 2018-10-23
ConcurrentModificationException when loading into world


Direcub:
Whenever i load my world my fps stutters heavily and eventually the game freezes 
When the game locks it never starts back up again,

Using custom modpack i made in twitch client, in Singleplayer

log: https://pastebin.com/X0kNfjdK

- Minecraft Version: 1.12.2
- AE2 Version: rv6-stable-3
- Forge Version: 14.23.5.2770

yueh:
Does this happen always? If so can you try and reduce the amount of mods to a mininum amount possible, or alternative remove just one or 2 until it no longer appears?

On the first glance it looks like something tries to use multithreaded worldloading, which is just wrong and we are just the first mod affected by it and crash therefore.

Direcub:
Will do this evening

gtokarz:
@Direcub what was your findings, having same problem in my quite random private SP pack

Direcub:
i couldnt find the mod causing it. No matter the combonation it seemed to happen at random and only in that world

Shybella:
Encountering this too. 

```Time: 1/7/19 3:23 AM
Description: Exception in server tick loop

java.util.ConcurrentModificationException
	at java.util.LinkedHashMap$LinkedHashIterator.remove(Unknown Source)
	at appeng.me.cache.EnergyGridCache.extractProviderPower(EnergyGridCache.java:316)
	at appeng.me.cache.EnergyGridCache.extractAEPower(EnergyGridCache.java:248)
	at appeng.me.cache.EnergyGridCache.onUpdateTick(EnergyGridCache.java:199)
	at appeng.me.GridCacheWrapper.onUpdateTick(GridCacheWrapper.java:43)
	at appeng.me.Grid.update(Grid.java:280)
	at appeng.hooks.TickHandler.onTick(TickHandler.java:228)
	at net.minecraftforge.fml.common.eventhandler.ASMEventHandler_1573_TickHandler_onTick_TickEvent.invoke(.dynamic)
	at net.minecraftforge.fml.common.eventhandler.ASMEventHandler.invoke(ASMEventHandler.java:90)
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:677)
	at net.minecraftforge.fml.common.eventhandler.EventBus.post(EventBus.java:634)
[...]

PR #3963, opened 2019-03-23
(#3800) Don't fire PowerStatusChange events while extracting power.

fluffle:
EnergyGridCache.extractProviderPower holds an iterator over its provider
set while calling extractAEPower on each provider.  When it extracts
power from its localStorage and empties it, firing a PowerStatusChange
event immediately can de-power a TileController and trigger the
destruction of the ME grid which both the TileController and
EnergyGridCache are part of. This in turn causes the TileController to
be removed from the EnergyGridCache's provider set while the iterator is
still live, resulting in a ConcurrentModificationException.

Instead of firing the event immediately, set a semaphore to notify the
EnergyGridCache the localStorage is out of power and rely on the
already-existing logic to fire the PowerStatusChange once grid power
extraction has completed.

Evidence in this backtrace: https://pastebin.com/yJLtAcx8

I described the scenario which allows me to reliably trigger the CME in
#3915. With this patch I can no longer trigger it. I don't know whether
there are any ramifications from delaying the event until later in the
update code, or if there are better ways to deal with this. Only trying
to help :-)

### DIFF
--- a/src/main/java/appeng/me/cache/EnergyGridCache.java
+++ b/src/main/java/appeng/me/cache/EnergyGridCache.java
@@ -110,6 +110,7 @@ public class EnergyGridCache implements IEnergyGrid
 	private double lastStoredPower = -1;
 
 	private final GridPowerStorage localStorage = new GridPowerStorage();
+	private boolean localStorageEmptiedThisTick = false;
 
 	public EnergyGridCache( final IGrid g )
 	{
@@ -193,6 +194,7 @@ public class EnergyGridCache implements IEnergyGrid
 
 		// power information.
 		boolean currentlyHasPower = false;
+		this.localStorageEmptiedThisTick = false;
 
 		if( this.drainPerTick > 0.0001 )
 		{
@@ -222,7 +224,7 @@ public class EnergyGridCache implements IEnergyGrid
 		{
 			this.publicPowerState( true, this.myGrid );
 		}
-		else if( !this.hasPower )
+		else if( !this.hasPower || this.localStorageEmptiedThisTick )
 		{
 			this.publicPowerState( false, this.myGrid );
 		}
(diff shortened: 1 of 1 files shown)


fluffle:
CI is failing because the local SonarQube server can't be found :-(

yueh:
No as this most likely won't fix the actual cause.

fluffle:
I think you're saying that to be sure, you have to enforce that no `extractAEPower` implementations can ever directly or indirectly cause any grid changes that could result in `this.providers` being modified? But I'd love to hear your description of the root cause: this has been a fun problem to bang my head against but I am sure I'm at risk of irritating you with my constant half-baked bullshit :-)

I skimmed over all of the implementations of extractAEPower I could find, and while some do post other events, they're all `MENetworkPowerStorage` events. AFAICT these are handled by the EGC and all end up modifying `this.requesters` instead of `this.providers`.

The wrinkle I noticed is that some of those implementations (`ChannelPowerSrc`, `TileChest`),  delegate `extractAEPower` back to an `IEnergySource`, which is almost certainly the EGC itself. I don't _think_ these things would end up as providers to the EGC, but proving that exhaustively is a large undertaking. However, I also think that if it did happen, you'd potentially end up with infinite recursion rather than the CME, because each layer of calls to `extractAEPower` ends up creating a new visited set, so that's a different bug.

Assuming this isn't the case, I can't identify any other routes by which extracting AE power can actually cause a grid change. That's not to say they don't exist, of course, because this codebase is large and complex and `grep` is not a replacement for a full understanding of things. But! You're getting a few bug reports a month about this, and closing off one proven way of triggering the exception is at least a step forward. Perhaps you might consider not making the perfect the enemy of the good?

Otherwise, I'll keep my patched .jar and get back to GregBlock. Sorry for all the hassle over the past few days!

PR #3972, opened 2019-03-25
Fixed broken link

mrbesen:
I know, this is not the best idea, but there is no master branch....


yueh:
Thanks, but I plan to reintroduce a master branch with 1.13/1.14 (or even earlier). So probably something, which will be overlooked again.

PR #4121, opened 2019-08-12
Fixes #4119 Sort the list of items so that missing items appear first

### LINKED ISSUE #4119: Sort crafting plan with unavailable items first
**Describe the feature**
Sort the items in the "Crafting plan" window so that missing items are displayed before everything else.

The crafting plan can get pretty big for complex builds. Missing items appear randomly scattered throughout the list in red, and it means you've got to slowly scroll down the screen to see what you don't have available.

**Additional Context**
Here's an example from Enigmatica 2: Expert: I'm building a bunch of infinity catalysts, but the flux singularities are the only thing I don't have enough of. But because it appears about half-way down the list, it's really hard to know (it also means I'm never 100% sure that this is the *only* thing I'm missing...) If missing items were sorted at the top, it would be so much easier to use.

Screen Shot 2019-08-10 at 6 24 58 PM

codeka:
So I'm not sure if this is in the right style or whatever, so happy to take suggestions. I do think this approach (i.e. sorting explicitly in the GUI code) is better than the current approach which relies on the fact that packets are received in a certain order to put things in the right order.

Here we explicitly sort the items so that missing items appear first. Secondarily, items with the biggest stack appear second.

### DIFF
--- a/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java
+++ b/src/main/java/appeng/client/gui/implementations/GuiCraftConfirm.java
@@ -243,6 +243,24 @@ public class GuiCraftConfirm extends AEBaseGui
 
 		final int offY = 23;
 
+		// Sort the list of items so that missing (and partially-missing) items appear first.
+		this.visual.sort(new Comparator<IAEItemStack>() {
+			@Override
+			public int compare(IAEItemStack lhs, IAEItemStack rhs) {
+				IAEItemStack lhsMissing = GuiCraftConfirm.this.missing.findPrecise(lhs);
+				IAEItemStack rhsMissing = GuiCraftConfirm.this.missing.findPrecise(rhs);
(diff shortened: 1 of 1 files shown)


yueh:
This is not really an approach we could take.

First it sorts the list every frame, which could get quite expensive. Also it would cause the normal crafting status to have items constantly jump around. Easily in a way, that it makes it impossible to follow what currently is crafted because the ingredients flip their position every tick.
Also it would mostly keep related chains together and not randomize them all over the place.

Insertion order is pretty much the only option the GUI can use. And that would be what would have to be changed.

codeka:
Yeah, I realised that every-frame thing after uploading this. But good point about things jumping around as well.

One idea I had would just be to reverse the order the missing/available items are sent from the server: send the missing items first and the available items seconds. Then all of the missing ones would end up at the start of the list, even if there were also some available.

But I don't know where in the server code the packets originate from, so probably not an easy patch for me :)

PR #4180, opened 2019-10-20
Update API JavaDocs for clarification

ruifung:
Update the API JavaDocs for IGridBlock.getMachineRepresentation() and IPart.canBePlacedOn() to better document the assumptions made by AE2 for these methods.

- Dense cables should not have functional (grid connected) parts.
- IGridBlock.getMachineRepresentation is non nullable.


yueh:
Not really sure about the dense part. It feels to be too specific for an API documentation, but I currently missing a better idea.

Issue #4112, opened 2019-08-05
Issue with Auto Crafting : Craft staying on scheduled


Cikei:
Hi, I'm playing on a mod pack: Ragnamod V. I'm trying to auto-craft items with AE2. I can auto craft simple items like piston or others. I want to craft solar panel IV to VIII from the mod Solar Flux Reborn but it stays on scheduled.

I'm playing on Minecraft 1.12.2. AE2 : appliedenergistics2-rv6-stable-6.jar and forge: forge-14.23.5.2838

yueh:
Please report all autocrafting related bugs directly to your modpack as most are using addons which intentionally break it.

Issue #3897, opened 2019-01-05
Autocrafting with non-stacking items for mekanism not craftable with autocrafting


ApostolosRisen:
Autocrafting solar panels, and advanced solar panels fails to finish construction

**issue**
When setting up autocrafting for mekanism solar panels, and subsequent types of solar equipment, advanced solar panels for reference, or an altered recipe for the solar neutron activator, altered to require a solar generator, The autocrafting reaches the final assembly stage and then stops without completing the crafting job, or ending the crafting process. All required recipes are in an interface connected to an assembler, Solar panel(crafts with no problem)< Solar Generator (fails to craft despite all resources needed being also included in the autocrafting recipes)< Advanced Solar Generator(fails to craft do to crafting failure of Solar Generators). this could possibly affect the production of induction matrix parts as well, I haven't progressed far enough to run into this problem

**To Reproduce**
place crafting recipes for all needed resources and components into autocrafting system, start a call order for either the Advanced Solar Generator, or the Solar Generator, ensuring that all the needed supplies to craft them from nothing are in the system.

**Expected behavior**
the system should produce either, 1 A Solar Generator/s, or 2 several Solar Generators then an Advanced Solar Generator and place it in the ME System.

**Environment**
both Single player world, and public server produced the same results for me, info and modpack are both available through technic launcher, and at this address https://www.technicpack.net/modpack/nw-the-end-of-days.1321792/about

large list of mods added in png format for viewing 
![modlist pt 1](https://user-images.githubusercontent.com/26664327/50719317-6e056000-1068-11e9-8c41-d84f3ba15705.PNG)
![modlist pt 2](https://user-images.githubusercontent.com/26664327/50719320-6fcf2380-1068-11e9-8bc4-00b843b44a23.PNG)
![modlist pt 3](https://user-images.githubusercontent.com/26664327/50719321-7362aa80-1068-11e9-8b51-29444bffedc0.PNG)
![modlist pt 4](https://user-images.githubusercontent.com/26664327/50719324-79588b80-1068-11e9-96a6-4b95a6718033.PNG)
![modlist pt 5](https://user-images.githubusercontent.com/26664327/50719335-93926980-1068-11e9-885d-16c2d92136c8.PNG)

- Minecraft Version: 1.12.2
- AE2 Version: rv6-stable-3
- Forge Version:14.23.5.2768

ApostolosRisen:
[...]

Issue #3531, opened 2018-06-06
Crafting Lag/Console Spam when Crafting Mekanism Items (Cables / Induction Cells / Induction Providers)


Vyraal1:
#### Description
Mekanism features a lot of nested crafting recipes (eg. to craft next tier of cables from Advanced Tier to Ultimate Tier, you need to craft 8 of Advanced Tier Cables with some alloys into 8 Ultimate Tier Cables).
![image](https://user-images.githubusercontent.com/26587049/41043435-ef1e3ebc-6971-11e8-9410-b8f1b4833146.png)
These items have the same resource ID and and metadata, and only really differentiated by NBT.

Whenever I order a stack or so of any of these nested crafting recipes (only happens with Mekanism, though I don't know of any other mod that heavily uses NBT to differentiate all of its blocks instead of metadata), I get the following console spam https://hastebin.com/mewecusexi.pl as well as a slowndown in my ME system's autocrafting. It seems as if it's trying to use other inputs and getting rejections? Not sure really.

A sample pattern looks like this:
![image](https://user-images.githubusercontent.com/26587049/41043369-cbd3cb3e-6971-11e8-9699-ad79db181981.png)

Another pattern for the ultimate cables, hard to see the color but those are elite cables in the left crafting grid - 
![image](https://user-images.githubusercontent.com/26587049/41043595-45872890-6972-11e8-851d-3843cbf7a8ce.png)

#### Environment
**AE2 Version:** rv5-stable-11
**Server Play:** hosted by SYS (OVH Canada), I'm the only person on the server
**Minecraft Version**: 1.12.2
**Modpack:** Enigmatica 2 Expert, v1.26c
**Forge Version:** 2705

Pingger:
my guess would be, that the 'ore substitutions' are causing this issue. Try to disable it on these patterns

yueh:
Yes, due mekanism's approach to stick everything into a few ids even without metadata and just use NBT, it makes it virtually impossible to do select the actually matching item for a recipe without trying every other item first and then failing it. So nothing we can actually do, except logging these cumbersome recipes for visibility.

The only solution would be for mekanism to change how they identify items and assign every item a unique it. Which is probably impossible for them currently and has to wait to for a 1.13 release, which will also be the suggested approach by minecraft itself. If they even consider it.

Vyraal1:
[...]

PR #4194, opened 2019-11-09
Fixes #4184 can't  crafting with damage item

### LINKED ISSUE #4184: Generator not crafting
Put ic2 generator in auto craft, accumulators, furnaces corpuses had been crafted, but the generator not crafting.
Adventure-Time_80--1920x1080
Adventure-Time_48--1920x1080

- Minecraft Version: 1.12.2
- AE2 Version: appliedenergistics2_rv6_stable_4
- Forge Version:14.23.5.2838

sagesse-cn:
If make a pattern with damage item  for third-party mods (such as ic2-exp), start autocrafting will not work properly.
You might think this is a third-party modpack author problem, but no.
`AESharedItemStack` will make a bounds for damage item, it using `setItemDamage` API to adjust item damage won't be a problem. 
But if `setItemDamage` doesn't work properly, It can't change the damage by `setItemDamage`. 
This is common because the damage must be modified by a specific device for damage items.
If you think this is a modpack author issue, this might be.
But it should work fine under non-fuzzy conditions，unfortunately it can't, because if can't using `setItemDamage ` change item damage, lower-bound equal to upper-bound, which will filter all items.
This is a problem that seriously affects the experience, and it has been around for a long time, can it be merge into the code as soon as possible?
@yueh

### DIFF
--- a/src/main/java/appeng/util/item/AESharedItemStack.java
+++ b/src/main/java/appeng/util/item/AESharedItemStack.java
@@ -46,6 +46,14 @@ final class AESharedItemStack implements Comparable<AESharedItemStack>
 		this.itemDamage = itemStack.getItemDamage();
 	}
 
+	public AESharedItemStack( final ItemStack itemStack, final int itemDamage )
+	{
+		this.itemStack = itemStack;
+		this.itemId = Item.getIdFromItem( itemStack.getItem() );
+		this.itemDamage = itemDamage; // setItemDamage may fail, so don't read from itemStack.
+	}
+
+
 	Bounds getBounds( final FuzzyMode fuzzy, final boolean ignoreMeta )
 	{
 		return new Bounds( this.itemStack, fuzzy, ignoreMeta );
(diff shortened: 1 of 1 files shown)


sagesse-cn:
More possible fixes #4112 #3897 #3531

bigenergy:
I tested this build. But craft nano chestplate (IC2) not working.

sagesse-cn:
Thx, I guess you use the JEI auto-fill pattern. 
if you put `energy crystal` with manually, it will work when it exists an item.
that does have some problems, and I'll try to fix it later.
image

yueh:
If `setItemDamage()` is broken, that is clearly a bug in the respective mod. Ignoring it might appear to work in a couple cases, but it will introduce invalid itemstacks into the cache, which might have unknown side effects.

Also if you confirm yourself, that players still have to take the same workarounds into account as currently, it doesn't appear to really fix it. Maybe just hide the actual bug and making it harder to actually identifiy.

sagesse-cn:
> that players still have to take the same workarounds into account as currently, it doesn't appear to really fix it.

If you are referring to @bigenergy's issue, his problem is not caused by this modification, it already exists in old version.

> If `setItemDamage()` is broken, that is clearly a bug in the respective mod.

For example, electric power equipment in ic2, magic equipment in mystery, RF equipment, it can not and should not using experience repair. it must be repaired in the specified equipment.
Anyway, that's how a lot of mods are implemented now, right.

> it will introduce invalid itemstacks into the cache

lower-bound and upper-bound is a "sentry", this PR that simply modifies them without affecting the original itemstacks.
If into an invalid item stack to the cache, the old version same.  
this may be a design issue. I think it's better to let player choose (some button in the pattern plane) whether to turn on fuzzy mode .

yueh:
In other words it does not fix the known issue, but something else I am not aware of. Nor haven't seen as actual bug.

All it does is add a mismatch between the cached itemstack damage value and the actual itemstack damage. Which is actually pretty important to not be different for our datastructures.

The rule of thumb for now was to always use the normal way. E.g. in this case rely on `setItemDamage()`/`getItemDamage()`. If then some mod decides to be special and wants to break vanilla behaviour, that is fine for them, but we won't start to introduce some special cases just to deal with them at the risk of breaking it for every mod complying with the standard APIs.

Issue #4189, opened 2019-11-03
AE2 won't let the game to launch


berkkrkc09:
I am the owner of the "Quick Quack" pack on twitch desktop client. I was trying to relase a new version for my modpack. Just after i launch minecraft, game crashes and it says AE2 is the potential cause of the problem.

**Describe the bug**
AE2 mod crashes the game after the launch. 

**To Reproduce**
You need "Quick Quack" pack from [CurseForge](https://www.curseforge.com/minecraft/modpacks/quick-quack) and these mods: AbyssalCraft, Advanced Inventory, Advanced Mortars, Advanced World Selection, AnimalNet, Apocalypse Dimension, Armor Toughness Bar, Chainmail Armor Crafting Recipe Mod, ChickenShed 1.1x, Chisel, ConnectedTexturesMod, Duck Craft, Controllable, DeathQuotes / Death Quotes, Draw Bridge, Ducky Mod, Environmental Tech, EverlastingAbilities, EverlastingAbilities-PotionCore, Potion Core, Giselbaer's Durability Viewer, Grid, Hunting Dimension, Item Got, Item Scroller, Just Enough Buttons, JustTheTips, Lightning Rods, Loot Slash Conquer, Modular Powersuits, MystAgrad Cloche Compat, Numina, Overloaded Armor Bar, Real Time Clock, RFTools, RFTools Dimensions, RFTools Power, Storage Tech, The Twilight Forest, Wesley's Roguelike Dungeons, YUNG's Better Caves, **and all of the dependencies of the mods listed** (The latest versions of the listed mods including beta versions relevant to this date: 3.November.2019) +OptiFine_1.12.2_HD_U_F4

**Expected behavior**
I expected the game to launch into the title screen.

**Additional context**
Crash Log:  https://pastebin.com/pxXW2eQK
![javaw_Y1bBQEK63g](https://user-images.githubusercontent.com/57318309/68085427-0c617e00-fe52-11e9-8999-baa3b320843b.png)

**Environment**
Java x64 Build 1.8.0_231-b11  ----  Windows 10
The latest versions of the above listed mods including beta versions relevant to this date: 3.November.2019

- Minecraft Version: 1.12.2
- AE2 Version: appliedenergistics2-rv6-stable-7
- Forge Version: forge-14.23.5.2847

berkkrkc09:
P.S. I was trying to relase a new version to my modpack

mindforger:
[...]

PR #4197, opened 2019-11-11
Better 'dimension ID taken' error message

Pingger:
Now includes the Name of the Dimension, which took our ID.
Should help resolving errors like these more easily: #4189

### DIFF
--- a/src/main/java/appeng/core/Registration.java
+++ b/src/main/java/appeng/core/Registration.java
@@ -197,8 +197,18 @@ final class Registration
 			config.save();
 		}
 		this.storageDimensionID = config.getStorageDimensionID();
-
-		DimensionManager.registerDimension( this.storageDimensionID, this.storageDimensionType );
+		try
+		{
+			DimensionManager.registerDimension( this.storageDimensionID, this.storageDimensionType );
+		}
+		catch (IllegalArgumentException iaexc)
+		{
+			if(DimensionManager.isDimensionRegistered(this.storageDimensionID))
+			{
+				throw new IllegalStateException("Our Dimension was taken by " + DimensionManager.getById(this.storageDimensionID).getName() ,iaexc);
+			}
+			throw iaexc;
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


yueh:
I'd recommend contributing this to forge. It cannot be our job to fix their broken shit.

They could easily use identifieres to register new dimensions similar to everything else like blocks, recipes and so on. Which works perfectly fine and hasn't caused id conflicts since probably forever.

ksx-kshan:
That's true, however, in the interim this would be a friendly way to help users resolve conflicts. Given that the dimension registration will need updated if Forge implements dimension registry anyway, I would have recommended merging this change.

yueh:
If people cannot read "dimension [Number] already registered", they certainly won't read "dimension is taken by [Mod]"

ksx-kshan:
That's definitely a valid point. I yield.

Even knowing which mod is conflicting doesn't help that much because you can easily just change the AE config to use some random higher number and it will most likely then succeed.

Pingger:
Yes, but at the moment we can't point people to the mod, that IS causing the issues. While I will try to get this added to forge, at the moment it isn't and I don't really have the time work on an additional project.
And while people don't read, that the dimension ID is already taken, with this change we can at least point them in the right direction to fixing the issue.

Also: YAY **yueh** is back! Welcome back after your break!

Ommina:
With the change away from integer dimension IDs for 14.0 and above, is it even worth worrying about at this point?

yueh:
Both mods are causing it. There isn't really someone to blame besides maybe the modpack author throwing a pack together without understanding it.

We simply decided to postpone the dimension registration to a later event phase (I think it was also the recommended one for it). So in cases some other mod uses the same id, we can actually break instead of forcing it to load and corrupt the dimensions.

But we could certainly have choosen a better starting id. E.g. some random number, but it's simply too late for this now.

And still, this should be forge notifying the user about which addons are clashing. In case some other mod uses the same approach, but crashes earlier, the users still won't have a clue about it, except this mod also has a custom error reporting.

PR #4220, opened 2019-12-10
Avoid quadratic behavior from rebuilding the entire crafting cache for each interface visited

talchas:
Instead rebuild it at most once per tick, like cpu clusters.

This is particularly important when channels are disabled, since in that case even adding a cable visits every grid node. 

The critical change for that is the one in `public void updateCPUClusters( final MENetworkCraftingPatternChange c )`. The similar changes in add/removeNode provide smaller but still noticeable speedups when removing or adding a cable that actually changes network topology.

This avoids 800ms tick spikes when just placing cables on my 160 interface channels-disabled network.

### DIFF
--- a/src/main/java/appeng/me/cache/CraftingGridCache.java
+++ b/src/main/java/appeng/me/cache/CraftingGridCache.java
@@ -114,6 +114,7 @@ public class CraftingGridCache implements ICraftingGrid, ICraftingProviderHelper
 	private IStorageGrid storageGrid;
 	private IEnergyGrid energyGrid;
 	private boolean updateList = false;
+	private boolean updatePatterns = false;
 
 	public CraftingGridCache( final IGrid grid )
 	{
@@ -136,6 +137,12 @@ public class CraftingGridCache implements ICraftingGrid, ICraftingProviderHelper
 		{
 			this.updateList = false;
 			this.updateCPUClusters();
+		}
+
+		if( this.updatePatterns )
+		{
+			this.updatePatterns = false;
+			this.updatePatterns();
 		}
 
 		final Iterator<CraftingLinkNexus> craftingLinkIterator = this.craftingLinks.values().iterator();
@@ -185,7 +192,7 @@ public class CraftingGridCache implements ICraftingGrid, ICraftingProviderHelper
 		if( machine instanceof ICraftingProvider )
 		{
 			this.craftingProviders.remove( machine );
-			this.updatePatterns();
+			this.updatePatterns = true;
 		}
 	}
 
@@ -219,7 +226,7 @@ public class CraftingGridCache implements ICraftingGrid, ICraftingProviderHelper
 		if( machine instanceof ICraftingProvider )
 		{
 			this.craftingProviders.add( (ICraftingProvider) machine );
-			this.updatePatterns();
+			this.updatePatterns = true;
 		}
 	}
 
@@ -337,7 +344,7 @@ public class CraftingGridCache implements ICraftingGrid, ICraftingProviderHelper
 	@MENetworkEventSubscribe
 	public void updateCPUClusters( final MENetworkCraftingPatternChange c )
 	{
-		this.updatePatterns();
+		this.updatePatterns = true;
 	}
 
 	@Override


yueh:
Many parts in AE are not designed to work with defered removal or insertion. While it might work fine in a short test, it can easily break down in some edges cases, like a chunk being unloaded but it's then accessed an unknown amount ticks later once the removal actually happens.

In general it shouldn't be recommended to disable channels in large server environments. It will simply cause multiple issues from weird behaviour or even stopping to work in certain conditions to performance issues like these.

talchas:
The particularly slow path is already tick-delayed, but I suppose other paths might not be, and depending on the ordering the caches have their handlers called in this could wind up delayed by a second tick.

If delaying the update by a tick could be an issue, just delaying (and coalescing) updates until the completion of PathGridCache's onUpdateTick would do the job, just by having it set a delay flag that CraftingGridCache checks. I suppose there might be other places in the tick that could also be coalesced, but moving it further out seems like more hassle, and this should cover by far the most.

This doesn't _just_ affect channels disabled mode, though it's more limited with channels, since the PathGridCache/etc avoid sending channels-changed events to literally everything in that case, but it still means that making large changes to network topology is unnecessarily slow.

yueh:
Path rebuilding is. Delaying gridnode removal/insertion is not (also not partly). I tried these things a couple times in the past and it was pretty much a guaranteed bug at some point.

I've also tested a setup with a couple thousand patterns. Not specifically for rebuilding the crafting patterns. Mostly in regards to optimizing the memory consumption. And I did not notice any rebuilding slowdown at all.

Pretty much anything which calls something like `postAlterationOfStoredItems()` has to be done immediately. If not it will cause the network inventory to become corrupt over time.

Also there is no bug report. No profiling data. Nothing. Just some wild guesses about the actual issue and trying to fix it by very likely introducing a bug. While it could actually be sponge, some AE2 addon or generic crafting addon acting up. Most AE2 addons are sadly somewhat notorious for performance issues, like PackagedAuto. Or hooking into the recipe events and validating every step against a the recipe registry and if that recipe even exists. By querying every existing items, if they'd accept these ingredients until one returns true (or none).

talchas:
https://sparkprofiler.github.io/#y0biM5i7PV is a spark slow-tick profile of placing cables with channels disabled, which is pretty clear that this quadratic behavior is the most obvious target. I should have posted that with the initial PR message, sorry.

An entirely different fix for the channels-disabled might be to avoid sending ChannelsChanged to every interface, but getting that right seemed trickier, and rebuilding the crafting cache many times when many interfaces were actually being added or removed still seemed unnecessary, though obviously rarer.

If you want a full minimized bug report, I can do that if you you think it would be helpful, but I'd prefer not to put in the effort for the more obvious channels-disabled case if you just don't care about it at all.

yueh:
The profile actually helps to some degree. At least as long as spark reports are usable.

It immediately makes it apparent, that the crafting grid is not the real issue here. Yes, it might be currently most noticable, but it will not stop there. Basically anything item/inventory related will experience these issues at some point.

The downside here is that by the nature of it, it is pretty much impossible to reproduce in a development environment. Or alternatively some mod does something extremely bad to itemstacks in general.

talchas:
I mean, yes, it would be best to not send a ChannelsChanged event to everything in a grid when placing a do-nothing cable, but avoiding that seemed likely to involve completely restructuring how no-channels mode used getUsedChannels/getLastUsedChannels or finding some other way to determine what interfaces/etc actually need to be updated, which seemed difficult. According to the profile interfaces each individually rebuilding the entire crafting cache is 99.7% of the time spent in that update so it doesn't seem likely to be that important so long as that one issue was dealt with. In particular so long as none of their ChannelsChanged handlers wind up looping over the most common grid entities they won't have the same scaling issues.


Here are repros with no other interesting mods, both channels and no-channels:
[ae2 debug 2.zip](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/3986417/ae2.debug.2.zip)

For the no-channels case it's very simple and obvious - just try to place cables connected to the network and you can see the lag (https://sparkprofiler.github.io/#UVIw81jTI7). 

For the channels case it's hard to come up with a reliable setup and needs a larger one, but the channels save there does manage. Deleting and recreating the dense lime cable under the torch does when it has to connect 192 interfaces (https://sparkprofiler.github.io/#whAsqiThjn), but so does removing the two blue cables under torches at the other end (https://sparkprofiler.github.io/#zuoJDdNX7U), for reasons less clear to me.  In either of these cases all the time is still being spent in updatePatterns, and given that just adding or removing a single interface isn't slow, it must still be being called many times. 

In both channels-enabled cases the lag spike is delayed until the ticksUntilReady has timed out so it's harder to see and you can't get it repeatedly tick after tick, and on top of that it requires the breaking/placing cables with a bunch of channels to trigger in the first place, but it still exists.

As additional confirmation, running with my potentially-unsafe fix does eliminate the lag.

Saved copies of the profiles for when spark expires them annoyingly quickly: 
[ae2 spark profiles.zip](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/3986480/ae2.spark.profiles.zip)

yueh:
As said, these thing should really be a bug report about the problem with as much information as possible and not just some random "solution" without reporting the actual issue.

While the pattern rebuild might cause it to do being `O(n^2)`, it isn't the real issue. Or at least not the only possible fix. As well as potentially affecting a couple other storage related system.

I already did a quick PoC yesterday, but didn't hat the option to test it quickly. But essentially I had to revert it, to reproduce the profiling results. Currently I do not have a CI at hand, which can keep artifacts, so I sadly have to revert to building it locally and attaching it here. If that's ok for you.

[Uploading appliedenergistics2-rv6-stable-7.zip…]()

talchas:
Heh, I'd wondered this morning how much caching hashCode somewhere in there might help. The base nochannels save there goes from 800ms to 100ms to place a conduit and the larger channels save converted to no-channels goes from 4500ms to 500ms. (https://sparkprofiler.github.io/#L5jdyiWTfp) With channels enabled the worst I get is around 300ms from that save (https://sparkprofiler.github.io/#zursITERz2). So it _helps_, but the no-channels scaling issue is still visible, so the people who somehow kept going with no-channels at 12s rebuilds would still probably see >1s.

talchas:
Hmm, I tried the "just put a pause flag on Grid/PathGridCache/something" and just pausing during the channel updates helped for the "just placing conduits" case, but when actually altering topology enough time is spent in grid traversals from other places that wind up in updatePatterns (validateGrid add/remove, energy events), often not starting from the _same_ Grid, that it's less pleasant than I'd like. There's an updated version in my fork, but "keep a global set of the caches that need to be updated" is a bit ugly, as is just putting calls to CraftingGridCache.pauseRebuilds in various places. The latter could be be less ad-hoc by sticking the "currently traversing many items" flag and set of callbacks in Grid or something instead, but it's still not great.

yueh:
I am not sure, if these might not be sampling artifacts. E.g. preventing inlining or similar effects.
When microbenchmarking the different parts, the first ones are kinda slow, but after some tries it looks like it goes down by a factor of about 10. Probably once the jvm decides to compile native code for it.

In most cases it appears that caching the hash makes sense. Especially as that's pretty much required immediately after any object is constructed. As long as it's around 50ms or just a bit more per rebuild it think it's fine. ME networks aren't usually changed every second a player is online.

Changing the behaviour is simply not an option. It will break too much things. Locking is probably even worse, which affects every other grid on the server. Especially just to compensate for disabled channels. AE2 wasn't really designed around it and it's more of an afterthought. It will always run into some problems everywhere and should really be treated as experimental option. It's fine for single player or small servers. But when performance is a requirement, it simply isn't an option.

For now better caching or similar approaches are pretty much the only option to maintain stability.

talchas:
Yeah, I mean caching the hash absolutely gives a pretty big speedup for these test maps, it's just not enough to entirely eliminate the noticeable lag as the networks get towards "modpack endgame".

The primary other-mod-visible behavior change I'm suggesting is "get 1 (or maybe 2 or 3 or so) provideCrafting() call per provider when adding or removing from a network, rather than a thousand".  They already can't rely on the exact number or ordering of these, the sort of plausible issue that I could imagine "post MENetworkCraftingPatternChange, expect it to be handled immediately, ditch all the information for provideCrafting() immediately", except that that would already have broken because of the tons of rebuilds.

I'm not suggesting any locking at all, you don't need it here even with my ugly pausing code because everything had better be single-threaded or otherwise locked at a higher level, or else updatePatterns() would already be broken. (If it /was/ multithreaded, the only new lock would be around the set of CraftingGridCaches that need rebuilds, and that would probably want to move to thread-local sets instead)

yueh:
Endgame is always a bit tricky. There is still a good chance, that some addons could affect it majorly and not just the overhead AE2 itself creates with rebuilding it a couple times.

The best approach would probably to redesign the grid/path build. E.g. introduce a `Pre/PostGridPathingEvent` and separate `updatePatterns()` into multiple steps. Like purge the current craftinggrid state in pre, interfaces can still do their local thing during the channels visited event and regenerate the crafting patterns once in post.

But that would be a pretty big API change and I am somewhat hesitant to do it now. As long as it happens during a single tick, I'm pretty confident that it works fine with AE2. But I also have to take addons into account and these probably don't expect something like this suddenly. Especially with some ones being pretty much dead now, but still being used in many modpacks.

talchas:
I mean, there isn't any sensible addon code that could break by only doing ~one updatePatterns per grid traversal instead of tons, since they already are getting many. It's maybe not impossible, but it's not the sort of code pattern that is likely to come up. (Honestly I can't come up with an even halfway sane way to write a pattern provider that would break here)

yueh:
Sensible isn't something I would attribute to most addons. They mostly do some horrible things while surprisingly still work somehow.

I did a quick look into refactoring it into multiple phases, but I do not see a real option for it. There are way to many entry points regarding how the patterns are refreshed. Not just a network rebuild. 
At least not without rewriting a major parts of how grids are internally handled. I would say there is a good chance when doing a rewrite, that it solves the `O(n^2)` here, it will introduce an `O(n^2)` or worse somewhere else, which then affects everyone and not just some servers with disabled channels.

talchas:
Sure, addon code often isn't sensible, but there comes a point when you say "ok, so fix your code or tell people to not randomly update addons in the modpack", because updating dependencies is already something users have to be careful about when using a modpack. And code that breaks from this is well past weird into completely insane.

Issue #125, opened 2014-09-27
Sort by Item ID (not 100 percent sure if it was posted or not)


AlgorithmX2:
Originally Posted by Link19658

This issue was migrated; For original details see http://ae-mod.info/Tracker/1117/

PR #4243, opened 2020-01-06
Sorting by ID

l7ssha:
Added sorting by id, additional metadata and durability. 
Pull request on that topic was closed a long time ago but having such a feature would be very useful and valuable. 
It sorts items in terminal in more natural and similar to jei way. It properly handles blocks which have same id but different metadata (planks, wood, stone, cobblestone, dirt). This feature also sorts tools by it's durability.
image


yueh:
Duplicate of #125

PR #4276, opened 2020-02-05
Fix infinite sources duping arbitrary fluids

Neeve01:
I'm surprised this hasn't been reported before. This is a cross-mod interaction that allows duping arbitrary fluids.

## Synopsys
When plugged to a network with a fluid storage bus, certain infinite sources from other mods can trick the network into duping arbitrary fluids, completely ignoring the fluid they're configured to. This is due to AE2 not checking the actual gathered fluid.

I fixed this by introducing a simple check if the gathered fluid is equal to the requested one.

For reference, see how [Extra Utilities 2](https://github.com/rwtema/Extra-Utilities-2-Source/blob/ff6b75ebc334ee0f1a2c2d1fd1394f8d7e4da0bb/1.10.2/src/main/java/com/rwtema/extrautils2/tile/TileDrum.java#L320) and [Thermal Expansion](https://github.com/CoFH/ThermalExpansion/blob/8ea25f853358bfe2d9f9359bf6d979f0de9e16fa/src/main/java/cofh/thermalexpansion/block/device/TileWaterGen.java#L363) provide infinite fluids.

image

## Steps to Reproduce
Among other mods, this can be easily reproduced with Aqueous Accumulators from Thermal Expansion or Creative Drums from Extra Utilities 2.

**TL;DR: see [this video](https://i.neeve.co/LovelyCooperativeBeardeddragon.mp4).**

1. Install Thermal Expansion (and everything it depends on) or Extra Utilities 2
2. If you've installed Thermal Expansion:
    1. Run the game once to create the config file
    2. Navigate into `instance/config/cofh/thermalexpansion/`
    3. Open `common.cfg` and switch the following to true
```
Device {
    WaterGen {
        # If TRUE, the Aqueous Accumulator will act as an Infinite Source and will also function in the Nether.
        B:Infinite=true
    }
}
```
3. Create a small network consisting of an ME Fluid Interface, ME Fluid Storage Bus, some kind of energy provider
4. Connect either a Creative Drum from Extra Utilities 2 filled with water **OR** an Aqueous Accumulator from Thermal Expansion to the network using the ME Fluid Storage Bus. With AqAc, enable auto-output towards the ME Fluid Storage Bus
5. Grab a bucket of lava and configure the ME Fluid Interface to expose lava

You should see 4 buckets of lava appear inside the ME Fluid Interface despite the network having no lava in storage. This works for any kind of fluid.

### DIFF
--- a/src/main/java/appeng/fluids/parts/FluidHandlerAdapter.java
+++ b/src/main/java/appeng/fluids/parts/FluidHandlerAdapter.java
@@ -114,6 +114,11 @@ public class FluidHandlerAdapter implements IMEInventory<IAEFluidStack>, IBaseMo
 			return null;
 		}
 
+		// Check if the gathered fluid stack is actually equal
+		if ( !gathered.isFluidEqual(requestedFluidStack) ) {
+			return null;
+		}
+
 		if( mode == Actionable.MODULATE )
 		{
 			try


yueh:
This does not really fix anything.

TE simply dupes everything you throw at it. I would recommend reporting it to them as we cannot fix a dupe bug in another mod.

XU2 is not as straightforward as it sounds. It only happens with a creative drum, so you are pretty much complaining about duping already duped items/fluids. Which in my opinion is a bit pointless.
It is a bit tricky due to XU2 kinda violating the `IFluidHandler` contract here, but again this happens only when you are already duping. Probably also something they might have to fix on their side.

Technically we have an issue here because we accidentally convert one fluid into another due to some other mod behaving incorrectly. But this should clearly be fixed at the correct place and not just hide it away by avoiding a bug in another mod.

Neeve01:
Yes, this does fix the dupe.

I don't quite understand why not sanitizing the gathered fluid before propagating it further down the line is not a severe oversight. Why does the code blindly trust whatever fluidstack a Fluid Storage Bus returns?

This behavior isn't limited to Thermal Expansion and ExtraUtils2 (even though I do agree it's not the best mod out there), any mod can trick AE2 into duping anything like that.

yueh:
I would recommend actually testing your patch before claiming it does fix it. E.g. just take the aqueous accumulator and see it still duping lava or any other fluid you want. Linking/Looking at the wrong methods probably is not helpful to understand why it does not do anything at all.

Exaxxion:
As I understand the code linked, TE's Aqueous Accumulator (not a creative item and easily available in popular packs) and XU2's creative drum function as infinite sources by providing a fluid stack of whatever fluid is in their tank (water for the former, whatever is configured for the latter) upon request from the drain method, and AE2 just assumes it has received some completely unrelated fluid, duping it inside of the fluid interface.

This lack of sanitizing on AE's part creates an extremely simple exploit that can and has been replicated unintentionally and destroys balance when used to dupe otherwise difficult fluids to accumulate.

Neeve01:
I have indeed tested the patch with both the Aqueous Accumulator and the Creative Drum before sending a PR, and was unable to reproduce the bug after introducing the mentioned check.

yueh:
Just use the exact same steps you already wrote down. 

In case you want to look at the  [TE code](https://github.com/CoFH/ThermalExpansion/blob/8ea25f853358bfe2d9f9359bf6d979f0de9e16fa/src/main/java/cofh/thermalexpansion/block/device/TileWaterGen.java#L356). 
It simply returns whatever we ask. How do you expect we can decide that the lava it returns when asking for lava is not lava?

Neeve01:
I took a closer look at the function and yes point taken, it seems like Thermal Expansion is basically just ping-ponging the fluid request back at AE2 without changing the actual fluid (or better, creating a new fluidstack) which makes it impossible to sanitize the output from it.

I do have to apologize since I was blinded by the overloaded method below the linked one, which actually takes water inside the AqAc into account and generates a new fluidstack based on it.

Though it would still make sense to sanitize the output.

Exaxxion:
You're right, they do just bounce back the item stack in that drain method if the Aqueous Accumulator is set to be an infinite source. So on that, I stand corrected.

CofH unfortunately has EOL'ed their 1.12.2 as of early last year, and they don't have a copyleft license, so it will be interesting trying to get that fixed. I don't understand the reasoning behind just returning the input item stack without even checking if it matches the fluid the internal tank is set to.

yueh:
No need to apologize.

In terms of TE, it was probably just an oversight and was never reported. Their internal tank does actually check for it.

It would actually also be pretty stupid to needlessly copy a fluidstack and then comparing the reference. Nothing states you cannot optimize it and just adjust the amount of the fluidstack being passed and just change an internal `int` or `long`. No idea if fluidstacks are as bad as itemstacks, but avoiding every unnecessary itemstack copy is pretty sane thanks to forge making them really costly.

As for sanitizing. It might be an idea, but does not fix it. The actual cause is that at some point we merge a water and lava stack into a lava stack, which should not happen and we should probably safeguard against it.

It is also pretty likely that the same thing can happen with items. But it is not likely that many mods will return a completely different item, when asking for something specific. It would break so many things. At some point we simply have to trust other devs, that they adhere to any contract. These simply are not untrusted user inputs. Otherwise we would have to validate anything at any level, because we can no longer trust an `IItemHandler` to not return a different itemstack between `simulate` and `real`. Which can lead to interesting situations like an export bus requesting an item, getting it during simulate but not when actually performing it. And now being left with an itemstack, which the target is not accepting and the source also no longer.

This is actually something similar to what will happen once we fix the conversion. It will extract water from the tank, then no longer being able to merge it with lava (or any other fluid) and in result destroying it. This is something we could then prevent by sanitizing it, but this simply does not fix the underlying behaviour, which could then be still appear with some other part, another AE2 addon, or whatever else.

Exaxxion:
Sure, I can understand the desire to optimize that. Checking if a stack is the same fluid as the internal tank contains just seems like it would be a trivial overhead, and correctness certainly outweighs a slight performance hit.

Neeve has apprised KingLemming of the issue so hopefully we'll see a fix in the near future.

PR #4293, opened 2020-02-23
added japanese translate

aoisensi:
3 hours work!!


yueh:
Nice, really appreciate the effort. 

Let me know, should there be issues like parts being cut off. Not sure, if the longer ones will always fit.

aoisensi:
Sorry. I didn't debug and play.
I tried build and play, but I could not success building.
I'm not Java expert.
Sorry.

underecho:
I'm tried build and check Item translation. It looks good, and helpful for Japanese. 
This translation is can make more better, but it isn't makes any bug.

Good work @aoisensi .

btw, There is some bug in TravisCI and the check does not seem to succeed I think.
Could you fix this? or should I create issue? pls tell me @yueh .
Thank you.

yueh:
I was mostly thinking about it being too long for some GUIs or needing some variables passed. Like the charge level of energy cells in case they might be needed for a different grammar. 

Of course within reasons. The translation system minecraft uses is really simple. Support something like multiple plurals is most likely not possible. (I actually cannot remember using singular and plural for the same translation currently)

Travis failing is simply a feature of their service. Technically I could fix it. But the fix is pretty much the same as just disabling it completely. Making it pretty much useless.

karakufire:
Great work!
But I found some typo.
 * line at 22: "フリックスブロック" → "フルーシュブロック"
 * line at 198: "...作られたケスタスクォーツの種を..." → "...作られたケルタスクォーツの種を..."

And the translation of "セルの表示", line at 420, seems strange, I think that writing like as "表示セル" is more natural.

----

> 
> 
> I'm tried build and check Item translation. It looks good, and helpful for Japanese.
> This translation is can make more better, but it isn't makes any bug.
> 
> Good work @aoisensi .
> 
> btw, There is some bug in TravisCI and the check does not seem to succeed I think.
> Could you fix this? or should I create issue? pls tell me @yueh .
> Thank you.
> 

I applied this PR as a resourcepack, it is correctly showed.

aoisensi:
Thank you for pointing that out!
I fixed at local.
I have a question to @karakufire.
I couldn't come up with a good translation on line 585. (achievement.ae2.ChargedQuartz=ガーン)
Do you have a better idea?

karakufire:
> I couldn't come up with a good translation on line 585. (achievement.ae2.ChargedQuartz=ガーン)

Hmmm....
If I translate this, I might choose from belows:
1. "ショック"
1. "びっくり"
1. "衝撃的な出会い"
1. ~~"You は shock"~~ ←Joke

aoisensi:
Thank you for translate.
Three sounds good for me.

yueh:
This is no longer compatible with 1.15. However we would gladly accept an updated one.

PR #4296, opened 2020-02-24
Create sv_se.lang

Regnander:


yueh:
This is no longer compatible with 1.15. However we would gladly accept an updated one.

PR #4352, opened 2020-04-20
Change remainingItemCount in BasicCellInventory to long to avoid overflows

Shad0wlife:
The underlaying issue was discussed here: https://github.com/ExtraCells/ExtraCells2/issues/684

Large liquid cells (>256k) will fail to store fluids because the itemsPerByte are 8000, causing empty cells to exceed (2^31)-1 remaining items.

since getRemainingItemCount() returns a long anyways the fix is not more than changing the variable type.

Explicitly cast the multiplication to avoid any potential overflowing there as well. The calculation is a long calculation already anyways.

### DIFF
--- a/src/main/java/appeng/me/storage/BasicCellInventory.java
+++ b/src/main/java/appeng/me/storage/BasicCellInventory.java
@@ -175,7 +175,7 @@ public class BasicCellInventory<T extends IAEStack<T>> extends AbstractCellInven
 
 		if( this.canHoldNewItem() ) // room for new type, and for at least one item!
 		{
-			final int remainingItemCount = (int) this.getRemainingItemCount() - this.getBytesPerType() * this.itemsPerByte;
+			final long remainingItemCount = this.getRemainingItemCount() - (long)this.getBytesPerType() * (long)this.itemsPerByte;
 			if( remainingItemCount > 0 )
 			{
 				if( input.getStackSize() > remainingItemCount )


yueh:
This is probably not the right fix.

The basic cells we provide are mainly intended to make it easier to provide the normal cell sizes for new storage channels. So the usual 1k-64k range. Personally I would not even rely on using them for larger cells. They might to work fine at first glance, but I honestly have no idea how it might behave in other situations.

In general for specialised cells, it is better to provide custom cell handlers. Which is a bit more work, but offers way more control.

Shad0wlife:
It is the right fix. AEStack is implemented explicitly using long for a stackSize. Thus, by extension, AEFluidStack is. It even explicitly masks the lower size capacity of the underlying Forge FluidStack. 
All the single line I changed does is disallow adding new items (or fluids) to storage with more than Integer.MAX_VALUE remaining space. This  entirely removes the need for using long stack sizes. Every other calculation is done using long values and works perfectly fine. Simply adding contents to a cell that already has the storage type works flawlessly. I spent hours tracking and logging myself through the code to find this, and I am very certain that this change fixes the large cell sizes.

yueh:
`IAEStack` is not using `long` to store its `MAX_VALUE` inside a cell, but to still be able to handle something storing `Integer.MAX_VALUE` externally (or internally). Like a DSU and a player having more than one completely full. And not to having to convert between an `IAELongStack` for networks and `IAEIntStack` for cells constantly.

This is still no reason to store `Long.MAX_VALUE` inside a single cell as this would force us to switch to `BigInteger` with an decrease in performance and increase in memory usage.

It might be a bit arbitrary, but in my opinion sticking to `int` for basic storage and `long` for aggregated (network) storage is the most straightforward option. Better than reserving a random amount of bits for the aggregated ones and constantly having to validate was it 12, 16, 24, or 42 bit. Or whatever else.

Shad0wlife:
Ah, thank you for the explanation. I did not know that the aggregated storage uses AEStack as well. I thought it was only used in the inventories themselves with eg. AEInventory (or however it was called) then consisting of multiple AEStacks (like the tanks in the FluidInterface being such an inventory iirc. So with what you're saying, if I used and filled ~2^32 * 256k fluid cells, I would also run into an issue because the contents added up would overflow an AEStack?

PR #4353, opened 2020-04-20
Replace the ChunkGenerator for StorageDimension with a primer-based implementation

Shad0wlife:
Accessing and filling the Chunk by directly setting Blockstates can cause issues with mods that change Chunk behaviour (by mixin for example) (This was experienced with a very old version of CubicChunks. It may not apply there anymore, but I still do this PR to make the code cleaner.). 

Also, this is the exact usecase a ChunkPrimer is fit for - set the blocks in locations for a chunk before it is generated.


shartte:
This PR is likely redundant with all the changes that occurred in 1.15.

yueh:
We no longer support any minecraft version older than 1.15 at this point.

Should you be able to confirm that this issue still applies to the most recent version, please update the issue and we will reopen it.

PR #4357, opened 2020-04-24
Fixes #4114: Fix Quartz glass rendering

### LINKED ISSUE #4114: Quartz glass incorrect side rendering
**Describe the bug**
Quartz glass doesn't check the proper block when deciding whether to render a side or not.
If a glass block is 1 block away, it'll not render the face on that side.
If an opaque block is 2 blocks away, it'll not render the face on that side.

**To Reproduce**
image

**Expected behavior**
Below the gold blocks, the side towards us should be rendered.
Below the iron block, the side towards us should be rendered.

**Additional context**
(see above)

**Environment**
Reproduced in SSP with no other mods.
- Minecraft Version: 1.12.2
- AE2 Version: rv6-stable-7
- Forge Version: 14.23.5.2838

Hinara:
Fix #4114

### DIFF
--- a/src/main/java/appeng/decorative/solid/BlockQuartzGlass.java
+++ b/src/main/java/appeng/decorative/solid/BlockQuartzGlass.java
@@ -105,7 +105,7 @@ public class BlockQuartzGlass extends AEBaseBlock
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


covers1624:
Can you show a screenshot of after this change?

PR #4365, opened 2020-04-29
Memory Card Clearing Fix - Removing onItemUse in ToolMemoryCard

nanobyte:
There is a bug in some forge servers where a memory card is set and then cleared at the same time during a sneak + right click. In some cases, reported online, removing an item in your offhand fixes the issue. In others restarting the client helps. In my case, and in other people around the internet, this doesn't work. I looked at the code, and it looks like the clearing logic happens correctly during the "onItemRightClick" call. The "onItemUse" call contains the same logic. I removed the "onItemUse" function, and from testing the memory card will still clear correctly while now functioning normally.

### DIFF
--- a/src/main/java/appeng/items/tools/ToolMemoryCard.java
+++ b/src/main/java/appeng/items/tools/ToolMemoryCard.java
@@ -183,23 +183,6 @@ public class ToolMemoryCard extends AEBaseItem implements IMemoryCard
 	}
 
 	@Override
-	public EnumActionResult onItemUse( final EntityPlayer player, final World w, final BlockPos pos, final EnumHand hand, final EnumFacing side, final float hx, final float hy, final float hz )
-	{
-		if( player.isSneaking() )
-		{
-			if( !w.isRemote )
-			{
-				this.clearCard( player, w, hand );
-			}
-			return EnumActionResult.SUCCESS;
-		}
-		else
-		{
-			return super.onItemUse( player, w, pos, hand, side, hx, hy, hz );
-		}
-	}
-
-	@Override
 	public ActionResult<ItemStack> onItemRightClick( World w, EntityPlayer player, EnumHand hand )
 	{
 		if( player.isSneaking() )


yueh:
As far as I remember we actually need both code paths. The whole interacton system in minecraft is a bit borked. Like it behaves differently when targeting a block vs targeting air. And then if the block decides to have a special behaviour but passes over because it is the held item is not the required one. I think even switching to creative mode will enter a whole different behaviour.

Iirc I even tried to remove one path previously and it prevented the memory card from being cleared in certain conditions.

nanobyte:
@yueh  That makes sense! Well, it doesn't make sense why minecraft is implemented like that, but you know what i mean :P. 

What if I extracted the memory clearing logic out of both of these paths, moving the extraction to happen whenever a shift-right click happens to initially set a p2p tunnel? That way, memory clears up whenever an input is set. It will remove the ability to just clear it outright,  but effectively since it will clear before the setting the input that won't matter.

I understand either way! It seems like this bug is only happening to some people's multiplayer servers.

yueh:
I am actually not certain what part here originates from Minecraft and what is "fixed" by Forge.

There is even `ItemStack#onItemUseFirst()` which is called before `ItemStack#onItemUse` and iirc the whole block activated part. (Technically the block copies the data to the memory card and not the memory card from the block). This should not even affected by the bug with holding something in your offhand. But it is handled way earlier and will prevent any further interaction. So it will probably clear it correctly. But nobody will be able to shift rightclick machines anymore as everything it can do is clearing the card.

As it also only happens in some cases, it really looks more like some mod interaction. E.g. like some protection mod intercepting any block interaction, cancel them, which will then cause the clearing, and afterwards it realises it should allow the player to perform the action and call the block part, too.

Imo the only real fix would be to stop using the same action for multiple things. Like using left click to paste, right click to copy and shift right click to clear. But even that will have issues like you could no longer break blocks when holding a memory card... There is a reason why some tools have some mode switches. However I am pretty certain having to use the mousewheel to switch between copy and paste mod will get very annoying very fast.

PR #4377, opened 2020-05-06
Fixed #4375

### LINKED ISSUE #4375: Autocrafting stuck at "sceduled"
### Mod versions
AE2 version: rv6.stable build 7
JEI version: 4.15.0.291
Mekanism version: 1.12.2-9.8.3.390

### Description
Unable to autocraft elite energy cube (and maybe more but thats not relevant atm)
image

### How to reproduce
- create AE system with autorafting capabilities
- create pattern for advanced and elite energy cube. **These must be done trough JEI**
- place the patterns into an interface connected to a molecular assembler
- add all the needed ingredients to the system (energgy tablet, enriched and reinforced alloy, gold, osmium)
- autocraft an elite energy cube

### Probable casue of the bug
When setting the recipes trough JEI the elite energy cube wants an advanced energy cube with the nbt tag "tier: 1". However when you craft an advanced energy cube it also gets another nbt tag "mekData: { security: 0 }" which very likely causes the system to not use it for autocrafting.

### Expected behavior
As I see it you have 2 options

1. When simulating the crafting also account for nbt tags and just say "missing advanced energy cube" (this will very likely cause many issues with users saying "I cant autocarft even thou I have the pattern")
2. Do the crafting how the simulation works so there are no differences

Edit:
I forgot to mention that if you edit the elite energy cube pattern by placing an crafted advanced energy cube in the center, autocrafting works with no problems.
The reason I still consider this a bug is because creating the crafting plan always works as expected as it somehow ignores the nbt difference (yes the autocrafting pattern for the advanced energy cube also says that it will have 2 nbt tags after crafting)

fdevx:
Now if a crafting plan can be created the item can also be crafted without the crafting cpu getting stuck

I have no clue which design patterns to use in your project so I just produced working code.

If there are any concerns just add a comment and I will try my best to fix it.
Or even better just do it yourself

### DIFF
--- a/src/api/java/appeng/api/networking/crafting/ICraftingPatternDetails.java
+++ b/src/api/java/appeng/api/networking/crafting/ICraftingPatternDetails.java
@@ -24,6 +24,7 @@
 package appeng.api.networking.crafting;
 
 
+import com.google.common.collect.ImmutableList;
 import net.minecraft.inventory.InventoryCrafting;
 import net.minecraft.item.ItemStack;
 import net.minecraft.world.World;
@@ -53,6 +54,17 @@ public interface ICraftingPatternDetails
 	 * @return if an item can be used in the specific slot for this pattern.
 	 */
 	boolean isValidItemForSlot( int slotIndex, ItemStack itemStack, World world );
+
+	/**
+ 	 * @param i the ingredient for which to fetch substitutions
+	 * @return returns all previously checked substitutions for i
+	 * with other words: if you previously called isValidItemForSlot
+	 * the item you provided will be added as a substitution if it was valid
+	 * with this method you can query all these valid substitutions
+	 *
+	 * Note: The first item in the substitution list will always be itself
+	 */
+	ImmutableList<IAEItemStack> getSubstitutionsForIngredient(final IAEItemStack i);
 
 	/**
 	 * @return if this pattern is a crafting pattern ( work bench )
--- a/src/api/java/appeng/api/networking/crafting/ICraftingPatternDetails.java
+++ b/src/api/java/appeng/api/networking/crafting/ICraftingPatternDetails.java
@@ -24,7 +24,6 @@
 package appeng.api.networking.crafting;
 
 
-import com.google.common.collect.ImmutableList;
 import net.minecraft.inventory.InventoryCrafting;
 import net.minecraft.item.ItemStack;
 import net.minecraft.world.World;
@@ -54,17 +53,6 @@ public interface ICraftingPatternDetails
 	 * @return if an item can be used in the specific slot for this pattern.
 	 */
 	boolean isValidItemForSlot( int slotIndex, ItemStack itemStack, World world );
-
-	/**
- 	 * @param i the ingredient for which to fetch substitutions
-	 * @return returns all previously checked substitutions for i
-	 * with other words: if you previously called isValidItemForSlot
-	 * the item you provided will be added as a substitution if it was valid
-	 * with this method you can query all these valid substitutions
-	 *
-	 * Note: The first item in the substitution list will always be itself
-	 */
-	ImmutableList<IAEItemStack> getSubstitutionsForIngredient(final IAEItemStack i);
 
 	/**
 	 * @return if this pattern is a crafting pattern ( work bench )
--- a/src/main/java/appeng/helpers/PatternHelper.java
+++ b/src/main/java/appeng/helpers/PatternHelper.java
@@ -27,7 +27,6 @@ import java.util.Map;
 import java.util.Set;
 import java.util.StringJoiner;
 
-import com.google.common.collect.ImmutableList;
 import net.minecraft.inventory.InventoryCrafting;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
@@ -65,7 +64,6 @@ public class PatternHelper implements ICraftingPatternDetails, Comparable<Patter
 	private final boolean canSubstitute;
 	private final Set<TestLookup> failCache = new HashSet<>();
 	private final Set<TestLookup> passCache = new HashSet<>();
-	private final Map<IAEItemStack, Set<IAEItemStack>> ingredientSubstitutionCache = new HashMap<>();
 	private final IAEItemStack pattern;
 	private int priority = 0;
 
@@ -93,7 +91,6 @@ public class PatternHelper implements ICraftingPatternDetails, Comparable<Patter
 		{
 			NBTTagCompound ingredient = inTag.getCompoundTagAt( x );
 			final ItemStack gs = new ItemStack( ingredient );
-			IAEItemStack aeItemStack = AEItemStack.fromItemStack(gs);
 
 			if( !ingredient.hasNoTags() && gs.isEmpty() )
 			{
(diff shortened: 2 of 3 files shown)


yueh:
The obvious one, we can no longer break the API this late into 1.12

The cache also has to take into account that it will contain many million entries for a single pattern plus potentially duplicated for every single slot. Also patterns are not deduplicated, so every single pattern will maintain their own version of it.

Might sound exaggerated but I already had the issue with EnderChest registering every single instance of their `16^3` different options into the oredict and causing our lookup table to explode to `4096^2` entries. Further due to the nature of usually affecting items with NBT, these are especially large objects and will be hold forever on a server, constantly growing over time.
E.g. in case of a advanced mekanism energy cube we have to at least take 1 million entries for the energy tablets and 2 million for the basic one into account. Potentially different caches for each tablet. Assuming 512 byte per instance, probably not enough but there are at least 2 `long` and 4 `int` on our side plus reference and not even considering the size of NBT. So easily in the range of 2GB memory per pattern. Without taking any other nbt tag into account, which can multiple these entries easily by an unknown factor.

fdevx:
I agree that these lookup tables will explode I initially planned to have a separate instance or something similar for every crafting job but the amount of code I would have to change got out of hand pretty quickly (especially if there are very few comments and the methods are pretty large).

Another option is to add the accepted item to the crafting node but then I would have to touch the slightly over sized executeCrafting Method in CraftingCPUCluster.java (this file imo needs a complete rewrite because its way too complex).

so I stuck with this admittedly shitty implementation.

Imo the best solution would be to (somehow) add a cache to CraftingTreeNode that stores what items were an acceptable replacement and then pass an entire CraftingTreeNode to executeCrafting instead of just the pattern.

For now I would assume deduplicating the PatternHelper should prevent such a massive explosion of the cache. Because there needs to be a pattern for an item in order for it to be considered for the autocrafting. Also the cache does not get updated on a call to isValidItemForSlot only if getOutput gets called which means it will only be updated for autocrafting.

I understand that you cant break the api so please consider some changes to CraftingTreeNode and CraftingCPUCluster for future versions inorder to make a better implementation of this possible

yueh:
With 1.15 it would be way easier. The cache would probably be even unecessary at all because we can handle item by item ids alone, fuzzy would actually only need to support actual damagable items and otherwise tags are just a list of item ids without any NBT overhead.

Deduplicating is also not that straigthforward. We cannot deduplicate based on outputs but have to use inputs as different players might (or have to) prefer certain items. Otherwise they would interfere with eachother. And that is simply not an easy approach. I already did an approach because I was a bit annoyed at the memory consumption of 10-20 MB or so with a couple thousand duplicated patterns. But the effect would actually pretty minimal as most patterns are actually different and for the maybe 10 duplicated patterns per network it isn't worth the effort especially with some addons being incompatible.

One really simple idea might be to aggregate the result of precise and fuzzy for the fuzzy path. This might solve issues like the TE satchels. But probably not for mekanism. But honestly mekanism causes so many issues, that I am very reluctant to support it in any way.

fdevx:
The Mekanism was the easiest to solve all the stuff you see with the cache was for the TE capacitor

all Mekanism needed was the change I made in  executeCrafting (the added fuzzy search if the precise search didnt find it)

Also I migth get away with removing the chache all together and just adding the precise search for the TE capacitor.

I will try this in a few hours wen I have time and update it if it works.

fdevx:
Good news the cache is not needed (at least for the test items I used) and it looks like it is very unlikely that there is an item that still falls trough all the checks.

I would assume since 1.15 features sound amazing that there will be no need for further bug fixes regarding "autocrafting gets stuck" 

I'll push the changes in a minute and it would be super awesome if you could take another look at it

yueh:
Honestly I am not really sure about it. It pretty much feels like a major hack. Which ultimately will introduce some pretty strange behaviour.

Basically there won't be any non oredict crafting at all. It is no longer possible to deactiviate it.
I am not aware of any existing recipe, but I would be surprised if this case does not exist.
The basic idea is that some recipe uses certain items, which are oredict with eachother but needs a precise one for a specific slot. Worst case it can now use the exact item as replacement for a different slot and then run out of ingredients once it reaches the fixed one. Thus stalling it again. It probably needs a couple other preconditions to happen. But still we replace a system a player can fix by using the right item for a pattern and now have something they can no longer fix on their side at all.

My current plan is also to basically release a final 1.12 build in case I actually manage to find time for porting to 1.15. And with the current build setup we will pretty much lose any option to release a 1.12 build afterwards in case bugs appear later on.

fdevx:
This system does not interfere with the players ability to fix it since it only kicks in if the job doesn't find an item.
And may I remind you of TE capacitors which cant be fixed by players atm.
Sure with some recipes there will be the bug that you described but this can be fixed by the player using the right item because the system always searches for the precise item first and if it is unable to fin it it tries to fuzzy search the crafting cpu inventory.

But yes I agree its a major hack just like the rest of ae2 autocrafting.

yueh:
The search when it is missing is exactly the case I am considering. With multiple co procs it will issue multiple task at once, which potentially can queue up one task, which is still waiting on another one to complete, yet it found some replacements via a fuzzy search which will then be missing for another task.

fdevx:
But a I said the player can fix it by providing the exact item in the pattern

yueh:
Imagine items `A`, `B`, and `C` provided by recipes with `? -> 2 * item`
According to the oredict `B` can be replace with `A` (but `A` not with `B`)
These need to be used in a recipe with `A, B, C -> ?`.

Due timing the tasks for `A` and `C` are done and there is a slot open to enqueue the task for the final product. This is now missing a `B` but luckily found an `A` as replacement. Which then leaves `2*B` and `C` for the next task and stalls. Probably needs different amounts, various timing problems and so on. But essentially it is replacing a known behaviour caused by a buggy item (TE capacitors have the same issue as satchels) with probably a race condition.

fdevx:
you know what - since I reported the issue you closed it within an hour. It took me like 5 comments and a day to even convince you this is indeed a bug in ae2 and you still didnt reopen the issue.
I think I'm done with it.

If you dont like this solution then come up with your own.
Imo this autocrafing code is a huge mess with methods longer than all the rivers on earth combined.
If you want a real fix you need to rewrite the whole autocrafting system.

Also I understand that this racecondition could be bad but to get there you had to invent 2 new items that are the same and at the same time aren't which probably casues worse things in this autocrafting code than "just" a race condition.
But even if these items existed you could always just say

> Please report this directly to your modpack. In many it actually is the intended behaviour.

yueh:
I really understand your frustration. You pretty much had the misfortune of reporting something which usually is caused by fake items or addons. In one case I even talked with the dev before and raised my concerns about this mod idea causing issues and I cannot provide support for it. It simply adds to many unverifiable variables to it.

You are pretty much the first one actually providing some details and some really useful feedback.

The mekanism issue is really a mixed bag. We would have to ignore NBT for it, but cannot really do it.
And it is already a potential performance issue. So being more lenient will sadly not be ideal.

The TE satchel/capacitor is honestly completely new to me. I never expected an item that behaves like them to exist. And this goes even much deeper than autocrafting. E.g. an bus with a fuzzy card also cannot handle them besides the basic and harded one. And not just the type but also charge level, which to some degree is reported as damage percent.
In case a player has the great idea to have a system export any capacitor, which is not fully charged, it will also export any fully charged one and only up to the hardened ones. In case someone expects it to only export uncharged resonant ones, they are in for a surprise.

Honestly no idea how to handle it currently. But if there is a way, it will also fix autocrafting for them.

And I mostly agree with the autocrafting behing a mess, which needs cleanup. Sadly adding a fix to it, which just hides the real problem won't be ideal. It will just adding more mess, which in the end is then useless when fixing the actual one, but still sitting around and making it harder to maintain.

So when considered as bugs, these are actually two different ones, which also need different fixes.

PR #4379, opened 2020-05-08
Fix README.md linking to contributing guidelines

Pasarus:
A very simple change in the Repo, looking for these guidelines and came across this error, it annoyed me so I fixed it 😄 

Any build failures should in theory be spurious, so apologies as it will probably set off a build.


yueh:
Thanks, but this is hopefully not needed much longer.

PR #4400, opened 2020-06-03
Re-implement materials.

eutropius225:
This will break things elsewhere that relied on the enum for switches.


yueh:
Thanks, but there is currently no decision made or even considered on how to deal with materials (or parts).

All I can say for noww is that this is exactly how it will not happen and I do not want to be forced to revert it at some point, including every other commit based on it.

PR #4441, opened 2020-06-30
Minor Cleanups / Refactoring

shartte:


src/main/java/appeng/me/cache/TickManagerCache.java
```diff
@@ -66,7 +60,7 @@ public long getAvgNanoTime(final IGridNode node) {
             return -1;
         }
 
-        return tt.getAvgNanos();
```
shartte, line 69: Since this method always return 0, it kinda draws the entire point of this method into question...

src/main/java/appeng/core/features/registries/GridCacheRegistry.java
```diff
@@ -18,54 +18,57 @@
...
-            this.caches.put(iface, implementation);
-        } else {
-            throw new IllegalArgumentException(
-                    "Invalid setup, grid cache must either be the same class, or an interface that the implementation implements. Gotten: "
-                            + iface + " and " + implementation);
+    public synchronized <T extends IGridCache> void registerGridCache(@Nonnull Class<T> iface,
+            @Nonnull IGridCacheFactory<T> factory) {
+
+        if (registry.stream().anyMatch(r -> r.cacheClass.equals(iface))) {
+            AELog.debug("Overriding grid cache factory for cache type %s", iface);
```
yueh, line 43: This should really crash not just log some debug statement. Replacing core behaviour should never be possible for addons.

src/main/java/appeng/me/Grid.java
```diff
@@ -52,6 +52,7 @@ public Grid(final GridNode center) {
 
         final Map<Class<? extends IGridCache>, IGridCache> myCaches = AEApi.instance().registries().gridCache()
                 .createCacheInstance(this);
+        this.caches = new HashMap<>(myCaches.size());
```
yueh, line 55: `IdentityHashMap` might be an idea here. Guava also provides a special class to instance map. Not sure if it might provide some additional benefits. E.g. creating a copy with new instances.

src/main/java/appeng/me/cache/TickManagerCache.java
yueh: I think this was never finished. I might have improved it when considering adding time dilation to grid ticking, but that was never merged. One of the debug tool makes use of it, but currently it is useless.

src/main/java/appeng/core/features/registries/GridCacheRegistry.java
shartte: Before this change, it just silently swallowed it :-D
So we're in agreement, I'll push the change.

PR #4442, opened 2020-06-30
Reduce Importance of PartType and Fix Upgrade Card Tooltip Grouping

shartte:
Removed reliance on PartType from code outside of Bootstrapping, and restored ability to group different supported items on
Upgrade Card tooltips.


shartte:
Git didn't get the move of PartType and converted it to delete+add, sadly.

PR #4445, opened 2020-07-01
Changed AEGlassMaterial to Materials

yueh:
Preparation in case we need more to support custom non solid, non
opaque, or similar blocks.
Removed custom fullSize/opaque blocks in favour of Material

### DIFF
--- a/src/main/java/appeng/decorative/solid/QuartzGlassBlock.java
+++ b/src/main/java/appeng/decorative/solid/QuartzGlassBlock.java
@@ -23,7 +23,7 @@ import net.minecraft.block.BlockState;
 import net.minecraft.block.material.Material;
 import net.minecraft.util.Direction;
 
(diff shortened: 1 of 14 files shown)


src/main/java/appeng/decorative/solid/QuartzGlassBlock.java
```diff
@@ -34,7 +34,7 @@ public QuartzGlassBlock(Properties props) {
     @Override
     public boolean isSideInvisible(BlockState state, BlockState adjacentBlockState, Direction side) {
         final Material mat = adjacentBlockState.getMaterial();
-        if (mat == Material.GLASS || mat == AEGlassMaterial.INSTANCE) {
+        if (mat == Material.GLASS || mat == Materials.GLASS) {
```
shartte, line 37: I don't like this because it's hella confusing. Material vs. Materials :-D

PR #4459, opened 2020-07-10
Rewrite drive model to support custom cell models

yueh:
Each cell can now register a mapping Item -> ResourceLocation for a
custom cell model. Fallback will always be a 1k item cell.


src/api/java/appeng/api/client/ICellModelRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.util.Map;
+
+import javax.annotation.Nonnull;
+
+import net.minecraft.client.renderer.model.ModelBakery;
+import net.minecraft.item.Item;
+import net.minecraft.util.ResourceLocation;
+
+public interface ICellModelRegistry {
```
shartte, line 34:
```suggestion
/**
 * A registry for 3D models used to render storage cells in the world, when they are 
 * inserted into a drive or similar machines.
 */
public interface ICellModelRegistry {
```

src/api/java/appeng/api/client/ICellModelRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Register a new model for an item baking a cell.
```
shartte, line 37:
```suggestion
     * Register a new model for a storage cell item.
```

src/api/java/appeng/api/client/ICellModelRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+     * 
+     * Model loading is not part of it at all. The only use case it provides is to
```
shartte, line 39:
```suggestion
     * <p>
     * You are responsible for ensuring that the given model is actually loaded by the game. 
     * See {@see  net.minecraftforge.client.model.ModelLoader#addSpecialModel}.
```

src/api/java/appeng/api/client/ICellModelRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+     * map an {@link Item} to a {@link ResourceLocation} which can be looked up from
```
shartte, line 40:
```suggestion
     * This method only maps an {@link Item} to a {@link ResourceLocation} which can be looked up from
```

src/api/java/appeng/api/client/ICellModelRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+     * the {@link ModelBakery}. No validation about missing models will be done.
+     * 
+     * For examples look at our cell part models within the drive model directory.
+     * 
+     * @param item  The cell item
+     * @param model The {@link ResourceLocation} representing the model.
+     * @return
+     */
+    boolean registerModel(@Nonnull Item item, @Nonnull ResourceLocation model);
```
shartte, line 49: I think the return value is redundant. It can only ever return true.

src/api/java/appeng/api/client/ICellModelRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * The model represented as {@link ResourceLocation} for an {@link Item}
```
shartte, line 52:
```suggestion
     * The {@link ResourceLocation} of the model used to render the given storage cell {@link Item}
     * when inserted into a drive or similar.
```

src/api/java/appeng/api/client/ICellModelRegistry.java
```diff
@@ -0,0 +1,68 @@
...
+     * 
+     * @param item
+     * @return
```
shartte, line 55:
```suggestion
     * @return A default model if no particular model is registered.
```

src/main/java/appeng/core/api/client/ApiCellModelRegistry.java
```diff
@@ -0,0 +1,106 @@
...
+
+        return this.registry.put(item, model) != null;
+    }
+
+    @Override
+    @Nonnull
+    public ResourceLocation model(@Nonnull Item item) {
+        Preconditions.checkNotNull(item);
+
+        return this.registry.get(item);
```
shartte, line 97: This violates the annotation. Should return a default model (any, really), if none is registered.

yueh: Would certainly make sense here.

yueh: Correction: Currently it's even unused. Mostly in case some addon might want to lookup existing ones. So might not be ideal to return a default one, which is shared with another one like a 1k cell. Otherwise there is no way to tell, if there is a custom model or not.

src/api/java/appeng/api/client/ICellModelRegistry.java
yueh: Return is simply to handle duplicates getting reject and a addon could handle it in some way. `model()` might be able to replace it, but only if it either returns `null` or a well defined default, which is always the default and not shared with a real cell.

src/main/java/appeng/core/api/client/ApiCellModelRegistry.java
```diff
@@ -0,0 +1,107 @@
...
+        this.registry = new IdentityHashMap<>();
+        this.registry.put(definitions.items().cell1k().item(), MODEL_CELL_ITEMS_1K);
+        this.registry.put(definitions.items().cell4k().item(), MODEL_CELL_ITEMS_4K);
+        this.registry.put(definitions.items().cell16k().item(), MODEL_CELL_ITEMS_16K);
+        this.registry.put(definitions.items().cell64k().item(), MODEL_CELL_ITEMS_64K);
+        this.registry.put(definitions.items().fluidCell1k().item(), MODEL_CELL_FLUIDS_1K);
+        this.registry.put(definitions.items().fluidCell4k().item(), MODEL_CELL_FLUIDS_4K);
+        this.registry.put(definitions.items().fluidCell16k().item(), MODEL_CELL_FLUIDS_16K);
+        this.registry.put(definitions.items().fluidCell64k().item(), MODEL_CELL_FLUIDS_64K);
+    }
```
shartte, line 82: What about the creative cell?

PR #4473, opened 2020-07-17
Added a priority to IAEPowerStorage

yueh:


src/api/java/appeng/api/networking/energy/IAEPowerStorage.java
```diff
@@ -71,4 +71,17 @@
      */
     @Nonnull
     AccessRestriction getPowerFlow();
+
+    /**
+     * The priority to use this energy storage.
+     * 
+     * A higher value means it is more likely to be used.
```
shartte, line 78:
```suggestion
     * A higher value means it is more likely to be extracted from first,
     * and less likely to be inserted into first.
```

yueh: How it is interpreted is up to the actual use case. That particular behaviour is really common within any priority within AE2. Or might not be correct at all depending on the use case. Across multiple chained networks it will certainly behave differently.

shartte: Well that's why I said more likely. But it's just not correct as it stands. Higher priority does **not** mean it's more likely to be used for power insertion.

yueh: Yes, the priority concept is just used all over the place in a similar manner. Would be really nice to have value types and be able to extract it and document it in a central location instead of copy&pasting it all over the code.

PR #4475, opened 2020-07-18
Added new japanese translate

aoisensi:
converted


aoisensi:
yay

yueh:
I would have added it earlier but the buildserver for 1.12 has some issues and I am not be able to make a build for it.

Issue #3546, opened 2018-06-17
Texture/Model cleanup


yueh:
The current folder/json structure for models and textures is not following some common conventions.

Examples:
* Texture aliases inside terminal models are named in various different ways like `front_bright`, `lightsBright`, or `l1`
* Some textures are scattered all over the place, like some textures for the same part/item/block are located in `textures/parts` and `textures/items`

This should be changed for 1.13 to follow some basic principles.

* Avoid duplicate textures/models at all costs. Some parts/blocks/items use different models with different textures, which actually are just copied around. E.g. the crafting terminal.
* Aliases for an equal layer/part mode/etc should always follow the same name. E.g. `layer_bright` for terminals.
* Models with more than 2 textures should ideally reside in their own folder otherwise at least group by the next higher feature level. E.g. `models/parts/buses`.
* Textures for the same feature should be grouped together. E.g. everything crafting terminal related in `textures/parts/terminals/crafting`
* Exception for grouped assets can be commonly shared one like channel indicators should reside in an appropriate folder. E.g. `textures/parts/shared`. This might be tentative depending on how much freedom we can/want give to resource packs.

fscan:
* Keep an eye out for forge/minecraft changes. I would really like to rip out our glowing texture hack.

PR #4483, opened 2020-07-19
Texture Cleanup

shartte:
Relates to #3546


shartte:
This is probably not really reviewable :-D

PR #4489, opened 2020-07-21
Don't use PacketBuffer#readString since it's client-only

### LINKED ISSUE #4488: Crash when Switching Sort By in Crafting Terminal
When switching the sort by in the crafting terminal, causes server to crash.

Version: 7.0.0-alpha.8

[crash-2020-07-20_20.21.00-server.txt](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/4950573/crash-2020-07-20_20.21.00-server.txt)

shartte:
Fixes #4488

### DIFF
--- a/src/main/java/appeng/core/sync/BasePacket.java
+++ b/src/main/java/appeng/core/sync/BasePacket.java
@@ -32,6 +32,15 @@ import appeng.core.sync.network.INetworkInfo;
 import appeng.core.sync.network.NetworkHandler;
 
 public abstract class BasePacket {
+
+    /**
+     * Sadly {@link PacketBuffer#readString()} gets inlined by Proguard which means
+     * it's not available on the Server. This field has the default string length
+     * that is used for writeString, which then also should be used for readString
+     * when it has no special length requirements.
+     */
+    public static final int MAX_STRING_LENGTH = 32767;
+
     private PacketBuffer p;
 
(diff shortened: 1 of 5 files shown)


src/main/java/appeng/core/sync/BasePacket.java
```diff
@@ -32,6 +32,14 @@
 import appeng.core.sync.network.NetworkHandler;
 
 public abstract class BasePacket {
+
+    /**
+     * Sadly {@link PacketBuffer#readString()} gets inlined by Proguard which means it's not available on the
+     * Server. This field has the default string length that is used for writeString, which then also should
+     * be used for readString when it has no special length requirements.
+     */
+    public static final int MAX_STRING_LENGTH = 32767;
```
yueh, line 41: I would actually consider reducing this to a more sensible amount. This will actually allocate a buffer of about 131kb for each string. Something in the range of 20-100 chars is probably more than enough for us.

shartte: There is one that synchronizes JSON based data between containers. So I was hesistent. The default for WriteString() is 32k too

shartte: oh, also: no, the buffer is not allocated based on this value, this is just a malicious client security feature.

PR #4509, opened 2020-07-26
When displaying an encoded processing pattern, clear slots not occupied by the pattern's outputs

### LINKED ISSUE #4508: Placing an Encoded Pattern in the Pattern Terminal does not Clear first
When placing an encoded pattern in the pattern terminal, the existing inputs and outputs should be cleared. Right now, only the non-empty slots occupied by the crafting patterns input/outputs will be overwritten, but existing filled slots will keep their setting.

shartte:
Fixes #4508


shartte:
Superseded #4511

PR #4570, opened 2020-08-08
One Spatial Dimension

shartte:
This may need further work, but should make the storage dimension work.


src/main/java/appeng/spatial/SpatialDimensionManager.java
```diff
@@ -18,182 +18,98 @@
...
-                String numericIdPart = path.substring(DIM_ID_PREFIX.length());
-                maxId = Math.max(Integer.parseUnsignedInt(numericIdPart), maxId);
-            } catch (NumberFormatException e) {
-                AELog.warn("Unparsable storage cell dimension id '%s'", path, e);
-            }
+    public ServerWorld getWorld() {
+        MinecraftServer server = getServer();
+        ServerWorld world = server.getWorld(WORLD_ID);
+        if (world == null) {
+            throw new IllegalStateException("The storage cell world is missing.");
```
yueh, line 68: Probably shorten these with `PreConditions.checkState(world == null, "The storage cell world is missing.")`.
A static import here might even be an idea.

src/main/java/appeng/spatial/StorageCellLot.java
```diff
@@ -0,0 +1,85 @@
...
+     * Id of the lot.
+     */
+    private final int id;
+
+    /**
+     * The storage size of this dimension. This is dicateted by the pylon structure
+     * size used to perform the first transfer into this dimension. Once it's set,
+     * it cannot be changed anymore.
+     */
+    private final BlockPos size;
```
yueh, line 22: Is this absolute or relative to the origin? Assuming that uses a `BlockPos` to encode a `Vector3i`-like point.

src/main/java/appeng/spatial/StorageCellLot.java
```diff
@@ -0,0 +1,85 @@
...
+    private final int owner;
+
+    public StorageCellLot(int id, BlockPos size, int owner) {
+        this.id = id;
+        this.size = size;
+        this.owner = owner;
+        if (size.getX() < 1 || size.getY() < 1 || size.getZ() < 1) {
+            throw new IllegalArgumentException("Lot size " + size + " is smaller than minimum size.");
+        }
+        if (size.getX() > MAX_SIZE || size.getY() >= MAX_SIZE || size.getZ() >= MAX_SIZE) {
```
yueh, line 37: This could lead to a couple issues in case some addon would add different cells. Multuple 512x256x512 cells would create an area without any border to the next one.

Assuming `size` is still the current variant with representing the area within the pylons and not the size of the pylon structure. Also technically `getY` would be limited to 254 (256) blocks or less depending on the set build height.
But that might be implicitly prevent as the inner area is always smaller than the pylon area, thus always 2 blocks smaller then the limit.

src/main/java/appeng/spatial/StorageCellLot.java
```diff
@@ -0,0 +1,85 @@
...
+            throw new IllegalArgumentException("Lot size " + size + " exceeds maximum size of " + MAX_SIZE);
+        }
+    }
+
+    public int getId() {
+        return id;
+    }
+
+    public BlockPos getSize() {
```
yueh, line 46: Depending on absolute or relative values, the other one as additional helper might be worth it. Iirc some values use the relative size, while others like moving might benefit from having an easy access to the absolute limit.

src/main/java/appeng/spatial/StorageCellLot.java
```diff
@@ -0,0 +1,85 @@
...
+        return size;
+    }
+
+    public int getOwner() {
+        return owner;
+    }
+
+    public BlockPos getOrigin() {
```
yueh, line 54: This probably needs some documentation on how it layouts the lots. This appears to at least try and use all 4 direction, which is pretty much a requirement. As far as I am aware, minecraft still uses floats for positions and at some point the start to get into rounding errors. This is usually not noticable, as players won't easily reach coordinates in the range of 100k or even more. But with cells someone could just loop through 1k cells and end up half a million blocks away from the world origin. Which might start to break down physics like collisions.

src/main/java/appeng/spatial/StorageCellLot.java
```diff
@@ -0,0 +1,85 @@
...
+        // mirror in one of 4 directions
+        if ((signBits & 0b01) == 0) {
+            posx *= -1;
+        }
+        if ((signBits & 0b10) == 0) {
+            posz *= -1;
+        }
+
+        // offset from cell center
+        posx -= 64;
```
yueh, line 79: Might be worth to move this to `[0,0,0]` or `[1,1,1]` of the region origin. Also will not clash with a potential 512x256x512 cell overwriting 192 blocks of an adjacent lot.

src/main/java/appeng/spatial/StorageCellLot.java
```diff
@@ -0,0 +1,85 @@
...
+        posz -= 64;
```
yueh, line 80: Same

src/main/java/appeng/spatial/StorageCellLot.java
```diff
@@ -0,0 +1,85 @@
...
+
+        return new BlockPos(posx, 64, posz);
```
yueh, line 82: 256 world height limit + 64?

src/main/java/appeng/spatial/StorageCellWorldData.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    public StorageCellLot getLotById(int id) {
+        return lots.get(id);
+    }
+
+    public StorageCellLot allocateLot(BlockPos size, int owner) {
+
+        int nextId = 1;
+        for (int id : lots.keySet()) {
```
yueh, line 48: Might be worth to have an additional counter, which is initialized when reading the list of lots.
But for a small amount of cell it should be fine.

src/main/java/appeng/spatial/StorageCellLot.java
shartte: I actually used picked the MAX_SIZE that was already present in 1.12, but never actually checked against. I suppose we should probably rename MAX_SIZE to PLOT_SIZE and add a SAFETY_MARGIN. Then we'd check against PLOT_SIZE - SAFETY_MARGIN ?

yueh: Safety margin would always be 1 block, or 2 overall. When taking the pylon structure into account, it would provide that by itself and we just have to limit it to `512x256x512`.

Also we'd still need `512` to align it with region files. So more like `PLOT_SIZE = REGION_SIZE - 2`. And then of course ensure it will always have at least 1 wall of matrix frames above and below. And hopefully you cannot glitch through 1 wide walls with an ender pearl.

shartte: If we can, the safety margin should ofcourse be such that you cannot see your neighbors. 1 block is certainly not sufficient for that.

PR #4577, opened 2020-08-09
Fix Special Model Registration

### LINKED ISSUE #4574: Pure Fluix -> Dust Duping Recipe?
**Describe the bug**
Current fluix dust recipe enables an infinite duping of pure fluix.

**To Reproduce**
Grind pure fluix at the grinder into fluix dust.  Use dust to create 2 seeds.  Repeat for infinite pure fluix.

**Expected behavior** Fluix dust recipe should require item:fluix_crystal rather than tag:crystals/fluix

- Environment: 
MC 1.15.2
Forge 31.2.31
AE 7.0.0-beta.2

### LINKED ISSUE #4575: Molecular Assembler Texture error - MC 1.16.1
The Molecular Assembler Texture show error and flashing.
image

**To Reproduce**

In the new world places a Molecular Assembler.

**Environment**

1.16.1-forge-32.0.105
appliedenergistics2-8.0.0-alpha.2

image

shartte:
Fixes #4575 by updating Forge (ModelRegistryEvent is a modbus event but was misflagged as not being one). And update Forge because their breaking change breaks our mod.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,7 +11,7 @@ artifact_basename=appliedenergistics2
 minecraft_release=1.16
 minecraft_version=1.16.1
 mcp_mappings=20200723-1.16.1
-forge_version=32.0.93
+forge_version=32.0.107
 
 #########################################################
 # Provided APIs                                         #
--- a/gradle.properties
+++ b/gradle.properties
@@ -11,7 +11,7 @@ artifact_basename=appliedenergistics2
 minecraft_release=1.16
 minecraft_version=1.16.1
 mcp_mappings=20200723-1.16.1
-forge_version=32.0.107
+forge_version=32.0.108
 
 #########################################################
 # Provided APIs                                         #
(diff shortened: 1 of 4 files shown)


gradle.properties
```diff
@@ -11,7 +11,7 @@ artifact_basename=appliedenergistics2
 minecraft_release=1.16
 minecraft_version=1.16.1
 mcp_mappings=20200723-1.16.1
-forge_version=32.0.93
+forge_version=32.0.107
```
yueh, line 14:
```suggestion
forge_version=32.0.108
```

Their breaking change in 107 appears to be a bit premature and causes the event to fire way too often.

src/main/resources/META-INF/mods.toml
```diff
@@ -16,7 +16,7 @@ description="A Mod about Matter, Energy and using them to conquer the world.."
 [[dependencies.appliedenergistics2]]
     modId="forge"
     mandatory=true
-    versionRange="[32.0.75,33.0.0)"
+    versionRange="[32.0.107,33.0.0)"
```
yueh, line 19:
```suggestion
    versionRange="[32.0.108,33.0.0)"
```

Same

PR #4595, opened 2020-08-14
Simple Texture Update

Ridanisaurus:
An early version of some of the textures I wanted to update for my Resource Pack. Mostly a Dust Update, and Single-Colour Grey Border removal. Trying to bring the new style of Minecraft Textures, while keeping the iconic shapes, and colours of the original AE2.


Ridanisaurus:
Outdated.

PR #4610, opened 2020-08-17
AE2 Retexturing

Ridanisaurus:
First Package ready for your testing, and approval.
Package contains:
- Items


shartte:
👍 This looks great.

I was about to complain that the fluix pearl is offset 1px to the lower right, but I noticed it's the same for vanilla's ender pearl.

Minor addition we can do, but I can do this later on my own: We can make the charged staff stop animating when its uncharged.

shartte (approved):

Ridanisaurus:
If I recall correctly, the first frame has no spark on it. If you want a non-animated version of it, just let me know.

yueh:
I moved to a separate branch, so we do not have to make a bugfix release with some new textures scattered around.
So this should be the always the target for further additions.

And it should be easier to backport it to 1.15 or 1.16.1, if desired.

yueh (approved):

PR #4616, opened 2020-08-18
Fix wrenches not harvesting blocks

### LINKED ISSUE #4584: Many block can't using Wrench to the drop - MC 1.16.1
**Describe the bug**

Many block (example: ME interface, ME Drive, Charger ...) can't using `Wrench` + `Shift` + `Mouse right click` to the drop.
image

**To Reproduce**

In the new world place a ME Interface, using wrench to operate.

**Environment**

1.16.1-forge-32.0.105
appliedenergistics2-8.0.0-alpha.2
image

shartte:
Fixes #4584: Wrench is not calling block use functions to harvest blocks

p.s.: Due to a change how Forge `onBlockHarvested` works, this will now play a block break sound + particles.

### DIFF
--- a/src/main/java/appeng/items/tools/quartz/QuartzWrenchItem.java
+++ b/src/main/java/appeng/items/tools/quartz/QuartzWrenchItem.java
@@ -25,6 +25,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.item.ItemUseContext;
 import net.minecraft.util.ActionResultType;
 import net.minecraft.util.math.BlockPos;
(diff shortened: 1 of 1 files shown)


yueh:
This probably needs some testing. If `doesSneakBypassUse` is still as broken as in 1.12, it will not work when holding something in your offhand is not returning true, too. Which is most likely everything except when dual wielding wrenches.

Also the same should work with the network tool. Which might even be more complicated.

shartte:
Hm, you are correct in that I also believe it will only do it if both return true.
We might have to do this event-based then. :-|

PR #4664, opened 2020-08-30
Remove GZipStreams from packets.

yueh:
This should now be handled by the normal network manager (hopefully).
Also renamed the `CompressedNBTPacket` to `MEInterfaceUpdatePacket` as that is the only use case.

### DIFF
--- a/src/main/java/appeng/core/sync/packets/MEFluidInventoryUpdatePacket.java
+++ b/src/main/java/appeng/core/sync/packets/MEFluidInventoryUpdatePacket.java
@@ -19,13 +19,9 @@
 package appeng.core.sync.packets;
 
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
 import java.nio.BufferOverflowException;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.zip.GZIPInputStream;
-import java.util.zip.GZIPOutputStream;
 
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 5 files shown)


src/main/java/appeng/core/sync/packets/MEFluidInventoryUpdatePacket.java
```diff
@@ -66,47 +59,18 @@
...
-                }
-            }
-
-            while (uncompressed.readableBytes() > 0) {
-                this.list.add(AEFluidStack.fromPacket(uncompressed));
-            }
-        } catch (IOException e) {
-            throw new RuntimeException("Failed to decompress packet.", e);
+        while (stream.readableBytes() > 0) {
+            this.list.add(AEFluidStack.fromPacket(stream));
```
shartte, line 72: I'd probably feel safer if we instead wrote out the count explicitly. I'll take another look at this packet

yueh: I do not think that is currently possible. We use the `BufferOverflowException` to split it across multiple packets. Once the first one is full, it is send off and another one is created. Or at least not as easy as it should be.

Like we would have to inject more data somewhere in the middle of the `PacketBuffer` or at least stitch a completely new one together. Like serialise the item, but only count the bytes, keep the `IAEItemStack` in a list and then create the final packet from it when calling `toPacket()`

shartte: Yeah that comment was more tentative than definitive. I have to check this in actual code (which I cant access right now).
As long as we have an unsent packetbuf, we can write integers at arbitrary positions though (just not varints)

PR #4670, opened 2020-08-31
Guava-Cache for Cable Bus Quads

shartte:
Implemented a cache with maximum quad-count it will cache for the cable bus,

which also fixes concurrent access to the hashmap.

### DIFF
--- a/src/main/java/appeng/client/render/cablebus/CableBusModelLoader.java
+++ b/src/main/java/appeng/client/render/cablebus/CableBusModelLoader.java
@@ -18,7 +18,6 @@ public class CableBusModelLoader implements IModelLoader<CableBusModel> {
 
     @Override
     public void onResourceManagerReload(IResourceManager resourceManager) {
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/client/render/cablebus/CableBusModelLoader.java
```diff
@@ -18,7 +18,6 @@ public CableBusModelLoader(PartModels partModels) {
 
     @Override
     public void onResourceManagerReload(IResourceManager resourceManager) {
-        CableBusBakedModel.clearCache();
```
yueh, line 21: Does should still be  required. Reloading the resources could cause the model geometry to change. Without clearing the cache, it can still return old models.

shartte: No because the model object itself is re-created on resource reload, and the cache is no longer static.

yueh (approved):

PR #4675, opened 2020-09-01
Fixes #4669: Take NBT into account for comparing.

### LINKED ISSUE #4669: Items of the same type but with different enchantments combine in me system
**Describe the bug**
When adding two or more of the same item with different enchantments (e.g. a diamond pickaxe with fortune I and another with unbreaking III) into an ME system, the second item stacks into the first, causing loss of the second item.

**To Reproduce**
1. Create any ME system
2. Get two of the same enchantable item
3. Ensure that at least one of the items is enchanted and that both have identical durability
4. Put both items into the me system via any method
5. Both items will occupy one slot and have the same enchantments

**Expected behavior**
Items with differing enchantments should not stack in an ME system

**Additional context**
Recording of bug: https://youtu.be/NP7CJjZ30lg
Screenshots of bug: https://imgur.com/a/0xC1utN
Issue also occurs on latest 1.16.1 build

**Environment**

Minecraft Versions: 1.16.1 and 1.16.2
Applied Energistics 2 versions: appliedenergistics2-8.0.0-beta.2 and appliedenergistics2-8.1.0-alpha.1
Forge versions: 32.0.108 and 33.0.23

Tested with only AE2 + forge on 1.16.2 and many mods + AE2 on 1.16.1 (mods list: https://imgur.com/a/yUdkoZn)

yueh:
This only covers combining different items together.
Fuzzy filtering is still a bit too fuzzy for corner cases.

### DIFF
--- a/src/main/java/appeng/util/item/AESharedItemStack.java
+++ b/src/main/java/appeng/util/item/AESharedItemStack.java
@@ -29,6 +29,7 @@ import net.minecraft.nbt.CompoundNBT;
 import appeng.api.config.FuzzyMode;
 
 final class AESharedItemStack implements Comparable<AESharedItemStack> {
+
     private final ItemStack itemStack;
     private final int itemId;
     private final int itemDamage;
@@ -97,7 +98,8 @@ final class AESharedItemStack implements Comparable<AESharedItemStack> {
             return damageValue;
         }
 
-        return 0;
+        return System.identityHashCode(this.getDefinition().getTag())
+                - System.identityHashCode(b.getDefinition().getTag());
     }
 
     private int makeHashCode() {
@@ -183,4 +185,5 @@ final class AESharedItemStack implements Comparable<AESharedItemStack> {
         }
 
     }
+
 }


src/main/java/appeng/util/item/AESharedItemStack.java
```diff
@@ -97,7 +98,7 @@ public int compareTo(final AESharedItemStack b) {
             return damageValue;
         }
 
-        return 0;
+        return this.hashCode - b.hashCode;
```
shartte, line 101: This is only intended as a stopgap I presume because hashCode is not _guaranteed_ to be unique, right?

yueh: Technically no, but that probably applies to anything with a finite number. At least until we reach something like 2^265 and can address every atom in the universe.

It still depends on the item id, which is unique and damage value, which is mostly unique for our use case. The chance of having two NBT objects with the same hashcode is more or less negligible. So in my opinion worth the risk.

shartte: You are assuming that this is some kind of cryptographic super-hash which it is not, why do we not use identity-hash-code here, which at least is supposedly serial?

yueh: Oh wait. I assumed it would use the identity hashcode here. I'll change that.

yueh: `System.identityHashCode(null)` will return `0`, so without a tag it will default to `0 - 0`

PR #4724, opened 2020-09-11
Add metrics using micrometer

yueh:


yueh:
Closing for alternative solution

Issue #4727, opened 2020-09-13
Allow Storage Bus configurations to be copied into a Cell Workbench


nBeebz:
**Describe the feature**
It should be possible to load a Storage Bus's partition config into a Cell Workbench using a Memory Card, and vice versa.

**Reasons why it should be considered**
A Storage Bus and a Cell Workbench have functionally identical configurations of 63 items, theoretically they should be compatible. Some interesting possibilities this would allow:
- Creating a partition for ME Drives and View Cells using any arbitrary container.
- Easily transferring the contents of a partitioned ME Drive via a Storage Bus using an I/O Port.
- Sets of items can be "abstracted away" to some external system by copying the Storage Bus's configuration into an inverted View Cell. 

I think this would give View Cells and partitions a lot more utility.

nBeebz:
Thinking about writing a PR for this @shartte, y/n?

Technici4n:
Implemented in #6638.

PR #4778, opened 2020-10-15
Allow memory cards to upload configurations to any valid machine

nBeebz:
https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/4727

Tested transferring configurations between a cell workbench, storage bus, and import bus with and without capacity cards.

(I tried to squash commit history but it only made them stronger, sorry)


yueh:
This is way to error prone in terms of accidentally applying a few setting to a completely different machine. E.g. applying the priority of an export bus to a drive by accidentally clicking it without any feedback about what has changed. Some for applying a config which needs capacity cards in one machine, but not in other and so potentially assigning inaccessible config slots.

There needs at least some sort of marking compatible machines like storage buses, formation planes and maybe the cell workbench into a single "storage machine" category. (Assuming it is actually possible and not cause new issues).
Basically anything where the config can be fully applied and not just partly.

PR #4811, opened 2020-11-04
Migrate to Forge BiomeLoadingEvent and make Worldgen Config more consistent

shartte:
Move to Forge's BiomeLoadingEvent and migrate WorldGenRegistry to biomes from dimensions (which didn't actually work anymore).

Also remove unused config for Meteorites dimension blacklist and add quartz and meteorite biome blacklists instead.


yueh:
The API change is technically a breaking change. Not sure if they even work currently, but at least keep the old methods with a NOOP implementation (assuming they don't even work correctly) and just deprecate it with a scheduled to be removed with 9.0?

shartte:
@yueh I wouldn't even say "technically". It's absolutly a breaking change. So I picked up your idea of adding stub methods and did that. (And no, neither the methods nor the config setting did anything, we didn't check the dimension anywhere in worldgen anymore).

yueh (approved):

PR #4837, opened 2020-11-23
Add CraftTweaker support

jaredlll08:
I figured since you guys used to have CraftTweaker support in 1.12 I would just PR it into 1.16 as well.

If you don't want the support anymore, just let me know and close this issue, and I'll just put it into a modtweaker-like mod or something like that.

Also if you ever add new recipes / change how current recipes work, just give me a ping and I'll update the support.

I put CraftTweaker as `compileOnly` because I doubt you actually want to load it every time you load the game when developing.

This is the script I used to test that it all worked, removals are all handled in the back-end, so you only need to worry about adding recipes.

```zenscript
import crafttweaker.api.item.IIngredient;

<recipetype:appliedenergistics2:inscriber>.removeRecipe(<item:appliedenergistics2:logic_processor>);
<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_inscribe_test", <item:minecraft:diamond>, <item:minecraft:dirt>);
<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_inscribe_test_top_input", <item:minecraft:dirt>, <item:minecraft:glass>, [<item:minecraft:dirt>] as IIngredient[]);
<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_inscribe_test_bottom_input", <item:minecraft:glass>, <item:minecraft:stick>, [<item:minecraft:air>, <item:minecraft:stick>] as IIngredient[]);
<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_inscribe_test_all_inputs", <item:minecraft:redstone>, <item:minecraft:apple>, [<item:minecraft:arrow>, <item:minecraft:stick>] as IIngredient[]);

<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_press_test", <item:minecraft:scute>, <item:minecraft:lapis_lazuli>);
<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_press_test_top_input", <item:minecraft:coal>, <item:minecraft:white_dye>, [<item:minecraft:white_wool>] as IIngredient[]);
<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_press_test_bottom_input", <item:minecraft:iron_ingot>, <item:minecraft:orange_dye>, [<item:minecraft:air>, <item:minecraft:orange_wool>] as IIngredient[]);
<recipetype:appliedenergistics2:inscriber>.addInscribeRecipe("inscriber_press_test_all_inputs", <item:minecraft:gold_ingot>, <item:minecraft:magenta_dye>, [<item:minecraft:magenta_wool>, <item:minecraft:light_blue_wool>] as IIngredient[]);

<recipetype:appliedenergistics2:grinder>.removeRecipe(<item:minecraft:flint>);
<recipetype:appliedenergistics2:grinder>.addRecipe("grinder_test", <item:minecraft:apple>, <item:minecraft:arrow> * 4, 5);
<recipetype:appliedenergistics2:grinder>.addRecipe("grinder_test_optional_outputs", <item:minecraft:apple>, <item:minecraft:arrow> * 4, 5, [<item:minecraft:diamond> % 50]);
<recipetype:appliedenergistics2:grinder>.addRecipe("grinder_test_with_ingredient", <item:minecraft:redstone>, <item:minecraft:redstone_lamp>, 4, 5);
<recipetype:appliedenergistics2:grinder>.addRecipe("grinder_test_with_ingredient_and_optional_outputs", <item:minecraft:apple>, <tag:items:forge:dyes>, 4, 5, [<item:minecraft:glass> % 50]);
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,6 +61,7 @@ dependencies {
     // compile against provided APIs
     compileOnly "mezz.jei:jei-${jei_minecraft_version}:${jei_version}:api"
     compileOnly "mcjty.theoneprobe:TheOneProbe-${minecraft_release}:${minecraft_release}-${top_version}:api"
+    compileOnly fg.deobf("com.blamejared.crafttweaker:CraftTweaker-${minecraft_version}:${crt_version}")
 
     // Runtime, Mods
     runtimeOnly fg.deobf("mezz.jei:jei-${jei_minecraft_version}:${jei_version}")
--- a/gradle.properties
+++ b/gradle.properties
@@ -21,6 +21,7 @@ forge_version_range=[35.0.1,36.0.0)
 jei_minecraft_version=1.16.4
 jei_version=7.6.0.53
 top_version=3.0.4-beta-7
+crt_version=7.1.0.69
 
 #########################################################
 # Deployment                                            #
--- a/src/main/java/appeng/integration/modules/crafttweaker/GrinderRecipeManager.java
+++ b/src/main/java/appeng/integration/modules/crafttweaker/GrinderRecipeManager.java
@@ -1,0 +1,61 @@
+package appeng.integration.modules.crafttweaker;
+
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.blamejared.crafttweaker.api.CraftTweakerAPI;
+import com.blamejared.crafttweaker.api.annotations.ZenRegister;
+import com.blamejared.crafttweaker.api.item.IIngredient;
+import com.blamejared.crafttweaker.api.item.IItemStack;
+import com.blamejared.crafttweaker.api.managers.IRecipeManager;
+import com.blamejared.crafttweaker.impl.actions.recipes.ActionAddRecipe;
+import com.blamejared.crafttweaker.impl.item.MCWeightedItemStack;
+
+import org.openzen.zencode.java.ZenCodeType;
+
+import net.minecraft.item.crafting.IRecipeType;
+import net.minecraft.util.ResourceLocation;
+
+import appeng.api.features.InscriberProcessType;
+import appeng.recipes.handlers.GrinderOptionalResult;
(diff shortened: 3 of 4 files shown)


jaredlll08:
Not sure why the build failed, doesn't appear to have anything to do with my code, seems it can't get the game's assets?

yueh:
Thanks, but the support in 1.12 was just an interim solution as long as datapacks were not sufficient for our requirements. As this is no longer the case, we will only support these from now on. (assuming there won't be any breaking change in the future rendering them useless again)

I'd guess we should actually look into downgrading visibility for recipe constructors. Which would allow us to ensure that the serializers are the only place we have to do validation and handle cases like adding default values to old jsons in case we add more options and now need to fix them without breaking.

jaredlll08:
> I'd guess we should actually look into downgrading visibility for recipe constructors

I personally feel that is a very bad take, as it doesn't even let me move the support to a modtweaker like mod.

At the end of the day, it is your mod, so do what you wish, but I didn't just make this PR randomly, people have been asking for support because modpack makers don't really like datapacks as they are annoying to work with.

yueh:
Recipes are explicitly no longer part of the API with jsons being the only source of truth. How these are then produced is no longer important, e.g. crafttweaker could just generate a datapack instead of directly modifying recipes. As far as I can remember kubejs can for example work with arbitrary json, so no need to explicitly support each and every mod out there. And it can be used for forge and fabric, so it will not force modpack makers to learn a completely different tool for each one.

Also some parts are still tentatively using the recipe manger directly, which in the past was a really bad idea in terms of performance. If this would still be the case, I still want to have the option to replace it with a lookup cache. Ideally with an immutable one I generate once during the server start as well as flush and recreate it on a resource reload. Which can easily interfere with another mod mutating the recipes at the same time outside the normal minecraft/forge approach.

PR #4899, opened 2021-01-01
Fixes #4893: Fix autocrafting when substitutions are used

### LINKED ISSUE #4893: Server crash when trying to autocraft pattern with substitutions
**Describe the bug**

I'm helping to test the next release of the ATM6 pack. We've updated AE2.

While trying to craft some Quartz Glass via pattern with substitutions enabled, the server crashed as soon as I submitted the crafting job. The world would not come back up and I had to restore a backup. See crash log below.

Could of course be a bad mod interaction, but I wanted to bounce the stack trace off you before I start removing 200+ mods one by one.

Just to be clear, this was an AE2 crafting recipe that was to be carried out in a molecular assembler. No actual cross-mod interaction via processing pattern or similar. The stack trace seems to point at an explosion when trying to resolve ingredients via tags.

**To Reproduce**

- Create pattern with substitutions
- Put pattern in terminal
- Submit a crafting job with said pattern

**Expected behavior**

- For the job to either submit or fail to submit

**Additional context**

Crashlog: https://hastebin.com/urezegefaj

**Environment**

- Mod pack: ATM6 1.3.8 (unreleased)

shartte:
I did some superficial testing of this with the cable anchor and crafting table recipes, and both cases seemed to work. But I'd say this needs more testing. However, it should at least not crash anymore. When substitutions break down, it should simply stall the crafting job now.

Fix retrieval of substitutions from a crafting pattern. Map indices from our sparse inputs to the probable indices of the crafting pattern and more gracefully handle mismatches.

When checking whether the inventory of a crafting CPU is sufficient to craft an autocrafting job, account for the new substitutions.

Fixes #4893

### DIFF
--- a/src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
+++ b/src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
@@ -443,38 +443,86 @@ public final class CraftingCPUCluster implements IAECluster, ICraftingCPU {
         return null;
     }
 
-    private boolean canCraft(final ICraftingPatternDetails details, final Collection<IAEItemStack> condensedInputs) {
-        for (IAEItemStack g : condensedInputs) {
+    private boolean canCraft(final ICraftingPatternDetails details) {
 
-            if (details.isCraftable()) {
+        if (!details.isCraftable()) {
+            // Processing patterns are relatively easy
+            for (IAEItemStack input : details.getInputs()) {
+                final IAEItemStack ais = this.inventory.extractItems(input.copy(), Actionable.SIMULATE,
+                        this.machineSrc);
+                final ItemStack is = ais == null ? ItemStack.EMPTY : ais.createItemStack();
+
+                if (is.isEmpty() || is.getCount() < input.getStackSize()) {
+                    return false;
+                }
+            }
+        } else if (details.canSubstitute()) {
+
+            // When substitutions are allowed, we have to keep track of which items we've reserved
+            IAEItemStack[] sparseInputs = details.getSparseInputs();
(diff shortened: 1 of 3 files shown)


Meowing-Cat:
When can we expect a live build? on curse etc?

shartte:
@Meowing-Cat We don't give out dates.

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -443,38 +443,86 @@ private IGrid getGrid() {
...
+    private boolean canCraft(final ICraftingPatternDetails details) {
+
+        if (!details.isCraftable()) {
+            // Processing patterns are relatively easy
+            for (IAEItemStack input : details.getInputs()) {
+                final IAEItemStack ais = this.inventory.extractItems(input.copy(), Actionable.SIMULATE,
+                        this.machineSrc);
+                final ItemStack is = ais == null ? ItemStack.EMPTY : ais.createItemStack();
+
+                if (is.isEmpty() || is.getCount() < input.getStackSize()) {
```
yueh, line 455: This is something we should test if it's needed. Calling `createItemStack()` is a bit expensive and the condition does not make use of any `ItemStack` relevant functions. So it could just be replaced by ` if (ais == null || ais.getStackSize() < input.getStackSize())`. Just needs to potentially take into account that `ItemStack` is limited to int, while `IAEStack` uses a long

shartte: Changed

yueh (approved):

PR #4905, opened 2021-01-04
Fixes #4890 : Kill CompassService when server is closed, and not when closing.

### LINKED ISSUE #4890: Error on shutting down server
Shutdown of server. We use quantum bridges in our world where we have a network dedicated for p2p and a network for storage. All chunkloaded.

**To Reproduce**
Shut down our server.

**Expected behavior**
A clean shutdown and not being in a state where it's not running anymore but also not still running in the console.

**Environment**
Log from the stop command till where it freezes in the console. No crashlog generated.
https://pastebin.com/Lh3fApC9

For a mod list you can look into this pastebin https://pastebin.com/W1T8eUBp
This is an unrelated pastebin however to this issue.

shroomdog27:
This is done in order to ensure that specified TEs, specifically CraftingTileEntity do not submit requests to the CompassService after it has already been killed.

If the better way to fix this issue is to instead make sure that no TEs submit requests after the CompassService has closed, then I am unable to help with that.

### DIFF
--- a/src/main/java/appeng/core/worlddata/WorldData.java
+++ b/src/main/java/appeng/core/worlddata/WorldData.java
@@ -110,12 +110,13 @@ public final class WorldData implements IWorldData {
 
     @Override
     public void onServerStopping() {
-        compassData.service().kill();
+
     }
 
     @Override
     public void onServerStoppped() {
         Preconditions.checkNotNull(server);
+        compassData.service().kill();
         instance = null;
         WorldData.server = null;
     }


yueh:
Currently this should work, however with a changed behaviour it can easily introduce data loss.

The `FMLServerStoppedEvent` is trigger exactly before the process itself is stopped and not the server itself. So at this point anything IO related is already closed and we cannot save any data any longer.

Also in terms of `FMLServerStoppingEvent`, which is called before the final tick of each world. So there is also a chance to loss data in case we have to revert the change and further save it with something else than `WorldSavedData`.

PR #4934, opened 2021-01-11
Extract common code in ItemP2PTunnelPart event handler-methods

AzazeII:
I've just noticed that all four event methods have common code, so i extracted it in distinct method, so it's more easy to maintain code


AzazeII:
Wait, i missed out that style uses spaces instead of tabs

AzazeII:
CI/CD check looks nice now :>

yueh:
Thanks, but the item p2p still needs some rework and this might not even be kept around.

Issue #4936, opened 2021-01-11
Memory card support for configured fluids and level emitters reportingValue


PrototypeTrousers:
Memory cards completely ignore fluids configured and do not save the configured level on emitters (item and fluids)

**Describe the feature**
Copying settings with the memory card saying its configuration has been copied, should actually be copied.

**Reasons why it should be considered**
This should make even easier to configure AE networks, and manage your setups

yueh:
Introduced by the merged PR.

PR #4937, opened 2021-01-11
Memory card support for fluids and level emitters settings

PrototypeTrousers:
Suggested fix for #4936
Allow memory cards to copy configured fluid slots
Copy level emitters reporting value


src/api/java/appeng/api/implementations/tiles/IConfigurableFluidInventory.java
```diff
@@ -0,0 +1,10 @@
+package appeng.api.implementations.tiles;
+
+import net.minecraftforge.fluids.capability.IFluidHandler;
+
+public interface IConfigurableFluidInventory {
```
yueh, line 5: This should not be public API interface currently. There are still things which can be handled better in general and I would like to avoid having to do an API break just for that. Especially as it then still relies on an internal interface to actually work.

yueh (changes requested):

src/api/java/appeng/api/implementations/tiles/IConfigurableFluidInventory.java
shartte: I moved it to appeng.fluids.helper.IConfigurableFluidInventory

shartte:
@PrototypeTrousers Thanks for your contribution!

PR #4950, opened 2021-01-16
Delay onReady for AE tile entities until they would normally tick

### LINKED ISSUE #4940: On a server after a while import and export busses stop working
**Describe the bug**

I play on a server and after a while (around half a Day) some (not all) import and export busses stop working.
If you insert an acceleration card it starts working again.

**To Reproduce**

Run an AE system on a server for a while

**Expected behavior**

The busses working normaly

**Additional context**

**Environment**

- Environment: 
FTB DireWolf20 1.16 v. 1.31
AE 8.2.0 Beta 2
Force crash client:
[crash-2021-01-12_14.57.36-client.txt](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/5802526/crash-2021-01-12_14.57.36-client.txt)

shartte:
Fixes #4940: Fix AE tile entities starting up too early - before the chunks are fully loaded and other tile entities begin ticking.

Also use the same method that vanilla uses to determine whether tile entities in a chunk should tick or not, instead of checking whether entities would tick.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -150,6 +150,8 @@ minecraft {
     runs {
         client {
             property 'forge.logging.console.level', 'debug'
+            // See https://github.com/Vazkii/Patchouli#mixin-troubleshooting
+            property 'mixin.env.disableRefMap', 'true'
             workingDirectory project.file('run')
             property "mixin.debug.export", "true"
             mods {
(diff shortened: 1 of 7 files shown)


Pingger:
My guess is that this will fix quite a lot of issues, that couldn't easily be reproduced. This could even fix quite a lot of spongeforge issues.

yueh (approved):

PR #4998, opened 2021-02-18
Adds a network wide chunkloader

### LINKED ISSUE #4966: Add a way to chunkload ME networks
# Goal

As we expect each ME network is fully chunkloaded for various reasons like consistency, performance and similar, we should provide a native way to achieve this without relying on 3rd party mods.

## Add a way to easily track which chunks/worlds a network span (prerequisite)

Currently there is no fast/easy way to determine which chunks a ME network does occupy. It would have to iterate over every grid node and their respective machine and use this to calculate the area each time a network changes.

A simple solution might be to introduce a new `IGridCache` to track the chunks/worlds just when a grid node is added or removed from a network. It could also track various other statistics for a later use like metrics.

## Add the chunk loader itself

Each chunkloader should be pretty straightforward and not force the player to deal with technical details like chunk borders and potentially missing a single one and causing various issues.

To provide a better QoL, each network should just require a single chunk loader similar to a security station and then simply load every chunk the network occupies without forcing a player to deal with the range at all.

In case it is also used for other mods, it might be an idea to calculate a continuous area. e.g. when a network occupies the outer ring of a 3x3 chunk area, but not the middle one. In this case adding that chunk might be a valid option. As well as add a 1 chunk outline as additional buffer around the occupied area. But that could easily load too many chunks.

## Balancing

As AE2 does not provide any regenerative resource beside some edge cases like matter balls or singularites, they should not really require some items or fluids as fuel.

An energy requirement like `x^numberOfChunks ae/T` on the other hand would be certainly be an option to strongly consider.

## Optional features and improvements

As it might be a bit tedious to add a chunk loader to every single subnetwork, it might be an option to provide an improved quartz fiber which extends the covered area as well as potentially also providing a proxy for the security station.
However as grids to some degree are directional and it depends on which side the quartz fiber between two networks is located, it behaves differently and might make this feature not possible or be inconsistent in its behaviour.

## Impact on servers

As this is would be a new block, servers could simply disable the recipe for them, if they don't want to offer this feature.

yueh:
This makes use of the readded `ForgeChunkManager` making it much more easy to restore the chunk loaders during load as well as also easily handling ones removed by external means. (Backups, nbt editing, etc).
**However this has the limitation that we can only load their own world/dimension.** Networks spanning multipe worlds via QNBs or other means need their own chunk loader. Otherwise we would have to implement handling the chunk tickets completely on your side.

Adds a new `MENetworkStatisticsEvent` with subtypes related to the grid entering or leaving new chunks. This could be further extended in the future for other use cases like debugging. 
**For now the new `IGridCache` providing these informations is not part of the public API and should not be used until changed**

It adds a new block, the corresponding tile entity, container and screen for it.
The UI is not fully polished yet and needs some work, but provides details about the energy consumption of this chunk loader, how many chunks it loads as well as how many chunks and worlds the whole network spans as well as how many of them are loaded. **This only takes our chunk loaders into account**

Power consumption per tick is not finalised and needs some evaluation. The basic idea is to have it increasingly scale with the amount of chunks and worlds. Probably with a minimum value aimed towards midgame, but certainly not endgame. This should be covered by the scaling.

Resolves #4966


src/api/java/appeng/api/networking/events/statistics/MENetworkChunkEvent.java
```diff
@@ -0,0 +1,63 @@
...
+ * IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+ * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+package appeng.api.networking.events.statistics;
+
+import net.minecraft.util.math.ChunkPos;
+import net.minecraft.world.server.ServerWorld;
+
+public class MENetworkChunkEvent extends MENetworkStatisticsEvent {
```
shartte, line 29: this needs a class doc, also: is this ever instantiated directly? if not, make it abstract

comment the subclasses too

src/main/java/appeng/block/spatial/SpatialAnchorBlock.java
```diff
@@ -0,0 +1,77 @@
...
+import net.minecraft.world.World;
+
+import appeng.block.AEBaseTileBlock;
+import appeng.container.ContainerLocator;
+import appeng.container.ContainerOpener;
+import appeng.container.implementations.SpatialAnchorContainer;
+import appeng.tile.spatial.SpatialAnchorTileEntity;
+import appeng.util.Platform;
+
+public class SpatialAnchorBlock extends AEBaseTileBlock<SpatialAnchorTileEntity> {
```
shartte, line 40: comment :-P

src/main/java/appeng/block/spatial/SpatialAnchorBlock.java
```diff
@@ -0,0 +1,77 @@
...
+                ContainerOpener.openContainer(SpatialAnchorContainer.TYPE, p,
+                        ContainerLocator.forTileEntitySide(tg, hit.getFace()));
+            }
+            return ActionResultType.SUCCESS;
+        }
+        return ActionResultType.PASS;
+    }
+
+    @Override
+    public void onBlockHarvested(World worldIn, BlockPos pos, BlockState state, PlayerEntity player) {
```
shartte, line 65: Why is this here? Isn't this handled by the standard TE removal?

src/main/java/appeng/client/gui/implementations/SpatialAnchorScreen.java
```diff
@@ -0,0 +1,257 @@
...
+            }
+
+            x++;
+
+            if (x >= SCROLL_COLUMNS) {
+                y++;
+                x = 0;
+            }
+        }
+
```
shartte, line 120: Minecraft UIs are such an ugly mess. Blaaaaaaaaaargh

src/main/java/appeng/client/gui/implementations/SpatialAnchorScreen.java
```diff
@@ -0,0 +1,257 @@
...
+        final int viewEnd = viewStart + SCROLL_COLUMNS * SCROLL_ROWS;
+
+        String toolTip = "";
+        int toolPosX = 0;
+        int toolPosY = 0;
+
+        for (int z = viewStart; z < Math.min(viewEnd, this.repo.size()); z++) {
+            final IAEItemStack refStack = this.repo.getReferenceItem(z);
+            if (refStack != null) {
+                ChunkPos chunkPos = new ChunkPos(refStack.getDefinition().getOrCreateTag().getLong("chunk"));
```
shartte, line 167: I don't remember. Did `getDefinition()` always create a new ItemStack?

src/main/java/appeng/client/render/overlay/IOverlayDataSource.java
```diff
@@ -0,0 +1,38 @@
...
+
+public interface IOverlayDataSource {
+
+    Set<ChunkPos> getOverlayChunks();
+
+    TileEntity getOverlayTileEntity();
+
+    DimensionalCoord getOverlaySourceLocation();
+
+    int getOverlayColor();
```
shartte, line 36: Comments 👀

src/main/java/appeng/client/render/overlay/OverlayRenderType.java
```diff
@@ -0,0 +1,86 @@
...
+
+    public OverlayRenderType(String p_i225992_1_, VertexFormat p_i225992_2_, int p_i225992_3_, int p_i225992_4_,
+            boolean p_i225992_5_, boolean p_i225992_6_, Runnable p_i225992_7_, Runnable p_i225992_8_) {
+        super(p_i225992_1_, p_i225992_2_, p_i225992_3_, p_i225992_4_, p_i225992_5_, p_i225992_6_, p_i225992_7_,
+                p_i225992_8_);
+    }
+
+    public static RenderType getBlockHilightFace() {
+        return makeType("block_hilight",
+                DefaultVertexFormats.POSITION_COLOR, GL11.GL_QUADS, 256,
```
shartte, line 45: Please check what the 256 was for... Was this the max size of the buffer or was this just the size until it flushes?
Might hit this limit when showing very large area

(same below)

src/main/java/appeng/container/implementations/SpatialAnchorContainer.java
```diff
@@ -0,0 +1,186 @@
...
+
+public class SpatialAnchorContainer extends AEBaseContainer {
+
+    public static ContainerType<SpatialAnchorContainer> TYPE;
+
+    private static final ContainerHelper<SpatialAnchorContainer, SpatialAnchorTileEntity> helper = new ContainerHelper<>(
+            SpatialAnchorContainer::new, SpatialAnchorTileEntity.class, SecurityPermissions.BUILD);
+
+    private IGrid network;
+    private int delay = 40;
```
shartte, line 67: Why does this start at 40 and below it checks for 15?
Shouldn't the expected delay be a constant?

src/main/java/appeng/me/cache/StatisticsCache.java
```diff
@@ -0,0 +1,170 @@
...
+
+import appeng.api.networking.IGrid;
+import appeng.api.networking.IGridCache;
+import appeng.api.networking.IGridHost;
+import appeng.api.networking.IGridNode;
+import appeng.api.networking.IGridStorage;
+import appeng.api.networking.events.statistics.MENetworkChunkEvent;
+import appeng.api.util.DimensionalCoord;
+
+public class StatisticsCache implements IGridCache {
```
shartte, line 41: comment.....

src/main/java/appeng/tile/spatial/SpatialAnchorTileEntity.java
```diff
@@ -0,0 +1,375 @@
...
+    public void read(BlockState blockState, CompoundNBT data) {
+        super.read(blockState, data);
+        this.manager.readFromNBT(data);
+    }
+
+    @Override
+    protected void writeToStream(PacketBuffer data) throws IOException {
+        super.writeToStream(data);
+        data.writeBoolean(displayOverlay);
+        if (this.displayOverlay) {
```
shartte, line 99: oh god. are you not sending this in a separate packet? 👀

src/main/java/appeng/tile/spatial/SpatialAnchorTileEntity.java
```diff
@@ -0,0 +1,375 @@
...
+
+        // Cleanup old data and remove it from the overlay manager as safeguard
+        this.chunks.clear();
+        OverlayManager.getInstance().removeHandlers(this);
+
+        if (this.displayOverlay) {
+            this.chunks.addAll(Arrays.stream(data.readLongArray(null)).boxed().map(c -> new ChunkPos(c))
+                    .collect(Collectors.toSet()));
+            // Register it again to render the overlay
+            OverlayManager.getInstance().showArea(this);
```
shartte, line 119: I am not the biggest fan doing this from readStream, but oh well, if it works...

shartte: p.s.: how do you decide to show the overlay when the chunkloader is not in view? 
and: how do you remove the overlay when it goes out of reach

src/main/java/appeng/block/spatial/SpatialAnchorBlock.java
yueh: argh. stupid naming with `remove`. Will change that.

src/main/java/appeng/client/gui/implementations/SpatialAnchorScreen.java
yueh: Yeah. And all these magic values, which are somehow related to eachother. Or not.

src/main/java/appeng/client/gui/implementations/SpatialAnchorScreen.java
yueh: That won't create a new one. But I guess a simple `getTag()` would be better to avoid corrupting the shared registry. The whole thing is just a massive hack to display a list of chunks. With the world overlay now I would even say we should remove it.

src/main/java/appeng/client/render/overlay/OverlayRenderType.java
yueh: No idea here. That particular source is also broken, so I cannot follow it at all.

src/main/java/appeng/container/implementations/SpatialAnchorContainer.java
yueh: Copy&Paste from another container. Probably just to ensure an immediate update after creating it.

src/main/java/appeng/tile/spatial/SpatialAnchorTileEntity.java
yueh: I did not want to touch the packets. And I do not think we have a single te sending it's own packets. So I really do not want to open that can of worms.

src/main/java/appeng/tile/spatial/SpatialAnchorTileEntity.java
yueh: It's just an entry into a map. But I am also not real fan of. However there is no ticking for our tiles client side, so it has to react to something.

yueh: In terms of distance, not at all. Minecraft itself stops rendering it after like 600 blocks and the source itself is there to not unload itself. While it is outside the normal chunk render distance, it is still easily visible and could still act as guide for not that far away sections. And with the way the anchor works, it also only reports the chunk position of the current world.

src/main/java/appeng/client/gui/implementations/SpatialAnchorScreen.java
yueh: This code is now completely removed again.

src/main/java/appeng/container/implementations/SpatialAnchorContainer.java
yueh: Added a constant to initialize it and change the amount of skipped ticks.

PR #4999, opened 2021-02-18
Fixes #4996: Fix breaking change introduced by MultipartModelData

### LINKED ISSUE #4996: Forge 1.16.5-36.0.34 broke all rotatable models using multipart models
This currently affects the ME Drive and Wireless Access Point, which visual rotation no longer matches the pyhsical one.

The controller is the only other block currently using multiparts and does not appear to affected. Most likely because it cannot be rotated.

yueh:


shartte:
What is the actual breaking change? Did they make the model data final?

yueh:
Any model using multiparts as in the vanilla ones for fences or similar are now wrapping their `IModelData` into a `MultipartModelData`. So calling `IForgeTileEntity#getModelData()` will no longer return our `AEModelData`, but instead the mentioned `MultipartModelData` decorating our instance (at least within the rendering path and the changes it applies, calling it directly still gives a `AEModelData`). Therefore all `instanceof` checks now fail and there is also no way to extract our own data without reflections or adding another mixin.

I guess I could only add the property for the `AEModelData` and not everything else and then just cast it again would have been an option. But just going with it seems to be the more future-proof approach. E.g. once they realise the change broke any rendering using a custom `IBakedModel` and `IModelData` in combination with a multipart model and then change everything again.

The `MultipartModelData` also appears to call `AEModelData#getData()` directly without checking `AEModelData#hasProperty()`, thus crashing with the `IllegalStateException` we throw in this case. I guess I should also fix the throw left after the conditions.

shartte (approved):

PR #5018, opened 2021-02-28
Fixes #5017: Skip border chunks until ready.

### LINKED ISSUE #5017: TileEntities in border chunks will never be initialized
Any `AEBaseTileEntity` within a border chunk will not be initialized as minecraft does consider the chunk as existing even when the chunk is actually loaded.

As we drop these tile entities from the queue and the chunk can swap from non existing to existing without it being loaded again, these will never be initialized.

yueh:
Refactors the `TickHandler` and now uses a distinct queue for each
chunk. This allows us to skip chunks which are not ready yet.

### DIFF
--- a/src/main/java/appeng/hooks/TickHandler.java
+++ b/src/main/java/appeng/hooks/TickHandler.java
@@ -21,11 +21,13 @@ package appeng.hooks;
 import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Queue;
 import java.util.Set;
 import java.util.WeakHashMap;
@@ -36,6 +38,7 @@ import com.google.common.base.Stopwatch;
 import com.google.common.collect.LinkedListMultimap;
 import com.google.common.collect.Multimap;
 
+import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.ChunkPos;
 import net.minecraft.world.IWorld;
 import net.minecraft.world.World;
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/hooks/TickHandler.java
```diff
@@ -135,16 +139,17 @@ public void addCallable(final IWorld w, final IWorldCallable<?> c) {
...
-            this.getRepo().tiles.add(tile);
+        if (!tile.getWorld().isRemote()) {
+            this.getRepo().addTile(tile);
         }
     }
 
     private HandlerRep getRepo() {
-        if (Platform.isServer()) {
-            return this.server;
+        if (Platform.isClient()) {
```
shartte, line 148: Is every call to getRepo now guarded with a check for the world? If so, I think we can drop this check.

src/main/java/appeng/hooks/TickHandler.java
```diff
@@ -286,6 +250,73 @@ public void registerCraftingSimulation(final World world, final CraftingJob craf
...
+     * Ready the tiles in this world
+     * 
+     * @param world
+     */
+    private void readyTiles(IWorld world) {
+        final HandlerRep repo = this.getRepo();
+        final Map<ChunkPos, Queue<AEBaseTileEntity>> worldQueue = repo.tiles.getOrDefault(world,
+                Collections.emptyMap());
+
+        for (Iterator<Entry<ChunkPos, Queue<AEBaseTileEntity>>> it = worldQueue.entrySet().iterator(); it
```
shartte, line 289: Not a big fan of this for-loop construct, but eh... Okay

src/main/java/appeng/hooks/TickHandler.java
yueh: Probably, but I would keep it as sanity check in case some addon hooks into it. Better have it crash then whatever else happens.

src/main/java/appeng/hooks/TickHandler.java
yueh: Yeah. looks ugly but it allows us to use `iterator.remove()` and not run into CMEs or having to maintain another collection to clean it up later. I do not want to keep the empty map entry around when not necessary. And should something constantly swap chunks in and out that additional object instantiation will not really affect anything when compared to whatever else will already happen.

Just in case that question comes up at some point. At least document the thought process here.

shartte (approved):

PR #5020, opened 2021-03-01
update ru_ru.json for 8.3.0+ update

Gregorius-MC:


yueh:
Thanks. 

We might still at more translation keys until a beta release. So it might be an idea to wait for it, if you do not want to update it constantly.

Gregorius-MC:
Well, in that case, I will take my time and release the next translation update when the stable version of the mod is released.

Gregorius-MC:
> 
> 
> Thanks.
> 
> We might still at more translation keys until a beta release. So it might be an idea to wait for it, if you do not want to update it constantly.

Well, in that case, I will take my time and release the next translation update when the stable version of the mod is released.

yueh:
Maybe not for the stable one, as there probably won't be a patch release just to update translations. Beta should be fine and I'll try to avoid adding new keys when possible.

PR #5032, opened 2021-03-03
Don't drop parts and facades in creative mode

### LINKED ISSUE #5016: All cables drop in Creative mode
**Describe the bug**
Cables drop their respective items when broken by a player in Creative mode, which is inconsistent with vanilla behaviour, in which things broken by Creative mode players do not drop their items. This effectively makes testing designs using cables in Creative mode to be a bit annoying due to the item drops that can pile up in the player's inventory or around them.

**To Reproduce**
- Start Minecraft with the Fabric loader and AE2 in the mods folder.
- Create a new world in creative mode.
- Place down any cable, then break it. It drops itself.

**Expected behavior**
Cables would not drop their items when broken in Creative mode.

**Environment**
Minecraft version: 1.16.5
Fabric Loader version: 0.11.1
AE2 version: 8.2.0 alpha 1

Technici4n:
Part contents are dropped by the BE, parts and facades are dropped in `getDroppedStacks`.
I tested that explosions still drop everything.
Closes #5016.

### DIFF
--- a/src/main/java/appeng/block/networking/CableBusBlock.java
+++ b/src/main/java/appeng/block/networking/CableBusBlock.java
@@ -18,6 +18,8 @@
 
 package appeng.block.networking;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Random;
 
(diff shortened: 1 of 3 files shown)


src/main/java/appeng/block/networking/CableBusBlock.java
```diff
@@ -158,19 +164,22 @@ public boolean canReplace(BlockState state, ItemPlacementContext context) {
...
+        LootContext lootContext = builder.parameter(LootContextParameters.BLOCK_STATE, state)
+                .build(LootContextTypes.BLOCK);
+        BlockEntity be = lootContext.get(LootContextParameters.BLOCK_ENTITY);
+        if (be instanceof CableBusBlockEntity) {
+            CableBusBlockEntity bus = (CableBusBlockEntity) be;
+            List<ItemStack> drops = new ArrayList<>();
+            bus.getCableBus().appendPartStacks(drops);
+            return drops;
+        } else {
+            AELog.warn("The block entity was either null or of the wrong type! Skipped cable bus drops!");
```
shartte, line 179:
```suggestion
            AELog.debug("The block entity was either null or of the wrong type! Skipped cable bus drops!");
```

shartte (approved):

Issue #4867, opened 2020-12-12
Modern Industrialization pipe fell into hopper attached to Inscriber, NPE on world load


WilloftheWay:
I was breaking and replacing some piping in my base in the most up to date version of All of Fabric 3. One of the pipes fell into the hopper that is attached to one of my inscribers, crashed with NPE. Now gives same on world load. Pastebin https://pastebin.com/EH6c1bcD

Technici4n:
Duplicate of #4840. I think I outlined a potential fix there.

Technici4n:
Should be fixed in 0.8.2-alpha.3.

PR #5034, opened 2021-03-03
Prevent NPE in InscriberTESR when getQuads() returns null

### LINKED ISSUE #4840: Crash when MI pipe enters inscriber
**Describe the bug**
Rendering crash when a Modern Industrialization pipe enters an inscriber.

**To Reproduce**
Place inscriber on the ground, put hoppers inputting into it, then put an MI pipe in, it will crash if it can enter.

**Additional context**
Crash report:
https://pastebin.com/d303p1eP

**Environment**
Fabric version, Specifics in the crash log.

Technici4n:
Closes #4840 and #4867.

### DIFF
--- a/src/main/java/appeng/client/render/tesr/InscriberTESR.java
+++ b/src/main/java/appeng/client/render/tesr/InscriberTESR.java
@@ -1,5 +1,6 @@
 package appeng.client.render.tesr;
 
+import java.util.List;
 import java.util.Random;
 
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/client/render/tesr/InscriberTESR.java
```diff
@@ -203,7 +205,9 @@ private void renderItem(MatrixStack ms, final ItemStack stack, final float o, Ve
             // for direction=null, while a block-model will have their faces for
             // cull-faces, but not direction=null
             BakedModel model = itemRenderer.getModels().getModel(stack);
-            if (!model.getQuads(null, null, new Random()).isEmpty()) {
+            List<BakedQuad> quads = model.getQuads(null, null, new Random());
+            // Note: quads may be null for mods implementing FabricBakedModel without caring about getQuads.
+            if (quads != null && quads.isEmpty()) {
```
shartte, line 210:
```suggestion
            if (quads != null && !quads.isEmpty()) {
```

Shouldn't it be !quads.isEmpty() ?

PR #5036, opened 2021-03-03
Set molecular assembler output max stack size to 64

Technici4n:
Otherwise the output gets voided if the resulting stack size would be > 1.

Place a log into a powered molecular assembler with a planks crafting pattern and the output will get voided. Same if there are > 1 leftover items after extraction into adjacent inventories.


Technici4n:
Oops 😄

PR #5037, opened 2021-03-04
Update ru_ru.json

Romz24:


src/main/resources/assets/appliedenergistics2/lang/ru_ru.json
```diff
@@ -171,14 +176,22 @@
   "gui.appliedenergistics2.Creates": "Делает",
   "gui.appliedenergistics2.Cyan": "Бирюзовый",
   "gui.appliedenergistics2.Drive": "МЭ накопитель",
-  "gui.appliedenergistics2.ETAFormat": "HH:mm:ss",
+  "gui.appliedenergistics2.ETAFormat": "ЧЧ:мм:сс",
```
artheus, line 179: I do not think that this will work properly, as the time format is probably expected to be in Latin letters.

artheus (commented):
Looks good, except for the time format.

src/main/resources/assets/appliedenergistics2/lang/ru_ru.json
Romz24: This is taken from [here](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/master/src/main/resources/assets/appliedenergistics2/lang/ru_ru.json#L180), and it works as expected there.

PR #5042, opened 2021-03-06
Draft: Fabric 21w08b

petabyteboy:
Spatial storage is probably heavily broken (search the code for FIXME 1.17), but everything else should work fine

You might need to build [LibBlockAttributes](https://github.com/petabyteboy/LibBlockAttributes/tree/0.8.x-1.17.x) and [RoughlyEnoughItems](https://github.com/petabyteboy/RoughlyEnoughItems) and `publishToMavenLocal`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,6 @@ plugins {
 
 repositories {
     mavenLocal()
-    jcenter()
     mavenCentral()
     maven { url = "https://maven.fabricmc.net/" }
     maven {
--- a/build.gradle
+++ b/build.gradle
@@ -45,6 +45,7 @@ repositories {
             includeGroup "mcp.mobius.waila"
         }
     }
+    maven { url "https://maven.shedaniel.me" }
 }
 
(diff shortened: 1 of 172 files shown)


petabyteboy:
The meteorite compass doesn't work in some cases (multiplayer?).

shartte:
@petabyteboy Thanks a lot for your work, but you should probably hold off.
We just learned Forge will be moving to official Mojang mappings for 1.17 (including class names), and we will switch ASAP to fully use mojang mappings too, for both Forge and Fabric. This will make syncing both versions MASSIVELY easier.

Technici4n:
I also don't think there are plans to merge this PR before 1.17 pre-releases or even RCs start rolling, as it would mean having to port patches between 3 branches instead of 2.

yueh:
Closing due for the current idea of moving fabric to mcp or both branches to official mappings if possible. As well as first trying to synchronize both to an ideally equal state.

PR #5052, opened 2021-03-10
Switched to official mappings instead of mcp mappings

yueh:


shartte:
I'd close this for now until FG becomes capable of remapping the classnames. Thoughts?

yueh:
Class names would be really nice, so I would wait for them. But it also a massive disadvantage that FG, SRG or whatever adds the few comments no longer does it with the official mappings. So it's from like reverse engineering 75% of the code to understand it to 100%.

shartte:
Supposedly, they're working on getting parameter names and other crowd-source info merged into the official mappings

PR #5060, opened 2021-03-13
Fixes #5040: Fixes to chest closing sound timings.

### LINKED ISSUE #5040: Sky stone chest closing sound plays too early, also plays if REI interface opened
**Describe the bug**
Both variants of the sky stone chest play their closing sound immediately after the chest GUI is exited by any means. This means that it plays too early when closing the chest compared to the vanilla chest, and it also plays when the player views a recipe/uses in Roughly Enough Items when the chest GUI is open.

**To Reproduce**
- Start Minecraft with the Fabric mod loader, with AE2, [Roughly Enough Items](https://www.curseforge.com/minecraft/mc-mods/roughly-enough-items/files/3218065) and the Fabric API in the mods folder.
- Create a new world in creative mode.
- Place down a vanilla chest.
- Open the chest, then close it. Note when the closing sound plays.
- Place down either variant of the sky stone chest.
- Open the chest, then close it. Note how much earlier the closing sound plays.
- Open the chest once more, and left click any item in the REI interface on the right side of the screen with a crafting recipe to view its recipe. Note that the chest closing sound plays when the REI interface opens.

**Expected behavior**
The chest closing sound would play at the correct time for both variants of the sky stone chest, and would not play when the REI recipe interface is opened from within the chest.

**Environment**
Minecraft version: 1.16.5
Fabric Loader version: 0.11.1
AE2 version: 8.2.0 alpha 1
Fabric API version: 0.31.0
Roughly Enough Items version: 5.10.184

markfowden:
- Only track users on the server side.
- Only play the closing sound when the becomes halfway closed (like native MC).

(Also we should prefer not to use public getters and setters inside the class.)

### DIFF
--- a/src/main/java/appeng/tile/storage/SkyChestTileEntity.java
+++ b/src/main/java/appeng/tile/storage/SkyChestTileEntity.java
@@ -28,6 +28,7 @@ import net.minecraft.tileentity.IChestLid;
 import net.minecraft.tileentity.ITickableTileEntity;
 import net.minecraft.tileentity.TileEntityType;
 import net.minecraft.util.SoundCategory;
+import net.minecraft.util.SoundEvent;
 import net.minecraft.util.SoundEvents;
 import net.minecraft.util.math.MathHelper;
 import net.minecraftforge.api.distmarker.Dist;
@@ -58,16 +59,16 @@ public class SkyChestTileEntity extends AEBaseInvTileEntity implements ITickable
     @Override
     protected void writeToStream(final PacketBuffer data) throws IOException {
         super.writeToStream(data);
-        data.writeBoolean(this.getPlayerOpen() > 0);
+        data.writeBoolean(this.numPlayersUsing > 0);
     }
 
     @Override
     protected boolean readFromStream(final PacketBuffer data) throws IOException {
         final boolean c = super.readFromStream(data);
-        final int wasOpen = this.getPlayerOpen();
-        this.setPlayerOpen(data.readBoolean() ? 1 : 0);
+        final int wasOpen = this.numPlayersUsing;
+        this.numPlayersUsing = data.readBoolean() ? 1 : 0;
 
-        if (wasOpen != this.getPlayerOpen()) {
+        if (wasOpen != this.numPlayersUsing) {
             this.setLastEvent(System.currentTimeMillis());
         }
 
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/tile/storage/SkyChestTileEntity.java
```diff
@@ -118,22 +105,43 @@ private void onOpenOrClose() {
             this.world.addBlockEvent(this.pos, block, 1, this.numPlayersUsing);
             this.world.notifyNeighborsOfStateChange(this.pos, block);
             // FIXME: Uhm, we are we doing this?
-            this.world.notifyNeighborsOfStateChange(this.pos.down(), block);
+            // this.world.notifyNeighborsOfStateChange(this.pos.down(), block);
```
shartte, line 108: Please don't leave commented out code. Just remove it (the FIXME comment too)

shartte (changes requested):

src/main/java/appeng/tile/storage/SkyChestTileEntity.java
markfowden: Ah, missed this, thanks :)

shartte (approved):

src/main/java/appeng/tile/storage/SkyChestTileEntity.java
```diff
@@ -58,16 +59,16 @@ public SkyChestTileEntity(TileEntityType<? extends SkyChestTileEntity> type) {
...
     @Override
     protected boolean readFromStream(final PacketBuffer data) throws IOException {
         final boolean c = super.readFromStream(data);
-        final int wasOpen = this.getPlayerOpen();
-        this.setPlayerOpen(data.readBoolean() ? 1 : 0);
+        final int wasOpen = this.numPlayersUsing;
+        this.numPlayersUsing = data.readBoolean() ? 1 : 0;
 
-        if (wasOpen != this.getPlayerOpen()) {
+        if (wasOpen != this.numPlayersUsing) {
```
yueh, line 71: This path is actually unused now.

src/main/java/appeng/tile/storage/SkyChestTileEntity.java
```diff
@@ -142,14 +148,6 @@ public void onChangeInventory(final IItemHandler inv, final int slot, final InvO
...
 
-    public int getPlayerOpen() {
-        return this.numPlayersUsing;
-    }
-
-    private void setPlayerOpen(final int playerOpen) {
-        this.numPlayersUsing = playerOpen;
-    }
-
     public long getLastEvent() {
```
yueh, line 151: Also unused now, same with the getter for it and field.

src/main/java/appeng/tile/storage/SkyChestTileEntity.java
```diff
@@ -58,16 +59,16 @@ public SkyChestTileEntity(TileEntityType<? extends SkyChestTileEntity> type) {
...
+        data.writeBoolean(this.numPlayersUsing > 0);
     }
 
     @Override
     protected boolean readFromStream(final PacketBuffer data) throws IOException {
         final boolean c = super.readFromStream(data);
-        final int wasOpen = this.getPlayerOpen();
-        this.setPlayerOpen(data.readBoolean() ? 1 : 0);
+        final int wasOpen = this.numPlayersUsing;
+        this.numPlayersUsing = data.readBoolean() ? 1 : 0;
```
yueh, line 69: As `numPlayersUsing` is now used client and serverside, the comment about it being server only should probably also be removed.

yueh (changes requested):

yueh (approved):

PR #5064, opened 2021-03-14
Fixes #5061: Use isSneaking() instead of isCrouching()

### LINKED ISSUE #5061: Use `isSneaking()` in favour of `isCrouching()`
Like #5031, just for forge.

yueh:
For example this allows to interact with the part while flying and
pressing shift + spacebar to maintain the height.


shartte (changes requested):
As discussed on Discord, we should extract this to a helper since what is the difference between crouching vs. sneaking...
I'll submit another change to the branch

PR #5072, opened 2021-03-16
Fixes #4753: Use the vanilla helper for SUCCESS or CONSUME

### LINKED ISSUE #4753: Sometimes the Hand swings twice
I think our block-interaction code is using outdated semantics.
I've observed this with a little lag on Fabric, but I think the concept also holds on Forge.

We currently return SUCCESS in block activation callbacks on both sides, but SUCCESS will send out a hand-swing.
Vanilla follows this pattern:
- On the side that should trigger the hand-swing, return SUCCESS.
- On the other side, return CONSUME

There's a helper `net.minecraft.util.ActionResultType#func_233537_a_`, sadly unnamed,  that is used by vanilla in this way:
`return ActionResultType.func_233537_a_(world.isRemote)` indicating it swings the hand on the client, but not again on the server.

yueh:
This makes now use of the vanilla helper to return `SUCCESS` or
`CONSUME` depending on  the side.

Also some general cleanup like unused imports, missing file headers, etc


src/main/java/appeng/block/grindstone/GrinderBlock.java
```diff
@@ -47,11 +46,11 @@ public ActionResultType onActivated(final World w, final BlockPos pos, final Pla
             final @Nullable ItemStack heldItem, final BlockRayTraceResult hit) {
         final GrinderTileEntity tg = this.getTileEntity(w, pos);
         if (tg != null && !InteractionUtil.isInAlternateUseMode(p)) {
-            if (p instanceof ServerPlayerEntity) {
```
shartte, line 50: is p nullable and this elides a null-check? I don't remember

yueh: `InteractionUtil.isInAlternateUseMode(p)` guards against it by throwing a NPE before it even reaches the `instanceof`.
But otherwise no idea. Looks like an outlier. There is also no special handler here for fakeplayers, that is just for the crank.
Otherwise anything else follows the pattern with `getTile && !isRemote`

src/main/java/appeng/hooks/ticking/TickHandler.java
```diff
@@ -333,6 +333,8 @@ public void registerCraftingSimulation(final World world, final CraftingJob craf
 
     /**
      * Simulates the current crafting requests before they user can submit them to be processed.
+     *
+     * @param world
```
shartte, line 337: I think we're just turning in circles .I remove them, you add them back :-D

yueh: Rebase with conflicts ftw.

src/main/java/appeng/items/tools/quartz/QuartzWrenchItem.java
```diff
@@ -41,21 +42,25 @@ public QuartzWrenchItem(Item.Properties props) {
...
 
-            Block block = context.getWorld().getBlockState(context.getPos()).getBlock();
+        if (!InteractionUtil.isInAlternateUseMode(p) && Platform
+                .hasPermissions(new DimensionalCoord(w, pos), p)) {
+
+            Block block = w.getBlockState(pos).getBlock();
             if (block instanceof AEBaseBlock) {
-                if (Platform.isClient()) {
+                if (w.isRemote()) {
                     // TODO 1.10-R - if we return FAIL on client, action will not be sent to server.
```
shartte, line 55: Should we just remove this

src/main/java/appeng/items/tools/quartz/QuartzWrenchItem.java
```diff
@@ -41,21 +42,25 @@ public QuartzWrenchItem(Item.Properties props) {
...
-                    return !context.getWorld().isRemote ? ActionResultType.SUCCESS : ActionResultType.PASS;
+                    return !w.isRemote() ? ActionResultType.func_233537_a_(w.isRemote()) : ActionResultType.PASS;
                 }
 
                 AEBaseBlock aeBlock = (AEBaseBlock) block;
-                if (aeBlock.rotateAroundFaceAxis(context.getWorld(), context.getPos(), context.getFace())) {
-                    context.getPlayer().swingArm(context.getHand());
-                    return !context.getWorld().isRemote ? ActionResultType.SUCCESS : ActionResultType.FAIL;
+                if (aeBlock.rotateAroundFaceAxis(w, pos, context.getFace())) {
+                    p.swingArm(context.getHand());
```
shartte, line 62: This seems hyper-fishy.... Is this because we're in onitemusefirst?

shartte (commented):
LGTM except for swingHand in the wrench code, no idea why that is there. It may behave differently for onItemUseFirst

src/main/java/appeng/items/tools/quartz/QuartzWrenchItem.java
yueh: No idea and I do not really want to touch it currently.

yueh:
There are a couple things overall missing. 

* The wrench does not swing when picking up a cable, however the network tool does.
* Wrench does not break ME blocks currently.
* Stuff I have already forgotten again.

PR #5076, opened 2021-03-18
Removed more uses of Platform#isClient/#isServer

shartte:


src/main/java/appeng/fluids/helper/DualityFluidInterface.java
```diff
@@ -403,6 +404,12 @@ public void setPriority(final int newValue) {
         this.priority = newValue;
     }
 
+    @Override
+    public boolean isRemote() {
+        World world = this.iHost.getTileEntity().getWorld();
+        return world == null || world.isRemote();
```
yueh, line 410: These checks might be tricky. The world can be null for server worlds for some time. iirc forge adds a hack to set it immediately, but for vanilla itself it was like one tick later.  Might also be interesting when transfering it via spatial io.

yueh (approved):

PR #5077, opened 2021-03-18
Some minor refactorings of AENetworkProxy

yueh:


src/main/java/appeng/me/helpers/AENetworkProxy.java
```diff
@@ -64,7 +64,7 @@
     private final String nbtName; // name
     private AEColor myColor = AEColor.TRANSPARENT;
     private CompoundNBT data = null; // input
-    private ItemStack myRepInstance = ItemStack.EMPTY;
```
yueh, line 64: This is a safeguard as we do not check it for null anywhere and some addons are reusing our internal classes, which can lead to NPEs.

src/main/java/appeng/me/helpers/AENetworkProxy.java
```diff
@@ -325,14 +325,6 @@ public IEnergyGrid getEnergy() throws GridAccessException {
 
     @Nonnull
     private <T extends IGridCache> T getGridCache(Class<T> clazz) throws GridAccessException {
-        final IGrid grid = this.getGrid();
-        if (grid == null) {
-            throw new GridAccessException();
-        }
-        final T eg = grid.getCache(clazz);
-        if (eg == null) {
-            throw new GridAccessException();
```
yueh: `GridAccessException` are actually often catched. In this case we would replace it with a NPE, which is usually not caught. It probably will not cause them, but the grid construction is not trivial, so it can run into edge cases. Also we should probably change the `Grid` constructor into a factory, so it no longer leaks the object through the event bus before fully initialized. 
This should eliminate the need to check as the caches a populated during the constructor, assuming nothing breaks it and it suddenly queries a non existing cache.

src/main/java/appeng/me/helpers/AENetworkProxy.java
shartte: No, it is set in every constructor. This was an IntelliJ warning (that the value set here is never used since every declared constructor overwrites it)

src/main/java/appeng/me/helpers/AENetworkProxy.java
shartte: No, since if you trace this value back, it CANNOT be null at this point since it was already de-referenced in getCache

src/main/java/appeng/me/helpers/AENetworkProxy.java
yueh: Ah yeah. in that case wrapping it into an `Objects.requireNonNull` would be better. And probably other fields like `gp`.

shartte (approved):

PR #5084, opened 2021-03-21
Fixes #5083: Include the forge cap NBT when writing to a buffer

### LINKED ISSUE #5083: Unable to withdraw Psi CAD from ME storage
**Describe the bug**

After inserting an assembled CAD of any kind from the mod Psi into an ME storage cell, it becomes impossible to get it back out, or even interact with it in any way. I've reproduced this in a newly created world, with a minimal setup.
<details>
<summary>Screenshots</summary>

Minimal setup

Psi CAD in ME terminal after insertion. No amount of clicking will get it back out again.

</details>

**To Reproduce**

1. Assemble any kind of CAD from Psi by putting a CAD assembly in the CAD assembler machine.
2. Put the finished CAD into an ME storage cell through a terminal, interface, ME chest or any other kind of IO.
3. Try extracting the CAD from said storage.

**Expected behavior**

Being able to extract the CAD from ME storage or interact with it in any way after inserting it.

**Additional context**

This isn't the first time this happened to me in practice. I have 2 Iron CADs stuck in my system, my friend recently added a Psimetal one to the growing collection. I have tried every workaround and IO interface I could think of, but I cannot get them out again. Please help.

**Environment**

I've encountered this issue in at least two different modpacks that I can recall the version of:

- **All the Mods 6 v1.5.5**
  - Minecraft 1.16.5
  - Forge 36.1.0
  - appliedenergistics2-8.3.0-alpha.2
  - Psi 1.16-91
- All the Mods 6 v1.5.1
  - Minecraft 1.16.5
  - Forge 36.0.13
  - appliedenergistics2-8.2.0-beta.2
  - Psi 1.16-91
- Limitless 3 v1.9.2
  - Minecraft 1.16.5
  - Forge 36.0.1
  - appliedenergistics2-8.2.0-beta.2
  - Psi 1.16-91

Both were started in MultiMC, using the newest Java 8 runtime, with a memory limit of 8 GiB and no additional command line arguments.

Let me know if you need any other details.

Related issues describing the same problem: #4882, #4816

yueh:


src/main/java/appeng/util/item/AEItemStack.java
```diff
@@ -108,7 +108,19 @@ public static AEItemStack fromPacket(final PacketBuffer buffer) {
...
         final long stackSize = buffer.readVarLong();
         final long countRequestable = buffer.readVarLong();
-        final ItemStack itemstack = buffer.readItemStack();
+
+        // based on buffer.readItemStack()
+        // Adapted to also handle forge capabilities
+        ItemStack itemstack = ItemStack.EMPTY;
+        if (buffer.readBoolean()) {
+            int i = buffer.readVarInt();
+            int j = buffer.readByte();
```
shartte, line 117: i, j. come on! :-P

src/main/java/appeng/util/item/AEItemStack.java
```diff
@@ -126,6 +138,9 @@ public void writeToPacket(final PacketBuffer buffer) {
         buffer.writeVarLong(this.getStackSize());
         buffer.writeVarLong(this.getCountRequestable());
         buffer.writeItemStack(getDefinition(), false);
```
shartte, line 140: Is it a good idea to rely on writeItemStack here given that we now do the reading ourselves? I'd be for inlining this method here to make it mirror our reading code.

src/main/java/appeng/util/item/AEItemStack.java
```diff
@@ -108,7 +108,19 @@ public static AEItemStack fromPacket(final PacketBuffer buffer) {
...
+        // Adapted to also handle forge capabilities
+        ItemStack itemstack = ItemStack.EMPTY;
+        if (buffer.readBoolean()) {
+            int i = buffer.readVarInt();
+            int j = buffer.readByte();
+            CompoundNBT nbt = buffer.readCompoundTag();
+            CompoundNBT forgeCap = buffer.readCompoundTag();
+
+            itemstack = new ItemStack(Item.getItemById(i), j, forgeCap);
+            itemstack.readShareTag(nbt);
```
shartte, line 122: We're serializing the full tag, but read the share-tag here... Why?

src/main/java/appeng/util/item/AEItemStack.java
yueh: So we can just copy&paste it once forge changes something.

src/main/java/appeng/util/item/AEItemStack.java
yueh: No idea. That's how vanilla/forge works and I don't like to change it and caus a different bug.

src/main/java/appeng/util/item/AEItemStack.java
yueh: Might be an idea. On the other hand we will get an immediate response when something in forge/vanilla changes instead of missing it because we duplicated the current behaviour.

src/main/java/appeng/util/item/AEItemStack.java
shartte: We should be aware that we're breaking their contract here.
They allow items to override readShareTag, and they might expect that they do get a share tag, and not the full server tag.....

PR #5102, opened 2021-03-28
ItemList Fixes and Cleanups

shartte:
- Switched to JUnit 5
- Added tests for ItemList
- Fixed ItemList/SubList Iterators hasNext breaking the iterator contract and advancing the iterator multiple times.


yueh:
lgtm, the only point of uncertainty would be the changes to the iterator, as this could reintroduce an old bug with CMEs.
This whould be ok with JDK collections as they would fail fast, should it appear again. However by using the fastutil ones it will be swallowed and can cause undefined behaviour.

src/api/java/appeng/api/config/FuzzyMode.java
```diff
@@ -24,9 +24,32 @@
 package appeng.api.config;
 
 public enum FuzzyMode {
-    // Note that percentage damaged, is the inverse of percentage durability.
-    IGNORE_ALL(-1), PERCENT_99(0), PERCENT_75(25), PERCENT_50(50), PERCENT_25(75);
+    /**
+     * Matches between 0% and 100% durability.
```
yueh, line 28: `between` is not really defined as inclusive or exclusive and I can assume someone could consider it being exclusive here when mentioned explicitly like all other breakpoints.

Maybe just use `Matches any durability`?

src/api/java/appeng/api/config/FuzzyMode.java
```diff
@@ -24,9 +24,32 @@
...
+     * Otherwise items with 50% durability or more are matched.
+     */
+    PERCENT_50(50),
+    /**
+     * If an item with less than 25% durability is used as the filter, items with less than 50% durability are matched.
+     * Otherwise items with 25% durability or more are matched.
+     */
+    PERCENT_25(75);
 
+    // Note that percentage damaged, is the inverse of percentage durability.
```
yueh, line 52: Maybe move it to the class javadoc? It is still a pretty important fact and here it looks a bit too easy to overlook.

src/main/java/appeng/util/item/FuzzyItemVariantList.java
```diff
@@ -0,0 +1,179 @@
+/*
+ * This file is part of Applied Energistics 2.
+ * Copyright (c) 2013 - 2020, AlgorithmX2, All rights reserved.
```
yueh, line 3: We should move the new or nearly completey files to `Copyright (c) 2021, TeamAppliedEnergistics, All rights reserved.`. Or whatever year will apply.

src/main/java/appeng/util/item/FuzzyItemVariantList.java
```diff
@@ -0,0 +1,179 @@
...
+ * This variant list is optimized for damageable items, and supports selecting durability ranges with
+ * {@link #findFuzzy(IAEItemStack, FuzzyMode)}.
+ */
+class FuzzyItemVariantList extends ItemVariantList {
+
+    static final SharedStackComparator COMPARATOR = new SharedStackComparator();
+
+    // NOTE: We only use Object as they key here so we can pass our special DamageBounds to the subMap method.
+    // We NEVER put any keys in this map that are not AESharedItemStacks.
+    private final Object2ObjectSortedMap<Object, IAEItemStack> records = new Object2ObjectAVLTreeMap<>(COMPARATOR);
```
yueh, line 45: We might also look into using `Object2ObjectRBTreeMap` as replacement here. AVL should be a bit faster for retrieval, but a bit more costly on insert. Hard to guess in advance really and also hard to reliably benchmark under real conditions.

The other hash based ones are not alternative as none implements `subMap()`

src/main/java/appeng/util/item/ItemList.java
```diff
@@ -63,6 +66,8 @@ public boolean isEmpty() {
 
     @Override
     public void add(final IAEItemStack itemStack) {
+        version.incrementAndGet();
```
yueh, line 69: Is this here on purpose to crash on trying to add something, even it the collections is unmodified? Otherwise it might be an idea to move it behind the null check. Or even move it into the `create` part of `getOrCreateRecord`.

src/main/java/appeng/util/item/FuzzyItemVariantList.java
shartte: Yeah, I'd leave it as-is for now and just observe.

yueh (approved):

PR #5103, opened 2021-03-28
Fixes #4273: Add teminalStyle button to Interface terminal

### LINKED ISSUE #4273: Bigger/ Customizeable Interface terminal GUI
#### Description
Quality of life Improvement, allow the player to select tall centered terminal option on ME Interface terminal, 6 rows are not many, especially when you have more than 200 Interfaces to search into for an empty space.
I find the small GUI extremely annoying while playing Enigmatica 2 Expert, and down here you'll find some screenshot that will show you, side by side, the size of the crafting terminal compared to the terminal and the Interface terminal

#### Screenshots
The 3 terminals near eachother
Interface Terminal alone on my big screen

markfowden:
Some basic changes for adding terminalStyle to the InterfaceTerminalScreen. To do so:
- I've resized the interfaceterminal GUI texture to be the minimum required for drawBG.
- Changed the drawBG and drawFG logic to be dependent on a numLines which is influenced by the terminalStyle button.
- Things are complicated a little bit by duplication of shading lines on the top and bottom of the scrollable area depending on the type of the first and last line, so there's some special logic for these cases in rendering.
- Reposition the player inventory slots after determining the number of lines.

Additionally I've made some small changes to the EnumCycler to use modulo instead of if statements, since modulo naturally cycles on a number range.
And also fixed 5 javadoc warnings.

How it looks:


yueh:
First many thanks, but a couple generic remarks first. Hopefully not off-putting too much.

Refactorings and new features should ideally be split into different PRs. Makes it easier to determine which one broke something and not potentially both. Or we just got lucky and for now eachother cancels out, but it will pop up at random and unexpected. Probably fine for now. But just for the future.

Bit shifting or similar should probably be avoided when it's not performance relevant. Rule of thumb should really that when having the desire to add more comment lines/words than the removed code had and it is already obvious to cause some raised eyebrows in 2 weeks, then it is most likely not worth it in the long run.

Last but not least, absolutely no more magic numbers within the GUI code. It already is a mess trying to change anything there without running into various equal but unrelated numbers and there is nearly no option besides checking each and everyone to determine which effect they have. Sucks, but certainly better to apply it to the limited scope of the updated classes instead of trying to apply it all at once to the whole GUI.
So no more `18` scattered all around. It really needs to be replaced with a meaningful constant. I am pretty sure some `18` are not the row height, but some random offset which has to stay `18`, even when the row height should ever change.
Same for any precomputed value of two other constants, just use both or maybe add a new constant expression should it make it more obvious to have a more fitting name as opposed to adding two or more constants. Especially if there is also the desire to add further comments to make it obvious. With the comment even using numbers which do not match the code at all. Further the javac will take care of constant folding, so we should really focus on correctness and comprehensibility over a few cpu cycles saved during compile.

Mari023:
18 is probably the size of a slot

yueh:
In other words you verified the underlying problem by having to guess what it might be.

markfowden:
> Refactorings and new features should ideally be split into different PRs. Makes it easier to determine which one broke something and not potentially both. Or we just got lucky and for now eachother cancels out, but it will pop up at random and unexpected. Probably fine for now. But just for the future.

Yep, I agree, just thought I could squeeze this two lil things in, but I can separate them out if you'd like :)

> Bit shifting or similar should probably be avoided when it's not performance relevant. Rule of thumb should really that when having the desire to add more comment lines/words than the removed code had and it is already obvious to cause some raised eyebrows in 2 weeks, then it is most likely not worth it in the long run.

Apologies but could you please point me to where the bit shifting is?

> Last but not least, absolutely no more magic numbers within the GUI code.

I completely agree, there's a lot of magic numbers. I found I had to open photoshop and measure pixels to figure out what the numbers corresponded to. I've tried to reduce the number of existing magic numbers with this diff, but I agree I could do more. And it's further complicated by the off-by-one that the duplication of the shading line causes.

In summary I agree with all your points yueh and there's definitely some great refactoring to be done on the GUIs. I had the beginnings of an attempt to do so but it's a mammoth task. In this instance I would've sorted the rest of this class out but a) I unfortunately don't have the time and b) it would probably ultimately be out of scope of the feature request this diff resolves.

shartte:
Hm, the player inventory doesn't seem to be correctly aligned:
image

I did fix some minor things I did notice, but am not done reviewing yet.
I did add something that was probably already borked (forwarding mouse clicks to the search field)

shartte:
Fixed the player slot positioning and did a bit of refactoring. From my end this is now good to go.

shartte (approved):

shartte (changes requested):
Oh wait, I forgot. The EnumCycler changes should probably be rolled back and re-applied with added unit tests.

yueh:
Yes, the enum cycler with the `%` operator is what I meant with the bit shifting. Also there is `Math.floorMod()`, which should do the same, is better documented and certainly better tested. It probably works, but it's always hard to guess every edge case by just scanning through the code.

yueh:
Pushed another commit with more changes. See the commit message.

The enum cycler is probably fine. I just replaced it with the JDK helper. Some quick tests did work correctly, so I would say keep it and write tests later as the intention is pretty clear and it is easy to add tests for it based on that (and then see how it breaks).

markfowden:
Awesome, thanks so much both :)

PR #5105, opened 2021-04-01
Cleaning up the ME GUIs (Items+Fluids)

shartte:


src/main/java/appeng/client/Point.java
```diff
@@ -12,6 +14,10 @@ public Point(int x, int y) {
         this.y = y;
     }
 
+    public static Point fromTopLeft(Rectangle2d bounds) {
```
yueh, line 17: Would it make sense to add this for all four corners? Mostly with the idea of positioning them relative to the main screen? E.g. upgrades/view cells are always relative to the `topRight`, network tool would be `bottomRight` and so on.
When changing something, this would avoid having to update all related widgets on top.

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -401,6 +382,21 @@ private void drawOptionalSlotBackground(MatrixStack matrixStack, IOptionalSlot s
...
 
+    // Convert global mouse x,y to relative Point
+    private Point getMousePoint(double x, double y) {
+        return new Point((int) Math.round(x - guiLeft), (int) Math.round(y - guiTop));
+    }
+
+    @Override
+    public boolean mouseScrolled(double x, double y, double wheelDelta) {
+        if (wheelDelta != 0) {
+            if (widgets.onMouseWheel(getMousePoint(x, y), wheelDelta)) {
```
yueh, line 393: Is `onMouseWheel` being used in vanilla? If so that is fine, however I would prefer using the same name whe forwarding a call, if possible. Keeps it consistent. And if `mouseScrolled` is not vanilla, but we cannot change it now, probably add a `@TODO`

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -426,23 +422,17 @@ public boolean mouseClicked(final double xCoord, final double yCoord, final int
             }
         }
 
-        // Forward left mouse button down events to the scrollbar
-        if (btn == 0 && this.getScrollBar() != null) {
-            if (this.getScrollBar().mouseDown(xCoord - this.guiLeft, yCoord - this.guiTop)) {
-                return true;
-            }
+        if (widgets.onMouseDown(getMousePoint(xCoord, yCoord), btn)) {
```
yueh, line 425: Same here. `mouseClicked` vs `onMouseDown`.

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -426,23 +422,17 @@ public boolean mouseClicked(final double xCoord, final double yCoord, final int
...
     }
 
     @Override
     public boolean mouseReleased(double mouseX, double mouseY, int button) {
-        // Forward left mouse button up events to the scrollbar
-        if (button == 0 && this.getScrollBar() != null) {
-            if (this.getScrollBar().mouseUp(mouseX - this.guiLeft, mouseY - this.guiTop)) {
-                return true;
-            }
+        if (widgets.onMouseUp(getMousePoint(mouseX, mouseY), button)) {
```
yueh, line 434: `mouseReleased` vs `onMouseUp`

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -453,8 +443,9 @@ public boolean mouseDragged(double mouseX, double mouseY, int mouseButton, doubl
         final Slot slot = this.getSlot((int) mouseX, (int) mouseY);
         final ItemStack itemstack = getPlayer().inventory.getItemStack();
 
-        if (this.getScrollBar() != null) {
-            this.getScrollBar().mouseDragged((int) mouseX - this.guiLeft, (int) mouseY - this.guiTop);
+        Point mousePos = new Point((int) Math.round(mouseX - guiLeft), (int) Math.round(mouseY - guiTop));
+        if (widgets.onMouseDrag(mousePos, mouseButton)) {
```
yueh, line 447: `mouseDragged` vs `onMouseDrag`

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -656,7 +638,7 @@ protected void moveItems(MatrixStack matrices, Slot s) {
             try {
                 renderAppEngSlot(matrices, (AppEngSlot) s);
             } catch (final Exception err) {
-                AELog.warn("[AppEng] AE prevented crash while drawing slot: " + err.toString());
+                AELog.warn("[AppEng] AE prevented crash while drawing slot: " + err);
```
yueh, line 641: A bit OT, but we should check at some point about removing `[AppEng]` or similar tags. The log should already contain the actual logger name. I'd consider doing it afterwards, to avoid merge conflicts now.

src/main/java/appeng/client/gui/me/common/MEMonitorableScreen.java
```diff
@@ -136,6 +136,38 @@ public MEMonitorableScreen(C container, PlayerInventory playerInventory,
         } else {
             viewCellBg = null;
         }
+
+        if (this.style.isSupportsAutoCrafting()) {
```
yueh, line 140: Just `supportsAutoCrafting` or if the `is` needs to be kept for boolean at least `isSupportingAutoCrafting`?
Looks a bit strange currently.

src/main/java/appeng/client/gui/me/common/MEMonitorableScreen.java
```diff
@@ -136,6 +136,38 @@ public MEMonitorableScreen(C container, PlayerInventory playerInventory,
...
+            this.craftingStatusBtn = new TabButton(2 + 11 * 16,
+                    GuiText.CraftingStatus.text(), this.itemRenderer, btn -> showCraftingStatus());
+            this.craftingStatusBtn.setHideEdge(true);
+            this.widgets.add("craftingStatus", this.craftingStatusBtn);
+        }
+
+        if (this.style.hasSortByButton()) {
```
yueh, line 147: Maybe also something like `isSorted`, `isSortable` or `canBeSorted`?

src/main/java/appeng/client/gui/AEBaseScreen.java
shartte: We have to make a pass once we migrate to actual official names. MCP is all over the place.

src/main/java/appeng/client/gui/AEBaseScreen.java
shartte: onMouseUp is web-speak

src/main/java/appeng/client/gui/AEBaseScreen.java
shartte: See above w.r.t. naming

src/main/java/appeng/client/gui/AEBaseScreen.java
shartte: Yeah, I have no idea where that is coming from.

src/main/java/appeng/client/gui/me/common/MEMonitorableScreen.java
shartte: It's just the awful java bean getter naming though :-|

src/main/java/appeng/client/gui/me/common/MEMonitorableScreen.java
shartte: Renamed

src/main/java/appeng/client/gui/AEBaseScreen.java
shartte: Also webspeak

src/main/java/appeng/client/Point.java
shartte: Adding it makes no sense right now because it wouldn't be used :D

starcatmeow:
Thanks a lot for huge amazing work you've done, but when I downloaded the CI dist mod and loaded it to my server. However when the client click on the crafting terminal, the server crashes and seems because the mod called a client api.

Here's the crash report.

https://gist.github.com/starcatmeow/31dd5c45594ce052f8ad5ded54cd0cf3

shartte:
Thanks, I did indeed forget to test it on a dedicated server. Will investigate.

gradle.properties
```diff
@@ -1,5 +1,5 @@
-version_major=0
-version_minor=0
+version_major=8
+version_minor=3
```
yueh, line 2: Please revert

yueh (changes requested):

gradle.properties
shartte: Ah fuck, accident. I had to do this locally to make it load in ATM for testing :-(

yueh (approved):

PR #5110, opened 2021-04-03
Fixes CME in CraftingCPU cluster when moving items back into the network

### LINKED ISSUE #5109: ChainedIterator ConcurrentModificationException
ChainedIterator ConcurrentModificationException happening in ``storeItems`` in ``CraftingCPUCluster``

**Describe the bug**
Canceling a craft can cause a CME in certain situations, usually on low space

**To Reproduce**
Attached are two zipped worlds. One will crash on load with the CME, the other one 
can crash by ask for 10 vibration chambers and then cancelling the crafting on the crafting status gui.

**Expected behavior**
Not crashing

**Additional context**
Crash log
https://gist.github.com/PrototypeTrousers/5140b1b2de38e08abab9227587f6bd74#file-chainediteratorcme
The CME can be avoided by copying the list before iterating but maybe there is a proper solution
```
IItemList<IAEItemStack> iList = Api.instance().storage().getStorageChannel(IItemStorageChannel.class).createList();
this.inventory.getItemList().forEach(iList::add);
for( IAEItemStack is : iList ) {
```

**Environment**
AdoptopenJDK 8u275
version: 0bdc408e950fe787277d1e48be61ebf928127429

[Ask to craft then cancel craft to crash.zip](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/6251754/Ask.to.craft.then.cancel.craft.to.crash.zip)
[Chained CME on load.zip](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/6251755/Chained.CME.on.load.zip)

shartte:
Fixes #5109: Don't mutate the item list while iterating over it.

This is also slightly more optimized than the old version.

### DIFF
--- a/src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
+++ b/src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
@@ -379,9 +379,7 @@ public final class CraftingCPUCluster implements IAECluster, ICraftingCPU {
             final Collection<CraftingWatcher> list = sg.getInterestManager().get(diff);
 
             if (!list.isEmpty()) {
-                for (final CraftingWatcher iw : list)
-
-                {
+                for (final CraftingWatcher iw : list) {
                     iw.getHost().onRequestChange(sg, diff);
                 }
             }
@@ -779,6 +777,9 @@ public final class CraftingCPUCluster implements IAECluster, ICraftingCPU {
         }
     }
 
+    /**
+     * Tries to dump all locally stored items back into the storage network.
+     */
     private void storeItems() {
         final IGrid g = this.getGrid();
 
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -790,20 +791,21 @@ private void storeItems() {
...
+        IItemList<IAEItemStack> itemList = this.inventory.getItemList();
+        for (IAEItemStack is : itemList) {
+            this.postChange(is, this.machineSrc);
+            IAEItemStack remainder = ii.injectItems(is.copy(), Actionable.MODULATE, this.machineSrc);
 
-            if (is != null) {
-                this.inventory.injectItems(is, Actionable.MODULATE, this.machineSrc);
+            // The network was unable to receive all of the items, i.e. no or not enough storage space left
+            if (remainder != null) {
+                is.setStackSize(remainder.getStackSize());
```
yueh, line 801: I am not sure how this will behave in some extreme cases like size > `Integer.MAX_VALUE`. Crafting is certainly one of the few cases, which can run into these.

shartte: This is all IAEItemStack, so it should be all longs?

yueh: Yes, but we should clamp these to `Integer.MAX_VALUE` when they leave the network, like being stored in an external storage. And I do not think that is actually safeguarded at all.
This is usually fine as nothing else should move more than a few k at once. Crafting is really the only case where it can actually happen. But that might already apply for the current code.

shartte: Yeah it was already stored in the CPU when it gets to this point, so I don't think this is any more "unsafe" than before, really.

PR #5118, opened 2021-04-06
Cleanup cable bus drop logic

Technici4n:
I removed the gross `disableDrops()` hacks, and used `Block#getDrops()` instead. This will also remove cable creative drops on Fabric when it's merged.


src/main/java/appeng/util/Platform.java
```diff
@@ -274,16 +275,7 @@ public static boolean checkPermissions(final PlayerEntity player, final Object a
...
-                    if (i.getCount() > 0) {
-                        final double offset_x = (getRandomInt() % 32 - 16) / 82;
-                        final double offset_y = (getRandomInt() % 32 - 16) / 82;
-                        final double offset_z = (getRandomInt() % 32 - 16) / 82;
-                        final ItemEntity ei = new ItemEntity(w, 0.5 + offset_x + pos.getX(),
-                                0.5 + offset_y + pos.getY(), 0.2 + offset_z + pos.getZ(), i.copy());
-                        w.addEntity(ei);
-                    }
-                }
+                InventoryHelper.spawnItemStack(w, pos.getX(), pos.getY(), pos.getZ(), i);
```
yueh, line 278: This is something we cannot use as it will split the itemstacks into a shitton of additional entities. Worst case like a factor of 7 times more entities.

src/main/java/appeng/tile/AEBaseTileEntity.java
```diff
@@ -434,7 +424,6 @@ public boolean hasCustomInventoryName() {
 
     public void securityBreak() {
         this.world.destroyBlock(this.pos, true);
-        this.disableDrops();
```
yueh, line 437: The whole method is currently unused. This needs to be investigated in case we did accidentally break it and it still needs to ability to disabled drops.

src/main/java/appeng/parts/CableBusContainer.java
```diff
@@ -877,34 +877,36 @@ public void readFromNBT(final CompoundNBT data) {
         this.getFacadeContainer().readFromNBT(data);
     }
 
-    public List<ItemStack> getDrops(final List<ItemStack> drops) {
+    /**
+     * Append the contents of the parts to the drop list, e.g. crafting patterns but not the parts or facades
+     * themselves.
+     */
+    public void appendPartContentDrops(final List<ItemStack> drops) {
```
yueh, line 884: This probably needs a better name.  And overall I would prefer to handle it within the part itself by passing the loot context or similar around and have the parts deal with it and not have a bunch of different code paths, but this requires an API break.

src/main/java/appeng/parts/CableBusContainer.java
```diff
@@ -877,34 +877,36 @@ public void readFromNBT(final CompoundNBT data) {
...
         }
-
-        return drops;
     }
 
-    public List<ItemStack> getNoDrops(final List<ItemStack> drops) {
+    /**
+     * Append the parts and facades to the drop list, but not their contents.
+     */
+    public void appendPartStacks(final List<ItemStack> drops) {
```
yueh, line 896: Same here, the part itself should really handle it and not hide the drops somewhere else. But again API break, so more for 1.17

yueh (changes requested):

src/main/java/appeng/parts/CableBusContainer.java
Technici4n: Should we have a list of possible API breaks for 1.17 somewhere so we don't forget about them?

Technici4n: What about `appendPartInventories`?

src/main/java/appeng/util/Platform.java
Technici4n: Are we dropping that many item stacks?

shartte: also wdym? how is this different from before?
i cant look at spawnItemStack vs. spawning ourselves

yueh: We use it in a couple of places. Like destroying a crafting cpu and droping the inventory. Which is already bad.
But the helper will further split the stack into entities containing 10-30 items, so worst case 64 items turn into 7 entities and best case 3.

yueh: Also this mutates the actually itemstack being passed to it. So no idea, if this isn't causing other side effects by not copying it.

shartte: Now had a chance to look at this, and this is correct. And I would not have expected that based on the function name lol.

Technici4n: Reverted, my bad.

src/main/java/appeng/parts/CableBusContainer.java
Technici4n: Marking as resolved.

src/main/java/appeng/tile/AEBaseTileEntity.java
Technici4n: `securityBreak` is used for non-part tiles such as full-block interfaces. I tested this with interfaces containing an inventory and it's working fine. There is really no reason to disable the drops because the block is already destroyed.

Fyi this was added in the original security commit in 2014, it's probably not relevant anymore assuming it ever was to begin with.

src/main/java/appeng/parts/CableBusContainer.java
```diff
@@ -877,34 +877,36 @@ public void readFromNBT(final CompoundNBT data) {
...
         for (final AEPartLocation s : AEPartLocation.values()) {
             final IPart part = this.getPart(s);
             if (part != null) {
-                part.getDrops(drops, false);
+                drops.add(part.getItemStack(PartItemStack.BREAK));
```
shartte, line 900: should this do an isEmpty() check on the returned stack in case an addon goes crazy?

yueh: We already use this in various other places like serialisation, so it would already cause issues. Also the helper method itself already ignores any empty itemstack. The whole thing really needs to be redesigned to delegate it fully to the part.

src/main/java/appeng/tile/AEBaseTileEntity.java
Technici4n: Now that #5127 is merged, I need to investigate if this function is useful to have at all.

shartte: This implements securityBreak for IGridHost, even though AEBaseTileEntity does not implement IGridHost. Don't ask.... other tile entities that implement IGridHost implement that function by inheriting it from AEBaseTile -> blergh.

shartte: I documented why this method exists since it's not obvious, and thus it should be fine.

shartte (approved):

src/main/java/appeng/tile/AEBaseTileEntity.java
Technici4n: Yup, looks good.

yueh: This is still not resolved. The whole code path no longer exists and might result in unexpected behaviour. E.g. there might be a special case in which a player is able to pickup a drive with its cells and it then suddenly drops the whole drive with cells included plus additionally the cells itself again. I just have no idea how to test it currently. Or in which cases it might actually apply (and has to).

Technici4n: > The whole code path no longer exists.

Then how is securityBreak() handled for tiles such as a drive?

> I just have no idea how to test it currently. Or in which cases it might actually apply (and has to).

That's what alphas are for.

Technici4n:
Better to do this from scratch with some API changes in mind.

PR #5119, opened 2021-04-06
Remove ObfuscationReflectionHelper usage and move MolecularAssemblerRenderer to the correct package

Technici4n:


yueh:
Instead of relying on reflections, AT, mixins, or whatever else, we should simply extend our own `RenderType` similar to the already existing ones for the [spatial overlay](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/7c3f64eea0f4103b850ad087fb724b4264581fd5/src/main/java/appeng/client/render/overlay/OverlayRenderType.java#L43-L54). So we also do not depend on some other mod deciding to hack into it and modify it.

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
```diff
@@ -37,16 +37,18 @@
...
 import net.minecraft.item.ItemStack;
 import net.minecraft.util.ResourceLocation;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.api.distmarker.OnlyIn;
 import net.minecraftforge.client.model.data.EmptyModelData;
-import net.minecraftforge.fml.common.ObfuscationReflectionHelper;
 
 import appeng.client.render.effects.ParticleTypes;
 import appeng.core.AppEng;
+import appeng.tile.crafting.AssemblerAnimationStatus;
```
yueh, line 50: `AssemblerAnimationStatus` also sounds like something rendering related.

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
```diff
@@ -56,7 +58,7 @@
 
     public static final ResourceLocation LIGHTS_MODEL = AppEng.makeId("block/molecular_assembler_lights");
 
-    private static final RenderType MC_161917_RENDERTYPE_FIX = createRenderType();
+    private static final RenderType MC_161917_RENDERTYPE_FIX = RenderTypeAccess.createRenderType();
```
yueh, line 61: This isn't some fancy hack to replace an AT. Just a custom rendertype for the molecular assembler (or maybe a workaround type for the vanilla bug)

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
```diff
@@ -139,22 +143,30 @@ private void renderStatus(MolecularAssemblerTileEntity molecularAssembler, Matri
     }
 
     /**
-     * See above for when this can be removed. It creates a RenderType that is equivalent to
-     * {@link RenderType#getTranslucent()}, but enables alpha testing. This prevents the fully transparents parts of the
-     * rendered block model from occluding our particles.
+     * This class gives us access to the protected RenderState.TRANSLUCENT_TRANSPARENCY field.
```
yueh, line 146: It is somewhat obvious that extending another class allows it to access protected fields or methods.

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
```diff
@@ -139,22 +143,30 @@ private void renderStatus(MolecularAssemblerTileEntity molecularAssembler, Matri
...
-        RenderState.TextureState mipmapBlockAtlasTexture = new RenderState.TextureState(
-                AtlasTexture.LOCATION_BLOCKS_TEXTURE, false, true);
-        RenderState.LightmapState disableLightmap = new RenderState.LightmapState(false);
-        RenderType.State glState = RenderType.State.getBuilder().texture(mipmapBlockAtlasTexture)
-                .transparency(TRANSLUCENT_TRANSPARENCY).alpha(new RenderState.AlphaState(0.05F))
-                .lightmap(disableLightmap).build(true);
-
-        return RenderType.makeType("ae2_translucent_alphatest", DefaultVertexFormats.POSITION_COLOR_TEX_LIGHTMAP,
-                GL11.GL_QUADS, 256, glState);
+    private static class RenderTypeAccess extends RenderType {
```
yueh, line 148: `MolecularAssemblerRenderType` or something similar is probably more fitting name.

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
yueh: Moving the documentation about the workaround here is probably more fitting than hiding it in some method. When coming back to it later and seeing there is a custom `RenderType` and asking yourself why it even exists, it is better to have the documentation here.

yueh (changes requested):

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
shartte: It's just used to fake the protected access, don't start bikeshedding names ^^

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
shartte: Well no, it's fixing a bug around alpha testing in the vanilla render type

shartte (approved):

shartte:
Probably would just have AT'd it now to be honest, but eh _shrug_ ;-)

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
yueh: The related vanilla issue is marked as resolved, so this might be the intentional behaviour for minecraft. For whatever reason.
If we are not happy with the defaults and want a different rendering behaviour, then it is no longer a workaround. Just our custom solution.

shartte: The behavior between fabulous and fancy rendering is different in vanilla

yueh: Which may or may not be intentional. Regardless we want a different behaviour than the defaults provided by either. And it should be always the same regardless of the setting.

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
Technici4n: My bad, updated.

src/main/java/appeng/client/render/crafting/MolecularAssemblerRenderer.java
Technici4n: Is it that hard to read the documentation of the only method in this class (besides the constructor)? We have bigger fishes to fry imo.

shartte: Added the explanation back in, but the comment about why we use a subclass is fine. BTW forge does the same thing (ForgeRenderTypes class)

PR #5122, opened 2021-04-06
Cleanup cable bus after the last part is security broken

### LINKED ISSUE #5120: Cable busses broken by a security terminal leave behind an empty container

Technici4n:
Fixes #5120.


Technici4n:
Not sure this is the right place to break the cable bus itself. We should probably add to `IPart` that it should also cleanup the host if it's empty.

yueh:
Yes, this is the wrong approach. If a part has to check the state of the host and then tell it to do something, it's already broken by design.

There are also potentially other issues with the current approach, which will not be solved by this solution at all.

PR #5125, opened 2021-04-07
Add IPart#getPartHost()

### LINKED ISSUE #5120: Cable busses broken by a security terminal leave behind an empty container

Technici4n:
Use that to cleanup cable bus after the last part is security broken. Closes #5120.
Also replace AEBasePart#getHost() by calls to getPartHost().


Technici4n:
Too complex for no real benefit.

PR #5126, opened 2021-04-07
Make networked parts cleanup the part host in securityBreak()

### LINKED ISSUE #5120: Cable busses broken by a security terminal leave behind an empty container

Technici4n:
Closes #5120. Take 3. 😄


shartte (approved):

yueh:
This still has the issue of giving the wrong entity the responsibility for cleaning up a part host, does not fix the underlying issue, and further introduces an API break.

PR #5127, opened 2021-04-07
Fixes #5120: Always remove a cable bus once empty.

### LINKED ISSUE #5120: Cable busses broken by a security terminal leave behind an empty container

yueh:
The tile entity for a cable bus will now be removed once all parts are
removed through whatever means. This needs to be limited to only ones
existing in a world, otherwise it can accidentally remove cable buses
which are currently loaded. Further it should avoid cases where it is
replacing the last part as this will leave it shortly without a single
part.


Technici4n (approved):
Looks better than what I did.

PR #5131, opened 2021-04-09
Origin/1.16.x-fr-translation

AsgelTaren:
I improved some of the old wrong translations, and added the missing ones:

- Added translation for the fluid storage components and cells
- Added translation for the portable cells
- Corrected cables
- Corrected paint balls
- Improved some other less significant translation

(This is my first time doing a PR, so i may have made some mistake trying to create a PR correctly)


Technici4n (approved):
Should be fine now, thanks. Make sure to use the same ordering as the `en_us` file next time you add translations. :)

PR #5138, opened 2021-04-10
Crafting docs, also tests and rework because why not

Technici4n:
Anything `// FIXME CRAFTING` needs to be fixed before this is merged


yueh:
While the crafting code certainly needs better documentation and maybe better names, it is pretty much has not seen a bug in probably 5 or even more years. If they were always related to another system in AE2.

Once we move to a more generalized approach, this will pretty much need a whole rewrite from the ground up including preparations in terms of how we handle screens and container in terms of code, most likely a refactoring for networking as well as overall UI/UX and gameplay design. So the current code could probably be seen as deprecated.

All this might achieve is introducing new bugs into a perfectly working system. For which we then have to waste time on fixing them, maybe write tests for it and then throw everything away with 9.0.0 due to a fundamental design change.

So for now there won't be any changes accepted to the crafting system besides an actual bugfix, should it ever arise.

Technici4n:
> While the crafting code certainly needs better documentation and maybe better names, it is pretty much has not seen a bug in probably 5 or even more years. If they were always related to another system in AE2.

Sounds a lot like "I don't know what I'm doing but it works so let's not change it".

> Once we move to a more generalized approach, this will pretty much need a whole rewrite from the ground up including preparations in terms of how we handle screens and container in terms of code, most likely a refactoring for networking as well as overall UI/UX and gameplay design. So the current code could probably be seen as deprecated.

The whole point of this is to prepare for a generalization of the crafting code to other kinds of stacks such as fluids, but that implies understanding the current code in order not to have to fix bugs later, and properly test and document the current behavior.

> All this might achieve is introducing new bugs into a perfectly working system. For which we then have to waste time on fixing them, maybe write tests for it

Which you will have to do anyway when the crafting code is extended if that's not done properly.

> and then throw everything away with 9.0.0 due to a fundamental design change.

Nobody is going to do this because nobody is going to spend two months rewriting the crafting code from the ground up, what needs to be done is properly untie the crafting system from `IAEItemStack` for processing patterns, but again that requires understanding it first.

PR #5143, opened 2021-04-11
Controller Lights Texture Changes

AsgelTaren:
-Improvement on the controller light animation
-Improvement on the controller column light animation


yueh:
Many thanks. However we are currently not looking for another texture update.

PR #5148, opened 2021-04-12
Improve compatibility with mods messing up world/chunk loading event order

shartte:
Try to improve compatibility with other mods abusing the world load/unload events to do nasty stuff.

Fixes #5147

PR #5149, opened 2021-04-13
Remove world load event

### LINKED ISSUE #5147: [1.16.5] Crash on existing world
Hi, 

With the lastest version of AE2, the server crash with an NullPointerException on an existing world. 
No problem on new world.

Edit : Same issue on solo with the same world.

Pastbin : https://pastebin.com/wA5K0h4d

------------------------------------------------------------------------------------------------------------------------------------

Forge : 36.1.4
AE2 : 8.3.0-alpha.3

Technici4n:
Fixes #5147 by simply removing the world loading event. I believe it's cleaner than #5148.

### DIFF
--- a/src/main/java/appeng/hooks/ticking/ServerTileRepo.java
+++ b/src/main/java/appeng/hooks/ticking/ServerTileRepo.java
@@ -22,6 +22,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
+import javax.annotation.Nullable;
+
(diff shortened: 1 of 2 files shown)


yueh:
I would still prefer the event based approach. While it might be a bit more complex, it's easier to adjust for future changes needing to be performed at these times. 

Further it also can risk implicitly creating new world queues for unloaded worlds on accident, which might never be cleaned up or risking to leak incorrect tile entities into a newly loaded world.

Issue #5158, opened 2021-04-16
1.16.5 Dedicated server crash on autocrafting - Exception in server tick loop  java.util.ConcurrentModificationException: null at appeng.util.item.ItemList$ChainedIterator.next


ProsperCraft:
**Describe the bug**
A player on our server says he was autocrafting Mekanim things, and the server went down, he was the only player on at the time.

**To Reproduce**
Asking him for steps to reproduce.

**Additional context**
Crashlog-
https://gist.github.com/ProsperCraft/82ef431b0d11ec862898bf98e0bbaeea

**Environment**
appliedenergistics2-8.3.0-alpha.3
forge-1.16.5-36.1.2
Dedicated server.
[TownCraft TWO for 1.16.5 modpack ](https://www.curseforge.com/minecraft/modpacks/towncraft)

shartte:
Do you know what the player was crafting?

ProsperCraft:
Mekanism power cables, but when they did it again, it did not crash.

yueh:
performant might also be a cause here.

ProsperCraft:
It happened again, here is the crash report
https://gist.github.com/ProsperCraft/2bfca0458235159f2e9f4ed2230ce128
Here is the debug log-
https://gist.github.com/ProsperCraft/3478b7575ba6941a58b439533f5a5bd9

shartte:
@ProsperCraft We've just pushed alpha.4 that includes more debugging code to hopefully track the root cause of this down.
In the meantime, I presume the craft that caused this was the same again? Can you describe the pattern that the player is using to craft?

zpprsprk:
I just experienced a similar issue. I was running version alpha.4 and this is the part of the server log file when the crash happened.
https://gist.github.com/zpprsprk/57247faf16fe6affb473826e5d9f4fa3
I was auto crafting with a mekanism ultimate smelting factory. The pattern was a processing pattern for 64 exnihilo nickel ore chunks to 64 nickel ingots. It first happened when I requested 1001 ingots (it said it would do 1024). After restarting the server, I asked for 128 with no issues, then 1024 (exactly) with no issues, then again 1001 (that it said it would do 1024 again) and triggered the crash again.

yueh:
@zpprsprk Thanks, could reproduce it thanks to the description

PR #5159, opened 2021-04-17
Guard against reinjection of items when pushing items into the grid from CPUCluster

shartte:
Potential fix for #5158: Guard against re-insertion of items into the crafting CPU when dumping items into the network after a job was completed.

### DIFF
--- a/src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
+++ b/src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
@@ -415,7 +415,8 @@ public final class CraftingCPUCluster implements IAECluster, ICraftingCPU {
             errorMessage.append("Final output: ").append(finalOutput.copy().setStackSize(this.startItemCount));
             throw new RuntimeException(errorMessage.toString());
         }
-        this.waitingFor.resetStatus();
+        Preconditions.checkState(waitingFor.isEmpty(),
+                "When completing a job, the CPU should not be waiting for more items");
 
         this.remainingItemCount = 0;
         this.startItemCount = 0;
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -397,6 +400,23 @@ private void completeJob() {
...
+                errorMessage.append(" - ").append(stack).append("\n");
+            }
+            errorMessage.append("Current inventory:\n");
+            for (IAEItemStack stack : inventory.getItemList()) {
+                errorMessage.append(" - ").append(stack).append("\n");
+            }
+            errorMessage.append("Final output: ").append(finalOutput.copy().setStackSize(this.startItemCount));
+            throw new RuntimeException(errorMessage.toString());
+        }
+        this.waitingFor.resetStatus();
```
yueh, line 418: Wouldn't that be actually dead code/useless?

shartte: It's for "the future" when we actually remove the fail-fast code. I do want to clear that list :-P

yueh: It's still wrong. A crafting job can never be complete, when it's still waiting on more items. So it's always reseting an empty list, which is a bit pointless. Or it's actually an unfixed bug.

shartte: It still shows what the invariant is. We _can_ also replace this with Precondtion.checkState(waitingFor.isEmpty()) if you prefer that. I just don't  want this to be forgotten when we do remove the in-depth check .

PR #5194, opened 2021-05-05
Fix "Incomplete server tags, disconnecting" error

adde0109:
### What does it?
This small PR fixes "Incomplete server tags, disconnecting" when trying to join vanilla servers with AE2 loaded.  

### Introduction
When you try to join a vanilla server, the client checks if the server got all the tags that the client has registerd.
If there is a tag that the server does not have, the client aborts and disconnects you.
bild
Unless the tags are optional and in that case, ingnored.
bild
The forge ignores all the tags that have the instance of OptinalNamedTag and thus, makes it possible to join a vanilla server with mods loaded.

### The problem
AE2 does not register all it's tags as optional.
bild
This makes you unable to join a vanilla server with AE2 loaded.

### The solution
By changing the method used for adding tags into the tag registry from the non-optional one to the optional one, 
you can now sucessfully join a vanilla server without the need of removing AE2 from the mods dir.
bild

### Tested on a vanilla server and i singleplayer. 
Joining and playing on a vanilla server **works**.
Creating and playing in singel player **works**.
Facades can be crafted and works as expected.
ME chest and other interactive blocks works by opening their ui.
ME compass works by following it to a meteor.
Oregeneration works as expected.

I have not tested any older versions (1.13+) to see if they have the same problem, only 1.16.5 has been tested.

Feel free to look at the changes to see what had to be changed and change anything if needed.

_This is my first pr. If i missed something, i'm sorry_

### DIFF
--- a/src/main/java/appeng/core/features/registries/MovableTileRegistry.java
+++ b/src/main/java/appeng/core/features/registries/MovableTileRegistry.java
@@ -55,8 +55,8 @@ public class MovableTileRegistry implements IMovableRegistry {
     private final ITag.INamedTag<Block> blockTagBlackList;
 
     public MovableTileRegistry() {
-        this.blockTagWhiteList = BlockTags.makeWrapperTag(TAG_WHITELIST.toString());
-        this.blockTagBlackList = BlockTags.makeWrapperTag(TAG_BLACKLIST.toString());
+        this.blockTagWhiteList = BlockTags.createOptional(TAG_WHITELIST);
+        this.blockTagBlackList = BlockTags.createOptional(TAG_BLACKLIST);
     }
 
     @Override
--- a/src/main/java/appeng/fluids/parts/FluidAnnihilationPlanePart.java
+++ b/src/main/java/appeng/fluids/parts/FluidAnnihilationPlanePart.java
@@ -74,7 +74,7 @@ import appeng.util.Platform;
 public class FluidAnnihilationPlanePart extends BasicStatePart implements IGridTickable {
 
     public static final ITag.INamedTag<Fluid> TAG_BLACKLIST = FluidTags
-            .makeWrapperTag(AppEng.makeId("blacklisted/fluid_annihilation_plane").toString());
+            .createOptional(AppEng.makeId("blacklisted/fluid_annihilation_plane"));
 
     private static final PlaneModels MODELS = new PlaneModels("part/fluid_annihilation_plane",
             "part/fluid_annihilation_plane_on");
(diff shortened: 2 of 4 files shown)


shartte:
Uhm,  could you describe **why** you want to "trying to join vanilla servers with AE2 loaded."?

adde0109:
> Uhm, could you describe **why** you want to "trying to join vanilla servers with AE2 loaded."?

For an example, some servers network uses hubs that are in vanilla the you need to go through before you join the modded server.

The all of the other mods that are stable allows you join vanilla servers. 
AE2 shouldn't be one of the mods that keeps you from joining vanilla.

This has worked on AE2 pre 1.13, before forge changed to use datapack tags instead of oredictionary. Did this work 1.13-1.16? idk.

But it doesn't matter really why people want to join vanilla, its just that AE2 right now doesn't allow it. I see this as a bug. Its a malfunction that just exist. But not keeping you from playing it as intended.

Anyways:
When you connect to a vanilla server forge changes its behaviour to vanilla behaviour, but forge cant tell the difference between vanilla tags and AE2 tags becaus they are registered in such way. Because of that the vanilla game thinks something is wrong because the server is missing "vanilla tags" because they are present on the client. Forge dodges this vanilla checks by allowing mod developers to register thier tags as optional.

Its just not good practice to hide mod tags as vanilla tags, if they were registered as optional, forge would see them as not vanilla and thus, not use them when comparing tags to a vanilla server.

shartte:
We have never supported a mismatch between AE2 client/server, which this includes, sorry.
The Tags are **not** optional.

shartte:
@yueh Or do you have a different take from me on this?

adde0109:
> We have never supported a mismatch between AE2 client/server, which this includes, sorry.
> The Tags are **not** optional.

That was not was i was saying. I meant, when you join vanilla, no mods on the server at all, totally vanilla mode. Forge even says on the multiplayer screen: "Compatible vanilla server" before you connect. 

I'm 100% sure that you could join a vanilla server in 1.12 with AE2 loaded, test it yourself.

adde0109:
https://www.reddit.com/r/Minecraft/comments/ib3xsg/help_with_error_code/?utm_medium=android_app&utm_source=share

mindforger:
this is an approach on what i was attempting with bungeecord way in the past, an vanilla bungeecord server acting as a proxy server, where you could use commands or portals that would look to the client like moving to another dimension but actually establishing a connection to a subserver that is hosting a whole separate world ... forge did support in the past to join vanilla servers but the session chnge never worked due to the missing handshake during the fake dimension change (because it was non existent)

PR #5222, opened 2021-05-17
Downgraded JEI missing slots error to a warning.

yueh:
Now using the new JEI feature added in 7.7.0.98 to indicate a warning by tinting the `+` button yellow/orange instead of a hard fail.
Update to using the new methods using `ITextCompoment` instead of `String`
Fixed a wrong text entry for charged certus quartz when the worldgen is disabled.

Updated to JEI 7.7.0.98 as minimum requirement
Updated to Forge 36.1.10 as minimum requirement to fix it ignoring minimum requirements for non mandatory dependencies.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -230,7 +230,8 @@ processResources {
     exclude '.cache'
 
     filesMatching("META-INF/mods.toml") {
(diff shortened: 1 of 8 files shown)


shartte (approved):

yueh:
See #5224 for future reference

PR #5225, opened 2021-05-17
Fixes #5219: Fully extend facade to a full block.

### LINKED ISSUE #5219: cable facades are not watertight
I am not quite sure if this should be a feature request or a bug report.

**Describe the feature**
Facades should not 

Stairs and slabs only let water through non-solid sides

**Additional Context**
Screenshots from fabric/v8.3.0-alpha.2 (latest fabric)

yueh:
This also fixes other issues like getting stuck at a flat wall with
normal blocks mixed with facades.


src/main/java/appeng/facade/FacadePart.java
```diff
@@ -49,7 +49,7 @@ public ItemStack getItemStack() {
 
     @Override
     public void getBoxes(final IPartCollisionHelper ch, boolean itemEntity) {
-        if (itemEntity || ch.isBBCollision()) {
+        if (itemEntity /* || ch.isBBCollision() */) {
```
shartte, line 52: Don't comment out code like that :-|

shartte (changes requested):

src/main/java/appeng/facade/FacadePart.java
shartte, line 52:
```suggestion
        if (itemEntity) {
```

yueh: That is for reference as I have no clue what it does. In case we have to revert it in 3 months. I'll certainly remove it for a stable release.

shartte: you'll have forgotten about it long before we come to a stable release :-P

It's quite obvious _what_ it does. It doesn't tell us _why_ it does it though.

shartte (approved):

PR #5232, opened 2021-05-18
Fixes #785: Adds a part variant of the energy acceptor

### LINKED ISSUE #785: Energy acceptor as cable accessory
I was thinking, why does the energy acceptor have to be a full block? it seems to add unnecessary bulk to the network. I was hoping that perhaps we could get an Energy Acceptor type attachment to place on cables, much like what we have for the ME Interface.

yueh:
As Forge now moved to pretty much FE only we can add a part variant of
the energy acceptor without causing various problems by directly
exposing a particular energy API on every single cable.

This also means we most likely will no longer be able to support any non
capability based energy API should they arise as this will cause a
mismatch between what the block and part variants can handle.


src/main/resources/assets/appliedenergistics2/models/item/cable_energy_acceptor.json
```diff
@@ -0,0 +1,82 @@
+{
```
shartte, line 1: Is the model functionally different or can you inherit the block model ?

yueh: It has a different model. Similar to the part and block interface.

shartte (approved):

PR #5233, opened 2021-05-18
Cleanup

yueh:


src/api/java/appeng/api/util/AEColor.java
```diff
@@ -160,9 +160,9 @@ public int getVariantByTintIndex(int tintIndex) {
             case TINTINDEX_MEDIUM_BRIGHT:
                 final int light = this.whiteVariant;
                 final int dark = this.mediumVariant;
-                return (((((light >> 16) & 0xff) + ((dark >> 16) & 0xff)) / 2) << 16)
-                        | (((((light >> 8) & 0xff) + ((dark >> 8) & 0xff)) / 2) << 8)
-                        | ((((light) & 0xff) + ((dark) & 0xff)) / 2);
+                return ((light >> 16 & 0xff) + (dark >> 16 & 0xff)) / 2 << 16
```
shartte, line 163: i don't _really_ see this as an improvement

src/api/java/appeng/api/util/WorldCoord.java
```diff
@@ -148,7 +148,7 @@ public WorldCoord copy() {
 
     @Override
     public int hashCode() {
-        return (this.y << 24) ^ this.x ^ this.z;
+        return this.y << 24 ^ this.x ^ this.z;
```
shartte, line 151: this neither, should probably use Object.hashCode instead anyway

src/api/java/appeng/api/util/AEColor.java
yueh: 5 parenthesis really doesn't help here either, I would even say there are too many to really keep track when reading it, so it causes more confusion than it helps
The pattern of `value >> offset & MASK` is in my opinion common enough to extract certain bits from a variable to not need additional `()`.

src/api/java/appeng/api/util/WorldCoord.java
yueh: Yes, I already consider it here.

src/api/java/appeng/api/util/AEColor.java
shartte: Yes it helps, because no, i do have to look up bit shifting operator precedence regularly.

Technici4n: Same for me, I have no clue about the operator precedence of bitwise operations.

yueh: Neither solution is ideal. Personally I understand the clean one better, even when taking into account that I have to lookup the operator precendence. Mostly because I only have to look it up once and then apply it everywhere. While keeping track of the shitton of parenthesis is actually more challenging for me and is pretty much guaranteed to move one randomly by accident without noticing it.

Otherwise there might be a helper somewhere to calculate the average of two rgb ints?

Technici4n: I guess `int rAverage = (((light >> 16) & 0xff) + ((dark >> 16) & 0xff)) / 2;` and so on would work better if we want to clean up that statement.

src/api/java/appeng/api/util/WorldCoord.java
yueh: This is actually pointless to change. Or actually care about. The class is mostly an artifact when `BlockPos` did not exist, and is pretty much unused except for the sublass `DimensionalCoord`. Something we should take care of in 1.17 and just remove it completely.

PR #5248, opened 2021-05-24
Add missing getInput() calls in FE P2P tunnels

Technici4n:
I suspect this is an issue, not yet tested though. (I need to find a FE producer that doesn't auto-push, might have to write it myself xD).

I moved the `mavenCentral()` in the buildscript above the forge maven cause they were interfering for some reason, preventing spotless from running.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,9 +18,9 @@
 
 buildscript {
     repositories {
+        mavenCentral()
         maven { url = 'https://files.minecraftforge.net/maven' }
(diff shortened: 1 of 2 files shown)


yueh:
This is intentional to avoid crashes.

Technici4n:
Why is it not an empty energy storage then?

Technici4n:
And why is this not done for items and fluids p2p tunnels?

Technici4n:
Superseded by #5251.

PR #5251, opened 2021-05-25
Fix item P2P tunnels. Closes #5249.

### LINKED ISSUE #5249: P2P (items) eats up bulks
Item P2P eats up items, when having items go through Mekanism logistical transporters into item P2P it eats up almost everything, having 64 items go though i will end up with only 4 items going through and the rest no where to be found. But if i put manually items in Molecular assembler the items go through another P2P (item) without any issues 

(Having 16 items go through P2P only to and up crafting only 4 planks from 1 wood, rest of the 15 wood has vanished) 

image
image

Technici4n:
TL;DR working with `ItemStack`s is much more error prone than working with `int`s. 

Additionally stops p2p tunnel attunement destroying it when it's the only part in the block.

Additionally use `ActionResultType.CONSUME` instead of `ActionResultType.FAIL` when a part activation succeeded server-side.

Additionally:
- Item and fluids: don't crash if `getInput` is null, and properly tax pulled items/fluids.
- Energy: correctly allow pulling energy.

### DIFF
--- a/src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
+++ b/src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
@@ -82,6 +82,8 @@ public abstract class CapabilityP2PTunnelPart<P extends CapabilityP2PTunnelPart<
 
                 return adjacentCapability == null ? emptyHandler : adjacentCapability;
             } else {
+                // This capability is already in use (as the nesting is > 1), so we return an empty handler to prevent
+                // infinite recursion.
                 return emptyHandler;
             }
         }
@@ -110,11 +112,15 @@ public abstract class CapabilityP2PTunnelPart<P extends CapabilityP2PTunnelPart<
     protected void sendBlockUpdate() {
         if (!inBlockUpdate) {
             inBlockUpdate = true;
-            // getHost().notifyNeighbors() would queue a callback, but we want to do an update synchronously!
-            // (otherwise we can't detect infinite recursion, it would just queue updates endlessly)
-            TileEntity self = getTile();
-            self.getWorld().notifyNeighborsOfStateChange(self.getPos(), Blocks.AIR);
-            inBlockUpdate = false;
+
+            try {
+                // getHost().notifyNeighbors() would queue a callback, but we want to do an update synchronously!
+                // (otherwise we can't detect infinite recursion, it would just queue updates endlessly)
+                TileEntity self = getTile();
+                self.getWorld().notifyNeighborsOfStateChange(self.getPos(), Blocks.AIR);
+            } finally {
+                inBlockUpdate = false;
+            }
         }
     }
 
@@ -128,22 +134,25 @@ public abstract class CapabilityP2PTunnelPart<P extends CapabilityP2PTunnelPart<
         if (!inBlockUpdate) {
             inBlockUpdate = true;
 
-            if (isOutput()) {
-                P input = getInput();
+            try {
+                if (isOutput()) {
+                    P input = getInput();
 
-                if (input != null) {
-                    input.sendBlockUpdate();
-                }
-            } else {
-                try {
-                    for (P output : getOutputs()) {
-                        output.sendBlockUpdate();
+                    if (input != null) {
+                        input.sendBlockUpdate();
                     }
-                } catch (GridAccessException ignored) {
-                    // :-P
+                } else {
+                    try {
+                        for (P output : getOutputs()) {
+                            output.sendBlockUpdate();
+                        }
+                    } catch (GridAccessException ignored) {
+                        // :-P
+                    }
                 }
+            } finally {
+                inBlockUpdate = false;
             }
-            inBlockUpdate = false;
         }
     }
 }
--- a/src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
+++ b/src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
@@ -21,7 +21,7 @@ public abstract class CapabilityP2PTunnelPart<P extends CapabilityP2PTunnelPart<
     // Prevents recursive access to the adjacent capability in case P2P input/output faces touch
     private int accessDepth = 0;
     private final CapabilityGuard capabilityGuard = new CapabilityGuard();
-    private final EmptyCapabilityGuard emptyAdjCapability = new EmptyCapabilityGuard();
+    private final EmptyCapabilityGuard emptyCapabilityGuard = new EmptyCapabilityGuard();
     protected C inputHandler;
     protected C outputHandler;
     protected C emptyHandler;
@@ -64,7 +64,7 @@ public abstract class CapabilityP2PTunnelPart<P extends CapabilityP2PTunnelPart<
      */
     protected final CapabilityGuard getInputCapability() {
         P input = getInput();
-        return input == null ? emptyAdjCapability : input.getAdjacentCapability();
+        return input == null ? emptyCapabilityGuard : input.getAdjacentCapability();
     }
 
     protected class CapabilityGuard implements AutoCloseable {
--- a/src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
+++ b/src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
@@ -1,9 +1,14 @@
 package appeng.parts.p2p;
 
+import net.minecraft.block.Blocks;
 import net.minecraft.item.ItemStack;
 import net.minecraft.tileentity.TileEntity;
+import net.minecraft.util.math.BlockPos;
+import net.minecraft.world.IBlockReader;
 import net.minecraftforge.common.capabilities.Capability;
 import net.minecraftforge.common.util.LazyOptional;
+
+import appeng.me.GridAccessException;
 
 /**
  * Base class for simple capability-based p2p tunnels. Don't forget to set the 3 handlers in the constructor of the
@@ -11,9 +16,14 @@ import net.minecraftforge.common.util.LazyOptional;
  */
 public abstract class CapabilityP2PTunnelPart<P extends CapabilityP2PTunnelPart<P, C>, C> extends P2PTunnelPart<P> {
     private final Capability<C> capability;
+    // Prevents recursive block updates.
+    private boolean inBlockUpdate = false;
+    // Prevents recursive capability queries.
+    private int insertionNesting = 0;
+    private final InputCapability inputCapability = new InputCapability();
     protected C inputHandler;
     protected C outputHandler;
-    protected C nullHandler;
+    protected C emptyHandler;
 
     public CapabilityP2PTunnelPart(ItemStack is, Capability<C> capability) {
         super(is);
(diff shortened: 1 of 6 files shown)


Technici4n:
For a subsequent PR: LazyOptional invalidation and correct neighbor updates. That is already broken and will likely be done at the same time as the API to define arbitrary P2P tunnel types.

Technici4n:
You know what I'll take a look at that too tomorrow.

src/main/java/appeng/parts/p2p/FEP2PTunnelPart.java
```diff
@@ -172,9 +172,15 @@ public int getEnergyStored() {
     }
 
     private class OutputEnergyStorage implements IEnergyStorage {
+        private IEnergyStorage getInputStorage() {
```
yueh, line 175: This is something we have to remove or not add as it can break worlds. It's is mostly trivial to add it in 1.17 when we can make a breaking change and apply it to every tunnel and not have to have custom handling for each and every one.

src/main/java/appeng/parts/p2p/FEP2PTunnelPart.java
```diff
@@ -172,9 +172,15 @@ public int getEnergyStored() {
...
+            @Nullable
+            FEP2PTunnelPart input = getInput();
+            return input != null ? input.getAttachedEnergyStorage() : NULL_ENERGY_STORAGE;
+        }
+
         @Override
         public int extractEnergy(int maxExtract, boolean simulate) {
-            final int total = FEP2PTunnelPart.this.getAttachedEnergyStorage().extractEnergy(maxExtract, simulate);
+            final int total = getInputStorage().extractEnergy(maxExtract, simulate);
```
yueh, line 183: Should simply use the null implementation for now.

src/main/java/appeng/parts/p2p/FluidP2PTunnelPart.java
```diff
@@ -179,26 +164,31 @@ public FluidStack drain(int maxDrain, FluidAction action) {
     }
 
     private class OutputFluidHandler implements IFluidHandler {
+        private IFluidHandler getInputHandler() {
```
yueh, line 167: Same here

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
```diff
@@ -120,34 +122,37 @@ public ItemStack insertItem(int slot, @Nonnull ItemStack stack, boolean simulate
...
+                    ItemP2PTunnelPart.this.queueTunnelDrain(PowerUnits.RF, amount - remainder);
                 }
             } catch (GridAccessException ignored) {
             }
 
-            return remainder;
+            if (remainder == stack.getCount()) {
+                return stack;
+            } else {
+                ItemStack copy = stack.copy();
```
yueh, line 152: A special case for an empty itemstack might be an idea. Copying is expensive, and setting a normal itemstack to 0 also maintains the item id and so on and can have interesting side effects.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
```diff
@@ -169,16 +174,21 @@ public boolean isItemValid(int slot, @Nonnull ItemStack stack) {
     }
 
     private class OutputItemHandler implements IItemHandler {
+        private IItemHandler getInputHandler() {
```
yueh, line 177: Again, null impl.

yueh (dismissed):

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: Good point!

Technici4n:
So I've updated this by properly sending block updates, and I've also used RAII guards to prevent recursive transfer while ensuring p2p tunnels can be pushed into and pulled from.

Common capability logic has been refactored into a base class, we can look into making it pluggable for 1.17 later on, that shouldn't be too hard.

src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
```diff
@@ -0,0 +1,149 @@
...
+
+                    if (te != null) {
+                        adjacentCapability = te.getCapability(capability, getSide().getOpposite().getFacing())
+                                .orElse(null);
+                    }
+                }
+
+                return adjacentCapability == null ? emptyHandler : adjacentCapability;
+            } else {
+                return emptyHandler;
```
shartte, line 85: This needs to be documented _why_ we do this (avoid recursion)

src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
```diff
@@ -0,0 +1,149 @@
...
+            }
+        }
+    }
+
+    // Override when there is no capability to retrieve.
```
shartte, line 90: Override ?

src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
```diff
@@ -0,0 +1,149 @@
...
+    protected class EmptyAdjCapability extends AdjCapability implements AutoCloseable {
+        @Override
+        public void close() {
```
shartte, line 93: Why does this override close? Isn't the depth already incremented when this object gets constructed hence it leaks the ref?

shartte: *edit2*: Overall I don't think I like the auto closable use here. In general I'd like it, but it's a bit obtuse with the empty handler in the mix.

src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
```diff
@@ -0,0 +1,149 @@
...
+        return input == null ? emptyAdjCapability : input.getAdjacentCapability();
+    }
+
+    // Send a block update on p2p status change, or any update on another endpoint.
+    // Prevent recursive block updates.
+
+    protected void sendBlockUpdate() {
+        if (!inBlockUpdate) {
+            inBlockUpdate = true;
+            // getHost().notifyNeighbors() would queue a callback, but we want to do an update synchronously!
```
shartte, line 113: Probably should be try {}finally{}

src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
```diff
@@ -0,0 +1,149 @@
...
+
+    @Override
+    public void onTunnelNetworkChange() {
+        sendBlockUpdate();
+    }
+
+    @Override
+    public void onNeighborChanged(IBlockReader w, BlockPos pos, BlockPos neighbor) {
+        if (!inBlockUpdate) {
+            inBlockUpdate = true;
```
shartte, line 129: another one for try {} finally{}

src/main/java/appeng/parts/p2p/CapabilityP2PTunnelPart.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    @Override
+    public void onTunnelNetworkChange() {
+        sendBlockUpdate();
+    }
+
+    @Override
+    public void onNeighborChanged(IBlockReader w, BlockPos pos, BlockPos neighbor) {
+        if (!inBlockUpdate) {
```
shartte, line 134:
```suggestion
        // Ensure that we do not enter an infinite loop if the input/output side of a P2P tunnel touch
        if (!inBlockUpdate) {
```

shartte:
LGTM. The alternative to using the AutoClosable is equally bad imho (try/finally mess at every location we use the caps, or alternatively, lambdas).

shartte (dismissed):

Technici4n:
Still working after the rename, just waiting for yueh's review!

yueh:
I still do not like the `AutoCloseable` approach. This is one of these cunning solutions, which ultimately does not fix the problem at all as it offloads the effort to everyone (and more) and we have absolutely no control about them not messing up. Due to having to wrap every single method in a `try-with-resources` block, otherwise it will not work.

Would java support something like classwide decorators like python (in a sane way, so no mixins or similar), this would be easier as we can ensure every method is wrapped. But even that will most likely break as we decorate too many and immediately trigger it.

Further I cannot even remember the last time I had to interact with resources at this level. The only fact I remember was that it was around Java6, so there was no `try-with-resources` at all. Usually these things are hidden behind some libs because it is one of these things, which are guaranteed to be used incorrectly. Even worse it this case we basically offload it to every other dev to basically write a library to avoid these kind of problems. Therefore being even more difficult to anyone wanting to implement their own variant.

Lambdas might be a better solution, but only if we can ensure people have to actually pay attention and read their docs or similar.
I guess it might be acceptable for 8.4, but certainly nothing which should be considered for 9.0 at this time and would need to be reverted then.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
```diff
@@ -104,7 +70,7 @@ public ItemStack getStackInSlot(int slot) {
         @Override
         @Nonnull
         public ItemStack insertItem(int slot, @Nonnull ItemStack stack, boolean simulate) {
-            final ItemStack remainder = stack.copy();
+            int remainder = stack.getCount();
```
yueh, line 73: Shouldn't we potentially use a fast return here for an empty stack? Technically we have one a couple lines down, but that is a bit hidden behind other conditions. Might also apply to anything with the concept of `empty`.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
```diff
@@ -118,36 +84,43 @@ public ItemStack insertItem(int slot, @Nonnull ItemStack stack, boolean simulate
...
-                    ItemStack received = ItemStack.EMPTY;
+                    try (CapabilityGuard capabilityGuard = target.getAdjacentCapability()) {
+                        final IItemHandler output = capabilityGuard.get();
+                        final int toSend = amountPerOutput + overflow;
 
-                    for (int i = 0; i < output.getSlots(); i++) {
-                        received = output.insertItem(i, fillWithItemStack, simulate);
-
-                        if (received.isEmpty()) {
+                        if (toSend <= 0) {
```
yueh, line 91: If `toSend` is actually negative here, would that not indicate an error and potentially shouldn't be swallowed?

Technici4n:
Well anyone looking into implementing a custom p2p tunnel will have a look at how we do it and see that we use the the guard with a try-with-resources, so I'm not buying the "people will not use it correctly" argument.

Lambdas hurt debugging and generate quite some allocation. It may not be a huge deal, but I don't see it as a particularly better option.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
```diff
@@ -190,17 +166,29 @@ public ItemStack insertItem(int slot, @Nonnull ItemStack stack, boolean simulate
...
+
+                return result;
+            }
         }
 
         @Override
         public int getSlotLimit(int slot) {
-            return ItemP2PTunnelPart.this.getInput().getAttachedItemHandler().getSlotLimit(slot);
+            try (CapabilityGuard input = getInputCapability()) {
+                return input.get().getSlotLimit(slot);
```
yueh, line 183: Are there any cases (in other mods), in which this could be called during an extract/insert? E.g. to split the request across multiple linked inventories while taking potential limits into account while using it to keep a ratio based on the slot limits?
If so how would this behave?

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
```diff
@@ -118,36 +84,43 @@ public ItemStack insertItem(int slot, @Nonnull ItemStack stack, boolean simulate
...
+                    ItemP2PTunnelPart.this.queueTunnelDrain(PowerUnits.RF, amount - remainder);
                 }
             } catch (GridAccessException ignored) {
             }
 
-            return remainder;
+            if (remainder == stack.getCount()) {
+                return stack;
+            } else if (remainder == 0) {
+                return ItemStack.EMPTY;
```
yueh, line 118: Cannot look it up currently, but I remember the helpers provided by forge usually returning `ItemStack.EMPTY` in case of it being empty and not the original stack set to a stacksize of 0. Not sure if this is defined somewhere as contract. This applies mostly in regards of the first condition, but currently this should never be 0. See the previous comment about a fast return.

Also similar to to capturing a negative `toSend`, what happens in case something returns negative amounts and the remainder is actually greater than `stack.getCount()`?

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: We can, but I assume it would be done by the caller already.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: That's an error in another mod, yes. I'm not sure how hard we should try to catch it? We can print the handler itself, the adjacent tile, the adjacent blockstate, etc... Not sure if that's worth it?

Technici4n: We can check that `0 <= sent <= toSend`, what do you think?

shartte:
@yueh Regarding AutoClosable: I think this fixes the underlying issue for now, and should be the final fix for 1.16.
In 1.17, we can revisit this, as you noted, we'd want to move away from having to subclass P2PTunnelPart anyway, which gives us more options in my opinion, and then we can revisit lambdas if we want.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: Forge's helpers don't check this, so it will already fail inside `ItemHandlerHelper#insertItem`. I don't think we should spend too much energy on that.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: I don't think this is useful.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: It's no use, it will fail in `ItemHandlerHelper#insertItem`.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: Nobody checks these methods, so I don't think it makes sense for us to check them - the forge helper will fail silently, we might not even be able to detect that.

src/main/java/appeng/parts/p2p/ItemP2PTunnelPart.java
Technici4n: I'm not sure, worst case we can bump the depth limit to 2 or 3. I suggest we do later it if that proves necessary in the future?

Technici4n:
@yueh Are you fine with this PR in its current state? We can revisit it in 1.17 if necessary. For now I'd say the priority is getting a stable 8.4.0 build so that we can work on 1.17 stuff.

shartte (approved):

PR #5265, opened 2021-05-30
Change CraftConfirm Cancel button to go back to Craft Amount

### LINKED ISSUE #5243: Add a back button to the crafting result
There should be a back button on the crafting result to go back to selecting the amount.

This would provide a convenient way to correct errors by clicking to fast or change the amount in case there are not enough materials at hand.

shartte:
Fixes #5243, and adds infrastructure for further improvements to send client->server GUI commands.

### DIFF
--- a/src/main/java/appeng/client/gui/me/crafting/CraftAmountScreen.java
+++ b/src/main/java/appeng/client/gui/me/crafting/CraftAmountScreen.java
@@ -41,6 +41,8 @@ public class CraftAmountScreen extends AEBaseScreen<CraftAmountContainer> {
 
     private final NumberEntryWidget amountToCraft;
 
+    private boolean initialAmountInitialized;
+
     public CraftAmountScreen(CraftAmountContainer container, PlayerInventory playerInventory, ITextComponent title,
             ScreenStyle style) {
         super(container, playerInventory, title, style);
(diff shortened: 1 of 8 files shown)


yueh:
Just in case, the idea was to offer it as alternative button.

So we can handle these cases (not just with this button)
* Completely close it, so also no back to the terminal (exists with `ESC` or similar)
* Go back to the terminal. Quick check if there are enough mats, some missing pattern, etc.
* Go back to the amount to update it when wanting to change it quickly.

I have not tested it yet, but based on the description, it sounds more like it replace the current behaviour with a new one.

shartte:
@yueh It does, I don't want another button on that screen, really. 
While possible, we cannot reuse the goback functionality be cause it'll just go back to the terminal.
In general I'd rather see this as a popup-dialog which - when canceled - goes back to the previous dialog.

yueh:
I think we should push this to 9.0 then. Currently I cannot really tell which option is more useful, so this might be more annoying for players being used to the current behaviour.

For 9.0 we could consider some more breaking changes like also making use of a modifier like shift, which we already do for the amount screen to immediately craft when possible without showing the plan. So we could use the back button to go back to the amount normally and to the terminal when holding shift.

Another alternative could be to design both screens to have the buttons located at the same position, so players can still go back pretty quickly by tapping the mouse twice without having to move it around.

shartte:
> Currently I cannot really tell which option is more useful

How does targeting 9.0 change that?

> so this might be more annoying for players being used to the current behaviour.

By this logic, we can never change anything anymore.

yueh:
It is not about the change itself, which certainly is a good one. Just the time of introducing it. 9.0 will give is more option to break things and maybe even document them, should we ever come around making a patchouli book.

However at the current point of 8.x it feels like a pretty random change, which could easily be seen as bug and not feature.
Same with hiding it behind shift and then potentially flipping it for 9.0 as well as probably never documenting it.

Actually, can we attach a tooltip to a button currently? If so we could roll it into a tooltip to indicate that the button will go back to the crafting amount normally or back to the terminal when holding shift. At the same time probably also add one to the craft button to indicate that shift clicking will skip the confirmation and just start it directly with enough materials (and no feedback on failure)

shartte:
It's a Cancel Button. Not rocket science.

yueh (approved):

Issue #5266, opened 2021-05-31
Wrench Capability


PssbleTrngle:
**Describe the feature**
Instead of making the wrench an implementable interface, make it a capability.

**Reasons why it should be considered**
There is no way for mod to provide their wrench without making AE2 a hard dependency

Technici4n:
From Creators-of-Create/Create#1691.

I would like to move to a system where the `forge:wrenches` tag is used for "simple" wrenches, i.e. those that don't cost any durability but just tell the block to have wrench behavior - think Thermal's Crescent Hammer, AE2's own wrenches, Create's as well, etc... I don't really care about not damaging wrenches that should be damaged after usage - I guess players would use non-damageable wrenches anyway.

PssbleTrngle:
I do think allowing items with the tag `forge:tools/wrenches` would suffice for almost all cases. Most of the wrenches provides by other mods do allow usage independent from any conditions like the player or itemstack and most of them don't even have any durability, so implementing an interface just to return `true` is kind of an overkill for most of them

PssbleTrngle:
Should I include checking if the itemstack has the `forge:tools/wrenches` in this pr?

yueh:
This is not as easy as adding a cap or tag to support other wrenches. Currently some logic is still handled by our own wrenches or would have to be handled by them to fix one or two smaller bugs.

As far as I can remember the wrenches handle rotating or disassembling blocks on their side, while only parts handle their own removal when being activated by a wrench.

We would need to completely rewrite the wrench handling to support any item being used. The usual example of using a stick for debugging. Probably removing `IAEWrench` completely due to being useless then.
Otherwise it would behave pretty randomly when using a 3rd party wrench instead of an AE2 wrench.

yueh:
Closing in favour of a tag based approach in 1.17

PR #5267, opened 2021-05-31
Capability Wrench

PssbleTrngle:
Implementation for #5266

Make the `IAEWrench` interface an item capability, allowing mods to implement it without requiring AE2 as a hard dependency.
Implementing the interface still works, but you can now alternativly provide it as a capability.

### DIFF
--- a/src/main/java/appeng/capabilities/Capabilities.java
+++ b/src/main/java/appeng/capabilities/Capabilities.java
@@ -18,7 +18,6 @@
 
 package appeng.capabilities;
 
-import appeng.api.implementations.items.IAEWrench;
 import net.minecraft.nbt.INBT;
 import net.minecraft.util.Direction;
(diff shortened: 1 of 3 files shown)


Technici4n:
I don't think this works for everything that an AE2 wrench does, as even the AE2 wrench does not correctly wrench some things (somewhat of a bug, but it requires someone to take a deeper look).

And I would prefer a tag, so that mods don't have to implement 10 capabilities for their wrenches to work.

Technici4n:
I don't think the capability is necessary - only the tag should be enough.

But sadly this will not work in 100% of the cases, as some wrench logic still needs to be refactored into the item use event handler.

In all cases, you should wait for a comment from another dev before changing your PR in case they disagree with me.

yueh:
Will be moved to 1.17 with it being nearly released and we can actually make API breaks without having to add various workarounds to keep 1.16 stable.

PR #5280, opened 2021-06-06
Revert unwanted changes from "Make annihilation planes fallback to a pickaxe. Close #5275"

### LINKED ISSUE #5275: Annihilation Plates voiding blocks
**Describe the bug**
I have this setup:
image
When I place a block in front of the plane, it gets broken, as expected.
However, it doesn't enter the Cell, it just gets voided. The Cell isn't partitioned or anything, it's a normal, blank Cell.

https://user-images.githubusercontent.com/15698722/120915593-1a990100-c66a-11eb-98fd-d576ae8727f0.mp4

When I remove the Drive entirely, it still breaks blocks (and voids them):

https://user-images.githubusercontent.com/15698722/120915618-4a480900-c66a-11eb-9403-7fa349bbcf50.mp4

**Environment**
MC 1.16.5, Forge version 36.1.4, AE2 version `appliedenergistics2-8.3.1`
Force crash report: https://pastebin.com/Qz8TktuL
[latest.log](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/6603763/latest.log)
[debug.log](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/6603764/debug.log)

yueh:
This is potentially broken as we cannot guess that a pickaxe is the correct tool when the method added by forge returns `null`.
By definition it means anything could break it, which might include a pickaxe. However this can still be an incorrect tool or result in a non expected outcome. 

Think of how cobwebs will drop strings when broken with a sword. 

Not that a sword is a tool, or forge would even provide a `ToolType` for it. Or a way to actually test every tool as this is not a public API. part.

### DIFF
--- a/src/main/java/appeng/parts/automation/AnnihilationPlanePart.java
+++ b/src/main/java/appeng/parts/automation/AnnihilationPlanePart.java
@@ -113,8 +113,7 @@ public class AnnihilationPlanePart extends BasicStatePart implements IGridTickab
 
         // For collision, we're using a simplified bounding box
         if (bch.isBBCollision()) {
-            // The smaller collision hitbox here is needed to allow for the entity collision
-            // event
+            // The smaller collision hitbox here is needed to allow for the entity collision event
             bch.addBox(0, 0, 14, 16, 16, 15.5);
             return;
         }
@@ -333,9 +332,8 @@ public class AnnihilationPlanePart extends BasicStatePart implements IGridTickab
 
         for (ItemStack item : items) {
             IAEItemStack overflow = storeItemStack(item);
-            // If inserting the item fully was not possible, drop it as an item entity
-            // instead
-            // if the storage clears up, we'll pick it up that way
+            // If inserting the item fully was not possible, drop it as an item entity instead if the storage clears up,
+            // we'll pick it up that way
             if (overflow != null) {
                 Platform.spawnDrops(w, pos, Collections.singletonList(overflow.createItemStack()));
             }
@@ -379,28 +377,24 @@ public class AnnihilationPlanePart extends BasicStatePart implements IGridTickab
         final Material material = state.getMaterial();
         // Note: bedrock, portals, and other unbreakable blocks have a hardness < 0, hence the >= 0 check below.
         final float hardness = state.getBlockHardness(w, pos);
-        final boolean ignoreMaterials = material == Material.AIR || material == Material.LAVA
-                || material == Material.WATER || material.isLiquid();
+        final boolean ignoreAirAndFluids = material == Material.AIR || material.isLiquid();
 
-        return !ignoreMaterials && hardness >= 0f && w.isBlockLoaded(pos)
+        return !ignoreAirAndFluids && hardness >= 0f && w.isBlockLoaded(pos)
                 && w.isBlockModifiable(Platform.getPlayer(w), pos);
     }
 
     protected List<ItemStack> obtainBlockDrops(final ServerWorld w, final BlockPos pos) {
-
-        Entity fakePlayer = FakePlayerFactory.getMinecraft(w);
-
+        final Entity fakePlayer = FakePlayerFactory.getMinecraft(w);
         final BlockState state = w.getBlockState(pos);
+        final TileEntity te = w.getTileEntity(pos);
 
         ItemStack harvestTool = createHarvestTool(state);
 
-        if (!state.getRequiresTool() && !harvestTool.isEnchanted()) {
-            // If the state does not require a tool, and the tool is not enchanted (with silk touch for example in case
-            // of the identity annihilation plane), then we don't use a tool.
+        if (!state.getRequiresTool() && state.getHarvestTool() == null && !harvestTool.isEnchanted()) {
+            // Do not use a tool when not required, no hints about it and not enchanted in cases like silk touch.
             harvestTool = ItemStack.EMPTY;
         }
 
-        TileEntity te = w.getTileEntity(pos);
         return Block.getDrops(state, w, pos, te, fakePlayer, harvestTool);
     }
 
@@ -457,9 +451,8 @@ public class AnnihilationPlanePart extends BasicStatePart implements IGridTickab
             return false;
         }
 
-        // This handles items that do not spawn via loot-tables but rather normal block
-        // breaking
-        // i.e. our cable-buses do this (bad practice, really)
+        // This handles items that do not spawn via loot-tables but rather normal block breaking i.e. our cable-buses do
+        // this (bad practice, really)
         final AxisAlignedBB box = new AxisAlignedBB(pos).grow(0.2);
         for (final Object ei : w.getEntitiesWithinAABB(ItemEntity.class, box)) {
             if (ei instanceof ItemEntity) {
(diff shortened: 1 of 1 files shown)


shartte:
Can you provide an actual example?

yueh:
I won't dig through every mod out there to find one using a more convoluted behaviour. E.g. drop different things depending on someone using an empty hand, a pickaxe, or an axe.

Simply following the contract forge gives us and that is 
```java
    /**
     * Queries the class of tool required to harvest this block, if null is returned
     * we assume that anything can harvest this block.
     */
    ToolType getHarvestTool(BlockState state);
```

So if forge returns `null` for some (vanilla) blocks actually requiring a tool, they break their own contract. I guess nobody of the forge team every played minecraft after 1.10 or 1.12 and simply never encountered new blocks.

There are certainly some mods with really broken behaviour out there. Notably pedestals, which cannot be broken in creative with a pickaxe. So who knows what else might exist. And once a single mod exist, which makes use of it, we break it for a workaround as forge is incapable of fixing their own stuff.

I also remember looking at things like waila or top on how they determined the correct tool and it was just a huge hack.
The approach was basically break a virtual block of the same type with every tool and then use the block break progress and once the progress is over a certain threshold it will consider it as valid tool and display it. I think the progress was also only clientside. Who knows what also happens when someone uses a similar approach to use with sky stone and just makes it slow enough to fall below the threshold.

shartte:
The reality is that the differentiation is split between: 
- Can it be broken at all? That is determined by hardness and we do check for it.
- Does it drop loot? And this is determined by "does it require a tool? do we have the right tool?"

I don't see why this part should not just act as the "right" tool. The examples you give that give different loot based on the tool used still work as-is given they now use loot-table predicates, and we do **not** circumvent them. The only check we circumvent is whether loot will be produced at all.

yueh:
In some way it certainly goes into a philosophical question. There are also other helper methods to test a block against a `ToolType`. But besides the usual pickaxe, axe, hoe, and shovel nothing is really public or provide a way to get a matching itemstack.

All we simply do here is making the assumption that everything can be broken with a pickaxe and then behaves like a player would expect, which is pretty much impossible to verify realistically. Just to provide a workaround for a Forge bug. And I certainly do not like the "works for me. Let's hope nobody uses it with other mods" approach. Once that is no longer true, we have to pick between sticking to the contract or using a workaround for a bug and hope that it still works as expected.

We also don't even consider `hoe` as valid tool, and default to an empty hand. Could also cause problems.

Further there are also shears, which are a tool without a tooltype. If I recall correctly, there are some blocks, which need them for drops and they certainly don't fall into the silk touch category. No idea how to handle them.

shartte:
Again, why should this part not simply act as the right tool at that point in the decision tree that decides whether the loot table is evaluated, or not?
I don't see how an annihilation plane is inherently more a pickaxe than it is a sword.

yueh:
I cannot really tell it. All I can say is that we encounter a bug in forge and decide that we ignore their own contract, make an assumption based on concrete and hope it will still works otherwise.

shartte:
We're not making assumptions based on concrete.
The code is relatively simple to follow, honestly.
We have to decide whether the annihilation plane is only supposed to act as a finite set of tools, or whether it should generally try to harvest blocks as best as it can. I am tending to go with the second option, honestly.

Technici4n:
Can you post an example of a case where the new behavior is worse than the old behavior, and explain why that is so important that we should revert the commit allowing annihilation planes to work correctly in most cases?

I don't care whether we are working around a Forge bug, or just vanilla hardcoding... I want our players to be able to use annihilation planes, period. Concrete is actually a very valid use case for people trying to setup its autocrafting.

yueh:
And the original only covers the first case of making an assumption that missing a tool means it wants a pickaxe.

In no way it tries to find the best possible tool. E.g. instead of defaulting to a pickaxe use the helpers to test a tooltype. So it still has a chance to detect it actually needs an axe.

All it now does is hardcode a couple cases, but not really fix in general.

shartte:
That is not how the harvest/drop logic works.

As I tried to explain there's essentially a flow chart.

It has the concept of *harvestability*. Which is used to determine if **any** drops will be produced or not. If harvestability is determined to be false, then evaluation of the loot table is skipped entirely.

The Forge "tool type", and the vanilla "requiresTool" and "canHarvestBlock" functions only relate to harvestability. They NEVER change, **what** drops. Only if **anything** drops.

Beyond this concept of harvestability, a loot table can apply whatever predicates it wants to apply to change what drops based on the used tool. This is entirely unrelated to harvestability/tool types.

**Cobweb Example**
*Harvestability logic*: `setRequiresTool`, no forge tool type set
In Vanilla `SwordItem` and `ShearsItem` return true for `canHarvestBlock(COBWEB)`.

This means that the loot-table will only be evaluated if a SwordItem or ShearsItem is used to break the block. (Disregarding any modded tools that also return `true` for canHarvestBlock(COBWEB))

*Loot table*: 
The loot table either drops a string (without any condition), or it drops the cobweb itself.
The cobweb is dropped via an OR-condition of SILKTOUCH or a direct match for shears.
Interestingly I don't think swords can be enchanted with silk touch, meaning the silk touch check would only apply to modded tools that are silk-touch enchanted and return true in the harvestability-stage.

yueh:
Which also raises further question as axes never override `canHarvestBlock()` and will return false for everything. Only shovels, pickaxes, shears and swords override it. However we use axes, when hinted. What happens in this case when a block is set to require a tool and it returns axe as type? In this case it shouldn't drop anything, except they have their own custom axe overriding it. Which we certainly cannot use.

There was also the idea to just ignore the default tools at all and just use a generic one, which will always return `true`. E.g. reuse a debug item for now or something similar. So we won't depend on some various edge cases regarding incorrect tiers and similar.

shartte:
As I said, harvestability is now completely ignored by annihilation planes. We're by definition always using the right tool.

For differentiating between tools in loot-tables (i.e. the cobweb example), canHarvestBlock is not used.

yueh:
There are actually a couple other problems or missunderstandings here.

The first one being `canHarvestBlock()` is completely irrelevant here. It was never used and technically we never tried to reimplement it.

Secondly the comment about fallback pickaxes is at best misleading and at worst just plain wrong and therefore confusing.
It could always return an `ItemStack.EMPTY` and still harvest pretty much all blocks. Except for a couple ones which take enchantments into account. In these cases a stick as tool is enough to work. Basically the used tool is completely irrelevant as long as you can enchant it. And ignoring whatever some mod makes with their loot tables and require their own custom tools, that's something we simply cannot take into account. Same with the forge hook to completely replace the loot table.

Further I am not really happy about using netherite tools, these could affect loot tables in a positive way without needing them as materials. I would say this could negatively influence mod packs, e.g. if they attach additional drops when using netherite tools for whatever reasons. This should clearly be capped at diamond or maybe iron. Not that this would have any effect as we completely ignore the harvest level.

Lastly the actual bug is never ever mentioned. To me it looks like it was plain luck that it appears to work.
But would be certainly helpful to document them in some way. It would certainly have helped to solve my confusion and not getting sidetracked into `canHarvestBlock()`, etc.

So still tending to revert it as overall it is still incomplete in terms of using the best approach as far as possible with forge.
E.g. it's still not handling hoes as tooltype.
As well as incorrect in terms of documentation, which will only hurt in the long run. Especially without ever mentioning the actual bug somewhere within a commit message or similar.

Technici4n:
> Secondly the comment about fallback pickaxes is at best misleading and at worst just plain wrong and therefore confusing.
It could always return an ItemStack.EMPTY and still harvest pretty much all blocks. Except for a couple ones which take enchantments into account. In these cases a stick as tool is enough to work. Basically the used tool is completely irrelevant as long as you can enchant it. And ignoring whatever some mod makes with their loot tables and require their own custom tools, that's something we simply cannot take into account. Same with the forge hook to completely replace the loot table.

So you are saying that pickaxes are totally fine as a fallback, which I agree with. Indeed, the comment is misleading, so let's just update the comment instead of reverting the entire commit.

> Further I am not really happy about using netherite tools, these could affect loot tables in a positive way without needing them as materials. I would say this could negatively influence mod packs, e.g. if they attach additional drops when using netherite tools for whatever reasons. This should clearly be capped at diamond or maybe iron. Not that this would have any effect as we completely ignore the harvest level.

Who cares, seriously?

> Lastly the actual bug is never ever mentioned. To me it looks like it was plain luck that it appears to work.

The actual bug was the `return Collections.emptyList()`, however silk touch would not work in all cases due to the null return in `getHarvestTool` which is why I added the pickaxe as a fallback. We can update the comment in `getHarvestTool` to reflect this.

> E.g. it's still not handling hoes as tooltype.

This commit fixes an actual bug, and I don't care about hoes. If you want to PR a better fix, or add hoe support, feel free to do so without reverting the bugfix, please.

shartte:
> There are actually a couple other problems or missunderstandings here.

You are missing that never calling canHarvestBlock if tooltype==null was the source of the previous bug :-P

The argument against netherite can be made, but the same argument could be made against diamond pickaxes too. I don't care too much either way as I stated before since passing an empty itemstack would usually be sufficient too.

yueh:
> You are missing that never calling canHarvestBlock if tooltype==null was the source of the previous bug :-P

Because the bug was never documented and comments point to something to something completely irrelevant as cause.

> The argument against netherite can be made, but the same argument could be made against diamond pickaxes too. I don't care too much either way as I stated before since passing an empty itemstack would usually be sufficient too.

Most other block breakers seems to use diamond tools. It simply has no negative effect to stick with diamond or iron on our side. We technically do not need diamonds to craft them, but gold/redstone and nether resources. Which indirectly needs a diamond pick for the portal or carrying a bunch of lava buckets around and hoping to not make a single mistake. So I guess mostly fair in terms of balancing. However it makes it impossible to reward manual mining in some mod packs (for whatever reason). Here we simply bypass it completely and kill their balancing.  So not really relevant for us, but we reduce options for other mods or mod packs in terms of balancing, which in my opinion should be something we should avoid as long as it has no negative effect for us.

Also `ItemStack.EMPTY` will not work in cases like ores. The loot tables here rely on the item being enchantable for fortune and as `ItemStack.EMPTY` cannot be enchanted, it will drop nothing at all. Even more relevant for 1.17 as every ore will now drop items and no longer the block itself. Not just limited to things like diamods or redstone, also iron/gold/copper.

Technici4n (approved):
Looks good now. 👍

PR #5297, opened 2021-06-15
Reduce ItemStack creations in ItemHandlerAdapter#extract

### LINKED ISSUE #5296: AE2 import/export massive amount of items
**Describe the feature**
At the current moment. with mods like Environmental tech and other ''get items fast'' mods. were processing 100-1000 items a second is not uncommon AE2 seems to fall behind in the performance section. AE2 gets the job done faster and effective than other pipe or storage mods. but at the cost of MASSIVE TPS lag. i always had the vision of AE" being the more server friendly autocrafting storage mod. but with these types of mods coming out. and being more commonly requested in modpacks. something needs to change related to how the code handles items in and out of the system every tick. 

This also is seen when shift clicking crafting items from a terminal. it will dead lock a server for a few seconds.

Generally i see modded being a more community type gameplay thing. so server performance is a must. which AE2 have done well so far. i just saddens me that i have to choose something like refined storage over my favorite mod AE2 and have ''crappy looking systems'' just to have a better running server. i hope this is something that you guys already is working on since it seems to be a growing problem.

Technici4n:
Close #5296.

This PR does that in two ways:
- Do not use a new stack to compare in `extractItems`. We can compare with the IAEItemStack directly (internally it uses the shared item stack for the comparison).
- ~~Cache the stack used by `insertItems`. This is especially useful when there are a lot of storage busses to iterate through for an injection. Most of those return the stack unmodified, and we can just reuse it for the next time, possibly adjusting the count.~~ Somewhat risky, reverted for now.

### DIFF
--- a/src/main/java/appeng/parts/misc/ItemHandlerAdapter.java
+++ b/src/main/java/appeng/parts/misc/ItemHandlerAdapter.java
@@ -25,6 +25,8 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import com.google.common.primitives.Ints;
+
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.items.IItemHandler;
 
@@ -42,7 +44,6 @@ import appeng.core.Api;
 import appeng.me.GridAccessException;
 import appeng.me.helpers.IGridProxyable;
 import appeng.me.storage.ITickingMonitor;
-import appeng.util.Platform;
 import appeng.util.item.AEItemStack;
 
 /**
@@ -93,9 +94,7 @@ class ItemHandlerAdapter implements IMEInventory<IAEItemStack>, IBaseMonitor<IAE
 
     @Override
     public IAEItemStack extractItems(IAEItemStack request, Actionable mode, IActionSource src) {
-
-        ItemStack requestedItemStack = request.createItemStack();
-        int remainingSize = requestedItemStack.getCount();
+        int remainingSize = Ints.saturatedCast(request.getStackSize());
 
         // Use this to gather the requested items
         ItemStack gathered = ItemStack.EMPTY;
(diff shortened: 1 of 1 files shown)


Technici4n:
Should this target 8.4.0 instead? Not sure as there can always be an unforeseen bug.

shartte:
I'd keep it for 1.17

yueh:
Changes to `extractItems` are fine, also for 8.4.

However I don't see the point in `injectItems`, this just increase the complexity on our side for the sole reason of Forge not giving a shit about performance. Also I would not trust the return from `ItemHandlerHelper.insertItem()` to be stable. Especially there are some edge cases with empty itemstacks, which can still leak the original item and become valid again. So this most likely will need a defensive copy at some point remove any potential improvement and actually be worse than before.

Also it is more easy to make API changes in 1.17, which could fix it in a different way and not have to deal with incorrect cache invalidation. E.g. we could change all extract/insert operations into using a context record, which can pass the original `ItemStack` along on top of the `IAEItemStack`. So anything which is a end to end communication between two different `IItemHandler` can just use this stack.

Technici4n:
Hmmm, do you think only caching if `returnedStack == originalStack` would be safe enough? That's the case I want to optimize in particular: iteration through many storage busses that reject the item, for example through a large array of deep storage units.

yueh:
No, we cannot rely on the other inventory not returning a copy. Even when unchanged. Inventory interaction in minecraft are really worst practice possible and we have to be defensive for every single case. Usually there are a couple broken mods around, which are highly used but are technically abandoned and would need months for a fix to be released, if ever.

Technici4n:
Ok, optimizing extract already provides a large gain, so let's do that only for now. I think I have a solution for the event problem after discussion with pup yesterday on forgecord, but that will be for 1.17 if it ever gets implemented.

yueh:
I doubt they will fix it, there was a push for a better `IItemHandler` API probably like 2 years ago, which as usually derailed into some transaction crap, but had a couple easy fixes and Lex or any other forge member happily ignores that.

We have our own cache in this case to be reused, but there is no way to validate it without scanning the whole inventory before every extract or insert.

Technici4n:
I was talking about making `ItemStack#copy` not fire the attach capabilities event, which would make copies very cheap. We'll see what can be done.

A better `IItemHandler` is probably never going to happen on Forge.

shartte (approved):

PR #5341, opened 2021-07-17
Added note to Github community Guidelines

Pingger:
Additionally fixed a typo in the contributing.md

PS: Feel free to ping me, in case you need someone to politely explain to someone that the issue templates are there for a reason and that politeness in issues is required


Technici4n:
I don't think this is needed. Explaining common sense should not be necessary, and I doubt this would help in any way.

Pingger:
You should still fix the typo in the second commit

shartte:
Yes, we should do that. 
I appreciate your thoughts on this, but I also think angry people are gonna rush right past those guidelines and vent in an issue regardless :)

PR #5355, opened 2021-08-07
Remove potential race condition in deserializing ME inventories

### LINKED ISSUE #5329: (Fabric) Items do not appear if you close and open terminal again
**Describe the bug**

Items and Recipes inside the ME network doesn't appear on the terminal but data is being occupied in Storage Cell as it should.

**To Reproduce**

 1. Set up a simple ME system with a ME Drive with a storage cell and a Terminal
 2. Open the terminal and put any item in the system (the item will appear properly)
 3. Close the terminal and re-open it, the item is not there anymore.
(this happens with an ME Chest too and ME Interface recipes).

**Expected behavior**

Items and craftings showing properly in the terminal

**Additional context**

I don't think there's something significant in the logs, but a couple of warning and models some models not loaded
[latest.log](https://github.com/AppliedEnergistics/Applied-Energistics-2/files/6786139/latest.log)

**Environment**

 - Game Setup:
   * MultiMC
   * Fabric Loader 0.11.6
   * Fabric API 0.36.0+1.16 (the same things occurs with API version 0.34.7+1.16)
   * Applied Energistics 2 8.4.0-alpha.1

- Environment: 
  * Ubuntu 20.04.2 LTS
  * Intel® Core™ i7-6500U CPU @ 2.50GHz × 4 
  * Mesa Intel® HD Graphics 520 (SKL GT2)

shartte:
to ensure we see the effects of a previous OpenContainer packet before trying to access the container for deserializing items.

Fixes #5329

### DIFF
--- a/src/main/java/appeng/core/sync/packets/MEInventoryUpdatePacket.java
+++ b/src/main/java/appeng/core/sync/packets/MEInventoryUpdatePacket.java
@@ -19,7 +19,6 @@
 package appeng.core.sync.packets;
 
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.function.Consumer;
 
@@ -62,33 +61,17 @@ public class MEInventoryUpdatePacket<T extends IAEStack<T>> extends BasePacket {
     private static final int INITIAL_BUFFER_CAPACITY = 2 * 1024;
 
     // input.
-    private final List<GridInventoryEntry<T>> list;
-
-    private boolean fullUpdate;
-
-    private int windowId;
+    private PacketBuffer bufferedClientSideData;
 
     public MEInventoryUpdatePacket(final PacketBuffer data) {
-        int itemCount = data.readShort();
-        this.windowId = data.readVarInt();
-        this.fullUpdate = data.readBoolean();
-        this.list = new ArrayList<>(itemCount);
-
-        // We need to access the current screen to know which storage channel was used to serialize this data
-        MEMonitorableContainer<T> container = getContainer();
-        if (container != null) {
-            IStorageChannel<T> storageChannel = container.getStorageChannel();
-            for (int i = 0; i < itemCount; i++) {
-                this.list.add(GridInventoryEntry.read(storageChannel, data));
-            }
-        }
+        // We need to wait until we're on the main thread to parse this data,
+        // because the storage channel is only known to the open container.
+        bufferedClientSideData = data;
+        bufferedClientSideData.retain();
     }
 
     @SuppressWarnings("unchecked")
-    private MEMonitorableContainer<T> getContainer() {
-        // This is slightly dangerous since it accesses the game thread from the network thread,
-        // but reading the current container is atomic (reference field), and from then the window id
-        // and storage channel are immutable.
+    private MEMonitorableContainer<T> getContainer(int windowId) {
         ClientPlayerEntity player = Minecraft.getInstance().player;
         if (player == null) {
             // Probably a race-condition when the player already has left the game
@@ -112,7 +95,6 @@ public class MEInventoryUpdatePacket<T extends IAEStack<T>> extends BasePacket {
 
     // api
     private MEInventoryUpdatePacket() {
-        this.list = Collections.emptyList();
     }
 
     // Byte offset to the field in the packet that contains the item count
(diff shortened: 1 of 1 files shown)


Technici4n (approved):
Works as expected.

Fourmisain:
Hm... I think this is still not 100% thread-safe, but it's very close to!

If I understood this right, the `bufferedClientSideData` field is written to on the networking thread while it is read on the main thread without any synchronization.
The reason it "works" should be due to the higher delay between the two actions, but it is no guarantee.

This is easily fixed by e.g. making `bufferedClientSideData` `volatile`.
While this adds _some_ overhead, the performance impact should be negligible considering how rare these packets are.

If performance _is_ a concern, `clientPacketData()` can be changed to read the `bufferedClientSideData` variable only once (by assigning it to a local variable and using that to parse the packet) - that way there is only 1 synchronized read instead of 5.

shartte:
That is not how that works.

shartte:
p.s.: to not sound like an ass here and elaborate on my point:
the packet is parsed on the network thread, then it's passed onto the main thread for processing, then it's discarded.
the buffer is a copy, it's not reused. 

The actual race came from two subsequent packets (1: open container, 2: inventory content), where both were parsed before packet 1 being handled on the main thread. the reliance of packet 2's parsing on packet 1 already being processed (because it needed the open container for parsing its content) was the actual race. by delaying the packet parsing until packet 2 is on the main thread solves this problem since packet processing on the main thread is serialized. 

This works so neatly because client-side the payload of custom packets is already copied by vanilla before being passed to the handler. So we can hang onto the buffer until our packet object is handled on the main thread without risk of the buffer being discarded or reused.

Fourmisain:
Hm... I think we kinda missed each other.

I'm not saying that there's an issue with how the packet data is processed on the main thread or anything, that is perfectly fine and as you mentioned it always happens in the correct order now.

What I am saying is that there's a (small) issue in how the packet is passed to the main thread.

Going by the [debug logs](https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/5329#issuecomment-894550690) in my comment, first the `MEInventoryUpdatePacket` constructor is called:

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/1efa129edb727cd29497abd1a8a8b7cf27194bcf/src/main/java/appeng/core/sync/packets/MEInventoryUpdatePacket.java#L66-L71

This writes the packet data to the local field `bufferedClientSideData` and this happens on the network thread.

Now, afterwards `clientPacketData()` is executed:

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/1efa129edb727cd29497abd1a8a8b7cf27194bcf/src/main/java/appeng/core/sync/packets/MEInventoryUpdatePacket.java#L237-L239

And this is where the local field `bufferedClientSideData` is read and this happens on the main thread.
By "read" I don't mean parsing the packet data but I mean reading the *reference*.

Since you are accessing the same field `bufferedClientSideData` from 2 different threads without synchronization, there is a (small) chance that the main thread reads `null` here.

shartte:
No that chance doesn't exist.
Even without knowing anything about the network subsystem, the `this` reference of the packet that is being constructed can only be passed onto the main thread after control has left the constructor, **unless** our own base-class would leak the reference before the constructor has finished (which we don't do).

So from that follows: If the reference to the packet can only be passed to the main thread after its constructor has run, and that constructor sets the field, how would it be possible for the field to be null when the main thread receives the reference?

Fourmisain:
Ah, you're right about it being set in the _constructor_, this is actually much better than any method, but I do remember reading that fields of constructed objects are only guaranteed to be visible to other threads if they are at least `final` or synchronized in some form.

Here's one reference I found, which quotes Java Concurrency in Practice:
https://stackoverflow.com/questions/6457109/java-concurrency-is-final-field-initialized-in-constructor-thread-safe

> Presence of final guarantees that other threads would see values in the map after constructor finished without any external synchronization. Without final it cannot be guaranteed in all cases

(The answer goes on to explain how it can be synchronized e.g. using `volatile` or `AtomicReference` too.)

If that's not enough, I looked up the formal definition too, here it is:
https://docs.oracle.com/javase/specs/jls/se7/html/jls-17.html#jls-17.5

> A thread that can only see a reference to an object after that object has been completely initialized is guaranteed to see the correctly initialized values **for that object's final fields**. 

This implies that there's no such guarantee for non-final fields.

So scratch the idea of using `volatile` but simply add a `final`, then it's guaranteed thread-safe without any overhead!

This is the surface answer, going by the language definition and in some sense, that's all there is to the language, but I do understand your specticism.

I'm not trying to sound like an ass too, but the truth is simply that multithreading/concurrency is **hard**, extremely hard, because it is often times almost impossible to *prove* that something is in fact thread-safe and code can work 99.999% of the time and then suddenly fail which makes it equally hard to *disprove* thread-safety and I do believe this is one of those 99.999% types of things, a quantum bug so to speak.

The answer to your question (which I remember seeing a bunch of times e.g. on StackOverflow)
> how would it be possible for the field to be null when the main thread receives the reference?

is really hard to demonstrate, because this is the scenario where one thread executes something *after* another thread on a global time scale, which is hard to model in a test program without creating the second thread after executing the first (which *does* guarantee that it can see the completely-initialized object including *all* fields) and without using any synchronization.
It's probably possible to write such a program, but it won't be easy.

Fourmisain:
P.S. To name one example of something which is commonly used but for which no proof of thread-safety exists, look no further than the implementation of `ConcurrentHashMap`.

Technici4n:
No need, the `ConcurrentLinkedQueue` used by the task executor guarantees that the write to the private field happens-before the read of the private field on the main thread. See also https://stackoverflow.com/questions/17108541/happens-before-relationships-with-volatile-fields-and-synchronized-blocks-in-jav.

shartte:
He is right though in that my initial assertion that control leaving the constructor guaranteeing the field being written is not true when the JIT inlines the constructor and starts reordering memory access except for final field initializers.
So it does rely on outside knowledge about how MC passes along objects network->main thread and the involved memory barriers.

Fourmisain:
> No need, the `ConcurrentLinkedQueue` used by the task executor guarantees that the write to the private field happens-before the read of the private field on the main thread. See also https://stackoverflow.com/questions/17108541/happens-before-relationships-with-volatile-fields-and-synchronized-blocks-in-jav.

Fantastic!
That clears everything up - I haven't thought about how the constructor and `clientPacketData()` are actually called, good to know it has strong guarantees.

> (However. If MC did not have a memory barrier in place when it hands off the data, this would probably be the least of our problems :-P)

Haha, probably so 😄

Sorry for pulling you into this long discussion which turned out irrelevant!

PR #5380, opened 2021-08-20
Simplify inventory adapters

Technici4n:
* Cleanup adaptors: remove unused classes and methods, document, reorganize a bit.
* Get rid of `MEMonitorIInventory` and `MEMonitorIFluidHandler`.
* Replace all uses of `InventoryAdaptor` by `IItemHandler` use through `ItemHandlerUtil`.


Technici4n:
(Draft because I haven't ran the game yet... tomorrow ;))

src/main/java/appeng/menu/implementations/InterfaceTerminalMenu.java
```diff
@@ -196,44 +195,46 @@ public void doAction(final ServerPlayer player, final InventoryAction action, fi
...
 
-                ItemHandlerUtil.setStackInSlot(theSlot, 0, playerInv.addItems(theSlot.getStackInSlot(0)));
+                ItemHandlerUtil.setStackInSlot(theSlot, 0,
+                        ItemHandlerUtil.insert(playerInv, theSlot.getStackInSlot(0), Actionable.MODULATE));
 
                 break;
             case MOVE_REGION:
 
-                final InventoryAdaptor playerInvAd = InventoryAdaptor.getAdaptor(player);
+                var playerInvAd = new PlayerMainInvWrapper(player.getInventory());
```
Technici4n, line 234: Actually the adaptor tried to stack existing items, while `PlayerMainInvWrapper` doesn't... might need to roll a wrapper around it.

yueh: There might be further issues. E.g. powah will craft stacks of unstackable items which the adaptor or itemhandler for the player hand could not handle and fail in various ways like not being able to craft it in our terminal.

Might not apply here, but should be something we have to verify.

src/main/java/appeng/parts/misc/AbstractStorageBusPart.java
```diff
@@ -331,10 +330,6 @@ private int createHandlerHash(BlockEntity target) {
         if (target != null) {
             var inv = this.getInventoryWrapper(target);
 
-            if (inv instanceof MEMonitorIInventory h) {
```
shartte, line 334: what about this?

Technici4n: Doesn't seem to be useful as interface inventories can be extracted from, so this has no effect in `MEMonitorIInventory`.

yueh:
This change breaks at least interfaces, which will no longer expose their configured inventory.

Technici4n:
Let's keep `InventoryAdaptor`. There is still more cleanup to do however.

PR #5385, opened 2021-08-21
Closes #5256: Replace `AEPartLocation` with `Direction`

### LINKED ISSUE #5256: Evaluate replacing `AEPartLocation` with `Direction`
This needs to take into account that we potentially use 8 states currently as opposed to the 7 (6 sides + `null`) provided by `Direction`.

Only possible if we never use `INTERNAL` and `null` with a different semantic. Also needs to consider if we never plan on using this option in the future.

yueh:
Completely replaced `AEPartLocation` with `Direction` with using `null` as center/internal.
Added a constant to `Platform` for an array including `null` which need to iterate over all 7 cases.
This is in preparation of moving from a custom interface to a capability or similar system.


Technici4n (approved):
Looks good.

src/api/java/appeng/api/parts/IFacadePart.java
```diff
@@ -52,7 +53,8 @@
     /**
      * @return side the facade is in
      */
-    AEPartLocation getSide();
+    @Nullable
+    Direction getSide();
```
shartte, line 57: Is it possible for a facade part to be on a null-side?

shartte: I am changing this. Hold on.

shartte (approved):

PR #5394, opened 2021-08-22
Fix nested alertDevice() calls replacing the tick queue

Technici4n:
To reproduce:
Chest -> import bus -> storage bus -> chest (put these things alone on a network with a creative cell).

The import bus doesn't get ticked after it successfully imports something.

### DIFF
--- a/src/main/java/appeng/me/service/TickManagerService.java
+++ b/src/main/java/appeng/me/service/TickManagerService.java
@@ -87,8 +87,12 @@ public class TickManagerService implements ITickManager, IGridServiceProvider {
 
         if (queue != null) {
             currentlyTickingQueue = queue;
-            tickQueue(queue);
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/me/service/TickManagerService.java
```diff
@@ -72,18 +74,30 @@ public void onLevelStartTick(Level level) {
...
-        this.tickQueue(queue);
+        this.tickLevelQueue(null);
+    }
+
+    private void tickLevelQueue(@Nullable Level level) {
+        var queue = this.upcomingTicks.get(level);
+
+        if (queue != null) {
+            currentlyTickingQueue = queue;
+            tickQueue(queue);
```
shartte, line 90: try finally?

PR #5428, opened 2021-09-14
Fluid autocrafting

Technici4n:
TODO:
- [x] Container item support in the CPU logic.
- [x] Allow interfaces to push fluids alongside items.
- [x] Allow encoding patterns with fluids.
- [x] Move pattern and stack return functionality to a new block/part (ME Crafting Interface).
- [x] Recipes for the crafting interface: initial recipe, and block/part conversion.
- [x] Crafting card in the fluid level emitter.

Left for later:
- [ ] Fluid substitution in crafting recipes.
- [ ] Properly display fluid output in the CPU monitor block.
- [ ] Fix strange behavior with `CraftingCpuLogic#injectItems` when the items can't fit in the target.


src/api/java/appeng/api/storage/IForeignInventory.java
```diff
@@ -0,0 +1,22 @@
...
+
+import javax.annotation.Nullable;
+
+import appeng.api.config.Actionable;
+import appeng.api.storage.data.IAEStack;
+
+/**
+ * Allows ME interfaces to inject crafting inputs regardless of the storage channel. Obtained from the storage channel.
+ */
+public interface IForeignInventory<T extends IAEStack> {
```
Technici4n, line 11: Should be removed for the time being.

src/main/java/appeng/core/definitions/AEItems.java
```diff
@@ -253,6 +249,7 @@
     public static final ItemDefinition<ReplicatorCardItem> DEBUG_REPLICATOR_CARD = item(AppEng.makeId("debug_replicator_card"), ReplicatorCardItem::new);
     public static final ItemDefinition<DebugPartPlacerItem> DEBUG_PART_PLACER = item(AppEng.makeId("debug_part_placer"), DebugPartPlacerItem::new);
     public static final ItemDefinition<FluidDummyItem> DUMMY_FLUID_ITEM = item(AEItemIds.DUMMY_FLUID_ITEM, FluidDummyItem::new);
+    public static final ItemDefinition<TestPatternItem> TEST_PATTERN_ITEM = item(AppEng.makeId("test_pattern"), TestPatternItem::new);
```
Technici4n, line 252: Can probably get rid of this now that we have processing pattern encoding with fluids,

src/api/java/appeng/api/crafting/IPatternDetails.java
```diff
@@ -21,21 +21,39 @@
...
+    IAEStack[] getOutputs();
+
+    interface IInput {
+        IAEStack[] getPossibleInputs();
+
+        long getMultiplier();
+
+        boolean isValid(IAEStack input, Level level);
+
+        boolean allowFuzzyMatch();
```
yueh, line 54: We should remove this and roll it directly into the crafting simulation by default.
All it accomplishes is to test for the same item just with different damage values. 
In general damageable items are not very common, so it would be more user friendly to just check if any damaged item fits into the recipe instead of completely enabling every substitute stack.

At least for crafting patterns. Processing ones cannot be verified so we have to stick with the exact item in this case.

src/api/java/appeng/api/crafting/IPatternDetails.java
```diff
@@ -21,21 +21,39 @@
...
+    IInput[] getInputs();
+
+    default IAEStack getPrimaryOutput() {
+        return getOutputs()[0];
+    }
+
+    IAEStack[] getOutputs();
+
+    interface IInput {
+        IAEStack[] getPossibleInputs();
```
yueh, line 48: This needs to ensure the array is sorted by the definition used for the pattern. Or we need a direct way to get it.

src/api/java/appeng/api/crafting/IPatternDetails.java
```diff
@@ -21,21 +21,39 @@
...
+
+public interface IPatternDetails {
+    ItemStack copyDefinition();
+
+    // TODO: is this even useful?
+    boolean isCrafting();
+
+    IInput[] getInputs();
+
+    default IAEStack getPrimaryOutput() {
```
yueh, line 41: Separate method for only the secondary outputs might also be useful.

src/api/java/appeng/api/client/AmountFormat.java
```diff
@@ -21,8 +21,10 @@
  * CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
  */
 
-package appeng.api.networking.crafting;
+package appeng.api.client;
 
-public enum CraftingItemList {
-    ALL, STORAGE, ACTIVE, PENDING
+public enum AmountFormat {
```
yueh, line 26: IIrc we already have a similar enum for the terminal slots, might overlap here.

src/api/java/appeng/api/networking/energy/IEnergyWatcherNode.java
```diff
@@ -27,7 +27,7 @@
 
 import appeng.api.networking.IGridNodeService;
 
-public interface IEnergyWatcherHost extends IGridNodeService {
+public interface IEnergyWatcherNode extends IGridNodeService {
```
yueh, line 30: `Host` is actually the more fitting name here. The actual watchers are implemented by use and act as the observer part. No idea what pattern here really applies.

src/api/java/appeng/api/networking/storage/IStackWatcherNode.java
```diff
@@ -29,7 +29,7 @@
 import appeng.api.storage.data.IAEStack;
 import appeng.api.storage.data.IItemList;
 
-public interface IStackWatcherHost extends IGridNodeService {
+public interface IStackWatcherNode extends IGridNodeService {
```
yueh, line 32: Same with `Host`

yueh (changes requested):

src/api/java/appeng/api/crafting/IPatternDetails.java
```diff
@@ -21,21 +21,39 @@
...
-     *
-     * @param job - final job
-     */
-    void calculationComplete(@Nonnull ICraftingJob job);
+import appeng.api.storage.data.IAEStack;
+
+public interface IPatternDetails {
+    ItemStack copyDefinition();
+
+    // TODO: is this even useful?
```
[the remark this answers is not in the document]
Technici4n: Removed.

src/api/java/appeng/api/crafting/IPatternDetails.java
Technici4n: Added javadoc mentioning this.

src/api/java/appeng/api/crafting/IPatternDetails.java
Technici4n: Might be, but it would require an allocation every time. Probably easier to use the full array and start iteration at 1 if people need it.

src/api/java/appeng/api/networking/energy/IEnergyWatcherNode.java
Technici4n: Renamed because of `ICraftingWatcherNode` (renamed during the grid refactor). The host referred to `IGridHost` imo, although I guess that's up to interpretation.

src/api/java/appeng/api/crafting/IPatternDetails.java
Technici4n: Removed.

PR #5439, opened 2021-09-23
Updated issue templates to use issue forms.

yueh:
Also added a dedicated discord link for the new issue page.


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,72 @@
...
+    attributes:
+      value: |
+        Provide a summary of the issue in the Title above, please do not "[Tag]" it.
+  - type: textarea
+    id: description
+    attributes:
+      label: Describe the bug
+      description: |
+        Please describe as best as possible.
+        Always use the most recent version from [our releases](https://github.com/AppliedEnergistics/Applied-Energistics-2/releases) or curseforge, it might already be fixed.
```
shartte, line 15: Why are you not linking to curseforge here? We don't publish binaries on the release-page

yueh: We actually do and it's pretty much guaranteed to be the most recent version. Curseforge is not under our control, so in case they decide to wait 2 days for their review, reject it for some reason, or it's an alpha which isn't shown by default, there might be discrepancy.

shartte: WTF, we actually do. I take it back :-P

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,30 @@
...
+    attributes:
+      label: Reasons why it should be considered
+      description: |
+        This can help us in various degrees, e.g. to determine if it fits our design ideas.
+    validations:
+      required: true
+  - type: textarea
+    id: Additional
+    attributes:
+      label: Reasons why it should be considered
```
Technici4n, line 27:
```suggestion
      label: Additional details
```

Technici4n (approved):

PR #5441, opened 2021-09-23
Changed `ItemGenBlockEntity` back to `IItemHandler`

yueh:


shartte:
Why are you not exposing the platform wrapper instead like the other inventories do?

yueh:
Complexity mostly, it's the most basic inventory possible, no UI or anything else backing it. So I don't think there is any benefit to throw another layer add it.

shartte:
The benefit is keeping is consistent across the master and fabric branches.

PR #5461, opened 2021-10-03
Add Site Export Run Configuration

shartte:


src/main/java/appeng/siteexport/SiteExport.java
```diff
@@ -0,0 +1,237 @@
...
+            runExport(client);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        } finally {
+            // Close next tick
+            GLFW.glfwSetWindowShouldClose(client.getWindow().getWindow(), true);
+        }
+    }
+
+    @OnlyIn(Dist.CLIENT)
```
Technici4n, line 88: Can probably be removed.

PR #5475, opened 2021-10-05
Provide After-AE2-Init Method for Mods

### LINKED ISSUE #5472: AE2 addon entrypoint
Add an entrypoint that fires after AE2 is finished initializing. Could be used by add-ons to ensure they register their stuff after AE2, and can thus access AE2's. Could also be used by mods who have a soft AE2 dependency to detect AE2's presence.

shartte:
Fixes #5472: Provide a method for addons to guarantee running after the mod itself has initialized


Technici4n:
Would it be possible to have this as an entrypoint for mods that wish to soft-depend on AE2 being loaded?

shartte:
I also forgot to actually call it, lol.

Technici4n (approved):

PR #5496, opened 2021-10-09
Fix custom block entity names

Technici4n:
Hmmm, now I realize that changing `getDisplayName` to `getHoverName` in `AEBaseEntityBlock` would have probably been enough, but this should be better in the long run anyway. (In mojmap, the display name is the chat display name, enclosed by `[]`, which is not a `TextComponent`).


Technici4n:
Closing due to github

PR #5614, opened 2021-10-29
Rework storage busses and internal network storage logic

### LINKED ISSUE #330: Unable to filter items presented to a subnet on the storage bus
The filter is ignored when the storage bus is used in a standard subnet

### LINKED ISSUE #3841: Subnetwork level emitters - viable vs extractable items
**Describe the feature**

I have a sub network with the storage bus configured to `insert only` and report inaccessible items set to `Yes`

I would have expected that level emitters on the sub network would "see" the items in the main network
they do if the storage bus is set to `extract` or `bi-directional`.

it would be nice if they followed the visibility setting or if storage buses could be configured with view cells and only be able to read/extract items allowed by the view cell

i have a system that autocrafts  rich slag which is being consumed constantly.  It uses cobble stone and i want it to only craft with the cobble overflow that exists in the subnetwork. however if the storage bus is `insert only` the level emitters always think there is no rich slag and craft all the overflow cobble even when there is enough rich slag. if the interface is set to extract then the level emitters work, but the crafting pulls from my main networks supply of cobble

**Additional Context**

### LINKED ISSUE #5390: Storage bus "show extractable items only" is likely broken.
It doesn't seem to be doing anything given the storage bus code.

### LINKED ISSUE #5584: Item Exporter causes lag
### Describe the bug

Noticed especially when exporting a large quantity of items. First noticed when exporting roughly 20-30k end stone into a quantum storage unit from Tech Reborn. It seems the exporter start out slow, then ramps up its speed over time (confirmed in AE2 discord as better for server performance). If there's a cap on the speed, the cap is too high, otherwise it's ramping up to a speed the game can't handle. As soon as the exporter is disconnected the lag immediately stops.
The lag seemed to be worse with 4 accelerator cards, presumably because it reaches untenable speeds faster, but still noticed the lag on an exporter without any cards, although not as extreme.

### How to reproduce the bug

Allow an item exporter to run continuously, the longer it runs the worse the lag gets.

### Expected behavior

No lag to be caused by the exporting of items, regardless of how long it runs.

### Additional details

Tech reborn quantum storage was a good way to notice the problem, as it allowed for a long period of continuous transfer. Disconnected exporter because of the lag, didn't let it get to the point of crashing, so unsure if it would get that extreme.

1.17

Fabric

### Crash log

https://pastebin.com/hayD2dTk

Technici4n:
Completely rework storage busses. Changes include:
- Storage bus on interface will properly filter subnetwork content. Closes #330.
- More generally, make storage bus filtering behave the same regardless of whether the target is a simple inventory, or a subnetwork.
- Make "report inaccessible items" work again. Closes #5390.
- By default, storage busses will filter both on insertion and extraction (previously it was only on insertion). Added an option in the storage bus GUI to only filter on extraction (off by default).
- When "report inaccessible items" is set to true, items will be made visible regardless of the configuration of the storage bus. Closes #3841.
- Only trigger the "recursive networking" advancement when a storage bus is placed on an interface and powered on.
- Never rebuild the network storage structure, even if the target handler changes.
- Closes #5584, which was due to storage busses doing constant storage structure resets, leading to being unable to take items out of the terminal due to the serial changing non-stop.


src/main/java/appeng/parts/misc/AbstractStorageBusPart.java
```diff
@@ -240,167 +227,158 @@ public final TickRateModulation tickingRequest(final IGridNode node, final int t
...
 
-        if (in != out) {
-            var after = getStorageChannel().createList();
-            if (out != null) {
-                after = out.getAvailableItems(after);
-            }
-            StorageHelper.postListChanges(before, after, this, this.source);
-        }
+    protected void onConfigurationChanged() {
+        updateTarget(true);
```
Technici4n, line 242: Should maybe just schedule a reset instead of doing it immediately?

PR #5618, opened 2021-10-30
Fix meteorites causing flat chunk generators to hang on world load

### LINKED ISSUE #5617: Fix latest meteorite spawning fix breaking flat chunk generators
Loading into a world will "hang" because a null pointer exception occurs during chunk generation.

The reason is us not specifying the meteorite variant that should spawn in the flat chunk generator.

shartte:
Fixes #5617: Register a variant of the meteorite for flat chunk generators


Technici4n:
Does this mean that meteorites spawn in flat worlds?

shartte:
I haven't actually tried, but I think so, yes.

Issue #5664, opened 2021-11-12
AEStack Removal / Major Inventory Overhaul


shartte:
This is in progress for 1.18.

- [x] fix chest monitor 
- [x] Fix memory cards with generalized inventory and remove special casing for fluids
- [x] Remove AppEngInternalAEInventory, move cell workbench over to generalstackinv
- [x] Remove fake "type only" slots (not needed with new config inventories in type mode)
- [x] Remove IAEFluidInventory, IConfigurableFluidInventory, IAETank et al
- [x] Remove FluidSlotPacket
- [x] Try removing GenericStackSyncHelper and move to wrapped GenericStacks and normal slots
- [x] Remove AEStack, IAEStack,  AEItemStack, AEFluidStack, et al.
- [x] Remove IAEItemList and all related classes
- [x] Introduce immutable replacement (`GenericStack`)
- [x] Make `GenericStack` unwrappable/wrappable into an ItemStack (replaces `FluidDummyItem`/`WrappedFluidStack`)
- [x] Implement rendering for a wrapped `GenericStack` in such a way that it delegates to `AEStackRendering`
- [ ] Move all filter and config inventories to use `Set<AEKey>` where possible (KeyCounter where not, potentially implement KeySet as a Set<AEKey> alternative with fuzzy match support)
- [ ] Investigate why PatternTermSlot has custom click behavior when `CraftingTermSlot` already seems to implement the same behavior?

PR #5665, opened 2021-11-13
Ported most code to use AEKey instead of IAEStack

shartte:
See #5664


src/main/java/appeng/api/crafting/IPatternDetailsDecoder.java
```diff
@@ -36,5 +38,12 @@
     boolean isEncodedPattern(ItemStack stack);
 
     @Nullable
-    IPatternDetails decodePattern(ItemStack stack, Level level, boolean autoRecovery);
+    IPatternDetails decodePattern(AEItemKey what, Level level);
```
Technici4n, line 41: Could be defaulted?

Technici4n: Or I'd just leave it out of the Decoder itself tbh, it should only be in the registry.

src/main/java/appeng/api/networking/crafting/ICraftingRequester.java
```diff
@@ -49,7 +49,7 @@
      *
      * @return unwanted item
```
Technici4n, line 50: Leftover or successfully inserted?

src/main/java/appeng/api/storage/IMEInventory.java
```diff
@@ -23,24 +23,28 @@
...
+ * <p>
  * Implementations should COMPLETELY ignore stack size limits from an external view point, Meaning that you can inject
  * Integer.MAX_VALUE items and it should work as defined, or be able to extract Integer.MAX_VALUE and have it work as
  * defined, Translations to MC's max stack size are external to the AE API.
  * <p/>
  * If you want to request at most a stack of an item, you need to use {@link ItemStack#getMaxStackSize()} before
  * extracting from this inventory.
  */
-public interface IMEInventory<T extends IAEStack> {
+public interface IMEInventory<T extends AEKey> {
```
Technici4n, line 47: I'd remove the default from `insert`, `extract` and `getAvailableStacks` as it can be quite error-prone I think.

Technici4n (approved):

PR #5668, opened 2021-11-13
Fix reducing empty blanks at incorrect recipe

Orygeunik:
Hello!

Fixed some problem that's mention into Discord

Step 1.
Make invalid craft
image

Step 2.
Click on encode action button
image

Step 3.
Stopping on break point into source and see decrement of blank patterns with comment such as "remove one, and clear the input slot."
image

Step 4.
Resume code executing and see into game. Empty black will be decrement, but imput slot is not wiped
image

So add stack of empty blanks into pattern terminal. Place some items (cobblestone for example) into crafting grid so that it's not give real recipes. Click encode pattern and ... voila... Your stack of empty blank at now have 63 items, and no else do (you simple kill one empty pattern).

---
At now:
The encoding do not running with no correct recipe or correct input-output result (processing pattern). (small refactoring method for more intuitive)

Clicking on button do not change count of empty blanks (in both case):
image
image


Orygeunik:
I thought a little and maybe when crafting grid is empty, and encoded pattern inserted, that should decode pattern back to empty blank patter?

Orygeunik:
@Technici4n 
i'm sorry, could you watch this if possible , please?

Technici4n:
Yes it's planned, but we are discussing internally how we want to manage branches and commit history going forward. Will be in the next release, no worries.

shartte:
Closed in #5670

PR #5676, opened 2021-11-18
Allow JEI Recipe/Usages/Bookmark on crafting status and confirm screens

PrototypeTrousers:
This pull request the allows entry stacks shown on the crafting status and confirm screens to be used for JEI integration

Since theses screens do not use custom slots or widgets, this does not make use of IIngredientSupplier


shartte:
Thanks for the PR!

I'll port this to the 1.18 master because we're unlikely to actually put this into 1.17.
Unless you are playing on 1.17 and actively want it 😁 
I have a bit of "juggling" in front of me to rebase the Forge branch onto master when Forge starts putting out 1.18 support.

shartte:
Ported to new stack system and closed in #5720

PR #5681, opened 2021-11-18
Better blocking mode and fix a few pattern issues

### LINKED ISSUE #5672: True Blocking Mode™️
### Describe the feature

Implement blocking mode by crafting the status of the current crafting job, instead of just checking if items can be extracted from the target.

Blocking mode doesn't work for most modern machines. Additionally, the current behavior is not very intuitive for players. It would be better to actually track the status of the current job.

### Additional details

I'd suggest giving each pushed crafting task a unique id, and keeping a list of currently active ids in a `WorldSavedData` to ensure that pattern providers properly unblock even if their CPU is not available anymore. We should also have a visual indication that the provider is currently blocked, and a button to unblock it if it is, both in case the player makes a mistake, and in case we forget to mark a task id as unblocked in the first iteration of the feature. :P

Technici4n:


Technici4n:
Going to merge already as there are more issues to fix with the AEKey refactor.

PR #5701, opened 2021-11-21
Pattern network status cleanup

### LINKED ISSUE #5647: autocrafting job not properly utilizing multiple interfaces with the same recipy rv6-stable-7
### Describe the bug

I have a bunch of interfaces with the same pattern present and they are not all being used to craft the item when requested. Only some of the interfaces are being used

image

### How to reproduce the bug

Have a large crafting core with 255 coprocessors and capable of 5 million bytes of crafting capacity.
Set up a 64 ME interfaces with the same processing pattern for the machine.
issue a job.

### Expected behavior

ALL machines with the same pattern, should be used for crafting provided there are enough co-processors.

### Additional details

My network configuration is as follows:
1 central network which handles routing for other networks
64 subnetworks (one with 28 P2P channels and one with 4 P2P channels)that have their channels routed via P2P through the central network.
All subnetworks interface with the central network
The central network interfaces with all the subnetworks
All channels are extracted with P2P tunnels and are fed into the main network, then all channels are extracted into ME quantum rings for use elsewhere with a pairing quantum ring.
ME storage is on a unit (32 channels [28+4])
Storage busses are on a seperate unit which are getting items from barrels (YABBA)

Because crafting jobs cannot be issued from subnetworks, P2P tunnels are extracting tunnels directly from the main network into a quantum ring. The channels exit to an AE storage cell dimention (1024 channels total) into 512 import busses, and 512 interfaces.

this is a picture of the processing pattern segment of autocrafting and a unit (28+4) handling basic item exports
(the bottom is the one that is connected directly to the central network, and it is where i am having problems)
image

this is a picture of the assembler segment of the autocrafting and the ME drive network
image

this is a picture of the central network and the side that is handling channels for the processing patterns.
image

this is a picture of a 'unit' (white rainbow). 2 subnetworks interfacing with the central network and the central network interfacing with both networks. This is the one handling ME drives. (the drives being used are 16384k drives from extra cells 2 if that makes any difference)
image

1.12 - UNSUPPORTED

Forge

### Crash log

https://gist.github.com/T3cube/fe4837a986f8b9ff348c9bea1c93ad71

Technici4n:
* Move pattern storage, craftable items, etc to a helper class.
* Dispatch craft across multiple instances of the same pattern. Fixes #5647.
* Remove pattern provider priority.
* Rework `ICraftingProvider` and related APIs to be more straightforward. Also update each provider separately.


src/main/java/appeng/me/service/helpers/NetworkCraftingProviders.java
```diff
@@ -0,0 +1,170 @@
...
+            return Collections.emptyList();
+        }
+    }
+
+    private static class CraftingProviderList {
+        private final List<ICraftingProvider> providers = new ArrayList<>();
+        /**
+         * Cycling iterator for round-robin.
+         */
+        private final Iterator<ICraftingProvider> cycleIterator = Iterators.cycle(providers);
```
Technici4n, line 118: Need to rebuild the iterator when the list changes due to an added pattern, otherwise we risk encountering a CME.

PR #5702, opened 2021-11-21
Refactor how AESubScreen works

### LINKED ISSUE #5684: Wireless Crafting Terminal doesn't return from a subscreen correctly
Adding 
```java
else if (menuHost instanceof WirelessCraftingTerminalMenuHost) {
    this.previousMenuIcon = AEItems.WIRELESS_CRAFTING_TERMINAL.stack();
    this.previousMenuType = WirelessCraftingTermMenu.TYPE;
} 
```
in [AESubScreen before the normal wireless Terminal](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/180a47210d22a286b2f6ab4c81d72c1927049615/src/main/java/appeng/client/gui/implementations/AESubScreen.java#L71)

would fix this

shartte:
Introduce ISubMenuHost/ISubMenu to allow addons easier access to the back button.

Fixes #5684

### DIFF
--- a/src/main/java/appeng/api/storage/ISubMenuHost.java
+++ b/src/main/java/appeng/api/storage/ISubMenuHost.java
@@ -11,12 +11,13 @@ import appeng.menu.ISubMenu;
  */
 public interface ISubMenuHost {
     /**
-     * Used to show the user interface of this part when returning from the priority GUI.
+     * Returns to the primary user interface for this host. Used by sub-menus when players want to return to the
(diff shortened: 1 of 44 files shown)


shartte:
@Mari023 FYI

Technici4n (changes requested):
Looks good besides this small thing.

Mari023 (approved):

PR #5706, opened 2021-11-22
Use biome category to determine fallout type

### LINKED ISSUE #5696: Restore Meteorite Fallout Matching the Environment
Previously the blocks used for fallout from meteorites depended on the top-level block used by the Biome's surface builder. This is no longer available in 1.18 since the surface building is now done using complex rules attached to the chunk builder.

See `TODO 1.18` comments in FalloutCopy & MeteoriteStructure. 

We might need to either:
- Hard code suitable blocks per biome (bad for modded biomes)
- Use biome tags for the respective fallout types and fall back to NONE if none matches

shartte:
Fixes #5696

### DIFF
--- a/src/main/java/appeng/worldgen/meteorite/MeteoriteStructure.java
+++ b/src/main/java/appeng/worldgen/meteorite/MeteoriteStructure.java
@@ -28,13 +28,9 @@ import net.minecraft.core.BlockPos;
 import net.minecraft.core.Registry;
 import net.minecraft.resources.ResourceKey;
 import net.minecraft.resources.ResourceLocation;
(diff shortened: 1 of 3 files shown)


src/main/java/appeng/api/ids/AETags.java
```diff
@@ -88,4 +89,8 @@ private AETags() {
         return TagFactory.BLOCK.create(new ResourceLocation(name));
     }
 
+    private static Tag.Named<Biome> biomeTag(String name) {
```
Technici4n, line 92: should be removed

Technici4n (changes requested):
looks good besides the useless `biomeTag`.

PR #5722, opened 2021-12-03
Update

DrHesperus:

PR #5723, opened 2021-12-03
Update russian translation

shartte:
See #5722


shartte:
Many thanks to @DrHesperus

PR #5751, opened 2021-12-08
Energy cells now always hold their charge - regardless of how they are broken

### LINKED ISSUE #5750: Energy Cell doesn't hold charge when broken
The energy cell doesnt keep its charge when it's broken with a pickaxe or wrench.

shartte:
Fixes #5750

### DIFF
--- a/src/main/java/appeng/blockentity/networking/EnergyCellBlockEntity.java
+++ b/src/main/java/appeng/blockentity/networking/EnergyCellBlockEntity.java
@@ -120,10 +120,9 @@ public class EnergyCellBlockEntity extends AENetworkBlockEntity implements IAEPo
     public void exportSettings(SettingsFrom from, CompoundTag data) {
         super.exportSettings(from, data);
 
-        if (from == SettingsFrom.DISMANTLE_ITEM) {
-            final CompoundTag tag = new CompoundTag();
-            tag.putDouble("internalCurrentPower", this.internalCurrentPower);
-            tag.putDouble("internalMaxPower", this.getInternalMaxPower()); // used for tool tip.
+        if (from == SettingsFrom.DISMANTLE_ITEM && this.internalCurrentPower > 0.0001) {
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/loot/ExportSettingsFunction.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class ExportSettingsFunction implements LootItemFunction {
+    public static final ResourceLocation ID = AppEng.makeId("export_settings");
+    public static final LootItemFunctionType TYPE = new LootItemFunctionType(new Serializer());
+
+    @Override
+    public ItemStack apply(ItemStack itemStack, LootContext lootContext) {
+        var blockEntity = lootContext.getParamOrNull(LootContextParams.BLOCK_ENTITY);
+        if (blockEntity instanceof AEBaseBlockEntity aeBe) {
+            aeBe.exportSettings(SettingsFrom.DISMANTLE_ITEM, itemStack.getOrCreateTag());
```
Technici4n, line 36: It could maybe be unfortunate that we create the tag in all cases?

src/main/java/appeng/blockentity/networking/EnergyCellBlockEntity.java
```diff
@@ -121,9 +121,10 @@ public void exportSettings(SettingsFrom from, CompoundTag data) {
         super.exportSettings(from, data);
 
         if (from == SettingsFrom.DISMANTLE_ITEM) {
-            final CompoundTag tag = new CompoundTag();
-            tag.putDouble("internalCurrentPower", this.internalCurrentPower);
-            tag.putDouble("internalMaxPower", this.getInternalMaxPower()); // used for tool tip.
+            if (this.internalCurrentPower > 0) {
```
Technici4n, line 124: Should probably check > 1e-9 or similar.

PR #5815, opened 2021-12-19
Refactor SlotSemantic to allow addons to add new semantics

### LINKED ISSUE #5813: make SlotSemantic not an enum
### Describe the feature

make SlotSemantic not an enum and make the constructor public

it would allow addons to add their own slots using their own names for those slots 
currently, addons have to use other existing SlotSemantics, whose names really don't fit (in most cases)
and they are limited to the amount of slottypes (although I am not sure why anyone would add that much slots)

### Additional details

_No response_

shartte:
Fixes #5813

### DIFF
--- a/src/main/java/appeng/menu/SlotSemantics.java
+++ b/src/main/java/appeng/menu/SlotSemantics.java
@@ -1,0 +1,102 @@
+/*
+ * This file is part of Applied Energistics 2.
+ * Copyright (c) 2021, TeamAppliedEnergistics, All rights reserved.
+ *
+ * Applied Energistics 2 is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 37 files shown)


shartte:
@Mari023 Does this work for you?

Mari023 (approved):

src/main/java/appeng/menu/SlotSemantics.java
```diff
@@ -0,0 +1,102 @@
...
+public final class SlotSemantics {
+    private SlotSemantics() {
+    }
+
+    private static final Map<String, SlotSemantic> REGISTRY = new ConcurrentHashMap<>();
+
+    /**
+     * NOTE: If you use this in an addon, use an Addon-Specific Prefix for your semantic id (i.e. your mod id).
+     */
+    private static SlotSemantic add(String id, boolean playerSide) {
```
Mari023, line 38: this has to be public, otherwise addons can't actually use it

Issue #5840, opened 2021-12-21
Starting a crafting task in the wireless crafting terminal (when in a trinket slot) does not return to the terminal screen


ramidzkh:
### Describe the bug

Starting a crafting task in the wireless crafting terminal (when in a trinket slot) does not return to the terminal screen once queued. The simulation screen however is updated in whether there are available crafting CPUs available (as the task is actually queued)

### How to reproduce the bug

1. Acquire a wireless crafting terminal
2. Equip it in a trinket slot
3. Try queuing a craft
4. Start after the craft has been simulated

### Expected behavior

The screen will return to the normal wireless crafting terminal screen once queued

### Which minecraft version are you using?

1.18

### On which mod loaders does it happen?

Fabric

### Crash log

N/A

shartte:
I am not 100% sure we can fix this, as I am not sure how trinkets are even stored/persisted.
I'd probably have to colab with @Mari023 to add support for it. I am not even sure how the GUI opens in the first place 😅

Mari023:
that is very likely an ae2wtlib issue

shartte:
Are the trinket slots represented as normal inventory slots in the player inventory? If so, then the MenuLocator with player+inv should just work, in theory at least.

Mari023:
I think trinkets are only accessible through their api, and I think they don't even have numbers, just String identifiers.
I currently mixin into MenuTypeBuilder#getHostFromPlayerInventory and change it, to search the host  in Trinkets for slot 100-200

It is somewhat a hack, and I have some ideas on how to do that properly, but that will take some time (and is a different issue)


Anyways, this is a bug in ae2wtlib, so I opened an issue there.

shartte:
@Mari023 I think we can actually make MenuLocator more generic and extendable

Mari023:
I found (and fixed) the bug
I use WirelessTerminalItem#openFromInventory to (re-)open the terminal.
It has to look for the itemstack, which is in a trinket slot, so it will find nothing.

PR #5846, opened 2021-12-21
Refactor MenuLocator

shartte:
Refactors MenuLocator to support being extended for addons that want to host their menus on items that aren't part of the player inventory (i.e. trinkets, baubles). Related to #5840

The addon can implement `MenuLocator` and register it with `MenuLocators#register`


shartte:
@Mari023 This should allow you to implement MenuLocator for trinkets, right? (I also made it so that slot is now optional for ItemMenuHost, and the AEBaseMenu doesn't try locking a slot that doesn't exist).

shartte:
I'll merge it for now, we can still refine it.

I have to check `openFromInventory` to form an actual opinion, but I think the easier way is to just call `locate`, and then use `instanceof ItemMenuHost itemMenuHost` on the result.

PR #5854, opened 2021-12-22
use vanilla findSlot insead of own getSlot Method

DrummerMC:


src/main/java/appeng/mixins/AbstractContainerScreenAccessor.java
```diff
@@ -0,0 +1,15 @@
+package appeng.mixins;
+
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.gen.Invoker;
+
+import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen;
+import net.minecraft.world.inventory.Slot;
+
+@Mixin(AbstractContainerScreen.class)
+public interface AbstractContainerScreenAccessor {
```
Technici4n, line 10: We tend to prefer access wideners (the equivalent of access transformers on fabric).

Technici4n (changes requested):

src/main/java/appeng/mixins/AbstractContainerScreenAccessor.java
DrummerMC: done

PR #5855, opened 2021-12-22
Fix #96: optionally reduce amount in crafting calculations on failure

### LINKED ISSUE #96: Adaptable crafting orders
If you put a crafting card in an export bus with 2 acceleration cards it will order crafting of 32 items at a time, but if you have only materials for 30 of said items if wont craft at all. I use acceleration cards as this items are used in other recepies and they take a while to craft so I make bigger orders and have a supply for next time they are needed. It would be cool if it would craft 30 instead of 32.
Maybe when you check if order is valid(there are enough materials for it) calculate how many it can craft and craft that many instead.

Originally Posted by auboli77

This issue was migrated; For original details see http://ae-mod.info/Tracker/0987/

Technici4n:
Only enabled for MultiCraftingTracker requests (interfaces and export busses) for now. Might look into offering the option to players one day?

### DIFF
--- a/src/main/java/appeng/crafting/CraftingCalculation.java
+++ b/src/main/java/appeng/crafting/CraftingCalculation.java
@@ -18,13 +18,20 @@
 
 package appeng.crafting;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 import com.google.common.base.Stopwatch;
 
+import org.jetbrains.annotations.Contract;
+import org.jetbrains.annotations.Nullable;
+
 import net.minecraft.world.level.Level;
 
 import appeng.api.networking.IGrid;
+import appeng.api.networking.crafting.CalculationStrategy;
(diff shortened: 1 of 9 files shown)


src/main/java/appeng/crafting/CraftingCalculation.java
```diff
@@ -36,50 +43,50 @@
...
-
     private final NetworkCraftingSimulationState networkInv;
     private final Level level;
     private final KeyCounter missing = new KeyCounter();
     private final Object monitor = new Object();
     private final Stopwatch watch = Stopwatch.createUnstarted();
     private final CraftingTreeNode tree;
-    private final GenericStack output;
+    private final AEKey output;
+    private final long requestedAmount;
```
shartte, line 53:
```suggestion
    // The initially requested amount of "output", may be reduced depending on the strategy used
    private final long requestedAmount;
```

Issue #1603, opened 2015-06-15
Fuzzy Exporting for ME Interface


petabyteboy:
The fuzzy card can't be placed in the interface, but the interface's export function also does not tolerate metadata difference from configured to real item. 
It should be possible to place a fuzzy card in an ME Interface in order to create a fuzzy exporting behaviour.

yueh:
And how should it merge two different items?

plumpudding:
what do you mean by merge? I mean that when there is a broken pickaxe in the config port and you put a brand new pickaxe into the slot below it should not pull that in and replace it.

yueh:
Fuzzy also matches any oredict entry as well as ignoring NBT data.

So for example if you set it to keep a stack of copper A it can also use copper B, but these are still two different items. Should it start converting them into the configured item? Which also mean that anything which can stack when fully repaired can fetch any item from the network and just repair it to stack them.

plumpudding:
for the first example: if the fuzzy comparasion says these are the same oredict: keep the one in there which is in there

shartte:
Implemented by #5863

PR #5863, opened 2021-12-23
Fix #1693: Add fuzzy card support for ME interface

### LINKED ISSUE #1693: Wireless Terminal Enhancement
I think it would be a good idea if we could colour the ME Wireless Access Terminals. People who have multiple sub-networks, with different colored cables (and terminals of course), like red for Blood Magic, lime for Botania, purple or magenta for Thaumcraft etc. And if I want wireless access to like every network, it would be a much easier way of searching the terminals in my inventory (even if I think I can rename them). The colours will be the same as all the other things: Fluix (the normal one) and than the 16 different coloured Wireless Access Terminals. It's just a minor enhancement.

And another thing I would like (and I think a bunch other people as well) are the Wireless Crafting Terminals. This might be too Overpowered, but it's just a suggestion, maybe that's why you didn't already added this Crafting Terminal in the game. But I think at least a Portable Crafting Cell would be a good idea. (and maybe the Portable Cells in multiple colours)

I hope these are good suggestions.

Technici4n:


shartte:
Fixes #1603

PR #5874, opened 2021-12-24
Fix JEI integration not working on the crafting status and confirm screens

PrototypeTrousers:
Behavior is now a bit closer to the REI plugin implementation.
It cant be the same as JEI needs a FluidStack and the REI works with a Fluid.


shartte:
Yeah, tested and it looks good. Stacks can be of size 0 which throws off the amount calculation here, and the amount should really not matter to JEI.

Will squash this with the Forge commit on the next Forge release.

PR #5894, opened 2021-12-26
Skystone tank

### LINKED ISSUE #5450: Add a Sky Stone Tank
Similar to the sky stone chest, we should provide a basic Sky Stone Tank which can hold like 32 B.
Mostly for completeness and to actually have a block to debug fluid buses and similar

Azercoco:
Add a Simple Sky Stone Tank with a capacity of 32 Buckets.

Closes #5450


src/main/java/appeng/blockentity/storage/SkyStoneTankBlockEntity.java
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        @Override
+        protected long getCapacity(FluidVariant variant) {
+            return FluidConstants.BUCKET * BUCKET_CAPACITY;
+        }
+
+        @Override
+        protected void onFinalCommit() {
+            SkyStoneTankBlockEntity.this.markForUpdate();
```
Technici4n, line 39: Should probably also call `setChanged()` (aka `markDirty`).

Technici4n: Testing that it now shows in #github.

Technici4n: Test again.

PR #5908, opened 2021-12-27
Allow clusters not being automatically selected by automation or players

### LINKED ISSUE #5886: Disable a certain crafting cluster being used by automated requests
### Describe the feature

A block which can be part of a crafting cluster which can suppress that cluster from being used for automated crafting requests caused by export buses or interfaces

If you have more crafting requests than processors, such as when items are being stocked in an interface/storage bus set-up, it becomes very difficult to queue a task manually

### Additional details

_No response_

ramidzkh:
Closes #5886


Technici4n (approved):
Looks amazing.

src/main/java/appeng/me/service/CraftingService.java
```diff
@@ -267,46 +291,52 @@ public ICraftingLink submitJob(ICraftingPlan job, ICraftingRequester requestingM
...
+            var firstPreferred = a.isPreferredFor(src);
+            var secondPreferred = b.isPreferredFor(src);
+            if (firstPreferred != secondPreferred) {
+                // Sort such that preferred comes first, not preferred second
+                return Boolean.compare(secondPreferred, firstPreferred);
+            }
 
-        return null;
+            if (prioritizePower) {
+                return FAST_FIRST_COMPARATOR.compare(b, a);
```
Technici4n, line 329: Is `(b, a)` instead of `(a, b)` intended here?

Technici4n (changes requested):

PR #5914, opened 2021-12-27
Add Portable Cell Energy Upgrade

### LINKED ISSUE #5891: Allow Portable Cells with More Energy Capacity
Allow Dense Energy Cells to be used for crafting a portable cell, increasing its energy storage capacity by 8x.
To avoid item ID explosion, use NBT for this?

Azercoco:
Add a Energy Card (Dense Energy Cell + Advanced Card) which can be used to upgrade a Portable Cell by right-clicking on it with the card. Closes #5891.
Also greatly improves the item tooltips.


shartte:
Not the biggest fan of the mechanics to add it to the portable cell right now. But it's better than not having it. We should port this to use normal upgrade slots when we actually add them to the portable terminals (#5935).

src/main/java/appeng/core/localization/Tooltips.java
```diff
@@ -1,19 +1,27 @@
...
+import appeng.api.config.PowerUnits;
 
 /**
  * Static utilities for constructing tooltips in various places.
  */
 public final class Tooltips {
 
+    private static final DecimalFormat format = (DecimalFormat) DecimalFormat.getInstance();
+    private static final DecimalFormatSymbols symbols = format.getDecimalFormatSymbols();
+    private static final char SEP = symbols.getDecimalSeparator();
```
shartte, line 23: DecimalFormat et al are not thread-safe and shouldn't be stored in a static field unless synchronized

shartte (changes requested):

Azercoco:
> Not the biggest fan of the mechanics to add it to the portable cell right now. But it's better than not having it. We should port this to use normal upgrade slots when we actually add them to the portable terminals (#5935).

I agree but the energy card will be already available when this will be implementend

PR #5945, opened 2022-01-01
Rework upgrades to allow addons to add their own upgrades / make portable items upgradable

### LINKED ISSUE #5935: Support Upgrades in Portable Terminals / Cells
The portable cells and portable terminals should have upgrade slots to allow both addons and AE2 itself to offer upgrades for them.

shartte:
Fixes #5935

Made wireless terminals have upgrade slots.
Reworked the energy card to have like a normal upgrade and make it compatible with wireless terminals, color applicator, matter cannon, and portable cells.
Allow the item charge rate to be item stack dependent.


shartte:
@Mari023 This will likely cause some major (?) breakage for you, but it enables you to make the infinity booster a "proper" upgrade card, I hope

PR #5956, opened 2022-01-02
fix #5955: fix network tool not save stack size on stack size changed

### LINKED ISSUE #5955: Network tool not save the amount correctly
### Describe the bug

The Network Tool does not save the stack size correctly

### How to reproduce the bug

- Add one upgrade card to the network tool
image
- Add more items to the same stack in the network tool
image
- Reopen the network tool and the stack size in incorrect
image

### Expected behavior

Stack size is correctly be saved on the network tool

### Additional details

_No response_

1.18

Fabric

### Crash log

none

DrummerMC:


shartte:
Alright, I did test that we do not get the Storage Bus on Drive regression that I was a little afraid of.
I put a breakpoint in `appeng.blockentity.storage.DriveBlockEntity#onChangeInventory` while a storage bus is attached to the drive, and it is not called every tick.

So in conclusion: LGTM

PR #5976, opened 2022-01-07
don't return more than the original injected input

Mari023:
should fix https://github.com/TeamAOF/All-of-Fabric-5/issues/65


Mari023:
interestingly, this is already accounted for in AEBaseBlockItemChargeable
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/94624c1c7cdeaaa553b70b990295cc731b392bc0/src/main/java/appeng/block/AEBaseBlockItemChargeable.java#L68

I think it would be better to adjust this fix to be more inline with that (it is technically the same, but it would be nice if it was consistent)

Technici4n:
> I think it would be better to adjust this fix to be more inline with AEBaseBlockItemChargeable (it is technically the same, but it would be nice if the code is consistent)

Agreed. Do you want to do it?

Mari023:
actually the bug was still in AEBaseBlockItemChargeable, I fixed it there too

Technici4n:
Thanks!

Issue #5974, opened 2022-01-07
Crafting terminal not clearing when using REI


xanthian:
### Describe the bug

1.18.1 Fabric
Beta 4 AE2
REI 7.1.357

https://user-images.githubusercontent.com/7688001/148534191-fcb42cd8-8213-461a-885f-8668e2e1362b.mp4
Easier to show in a video.

This did not occur in beta 3

no other mods installed (created a test instance to prove this)

### How to reproduce the bug

Using REI to add an item to the terminal.
Leave items there.
Using REI add another item to the terminal (different shape), items will remain from previous attempted craft.

### Expected behavior

clear terminal as per beta 3

### Which minecraft version are you using?

1.18

### On which mod loaders does it happen?

Fabric

### Crash log

.

k3nw4y:
Same version of Applied Energetics2 and Roughly Enough Items
Crafted Furnace
![image](https://user-images.githubusercontent.com/46229190/148650440-c9d52785-3452-43b1-bd5f-db012b99ec92.png)
Used REI to import recipie for Stone cutter, it doesnt clear the items used in previous craft
![image](https://user-images.githubusercontent.com/46229190/148650460-a7a1d169-605a-409e-86c1-eee57ea92066.png)

shartte:
I have observed this locally with the wireless crafting terminal if items are missing. It doesn't seem reproducible _all the time_.
Weird.
Both of you still have space in their network inventory? The grid will currently NEVER move items back to your player inventory.

xanthian:
This is not using wireless terminal.
It does it every time.
Plenty of space in inventory - 
![image](https://user-images.githubusercontent.com/7688001/148652660-d27cec0c-b1bf-4bb6-ab18-865b51d072fb.png)
As I said, rolling back to beta 3 'fixes' this

shartte:
I'll take a look, but I don't really have a good explanation so far because it does work for me 😓 
Do you have a security terminal? It does a permission check before moving it back to the network.

xanthian:
I just created a test world.
Here is the world save : 
If you cant be bothered to open it...
ME Controller, Creative Energy Cell, ME Crafting Terminal, ME Drive, 64k Storage Cell, Black Smart Cable.
64 Stone & 64 Amethyst Shard in storage.
Add stone stairs from REI, dont craft it.
Add amethyst block from REI, stone remains.

[...]

PR #5984, opened 2022-01-09
Reorganized recipe transfer again to bring it more in line with JEI

shartte:
Added a rotation button for multiple processing outputs.


shartte:
Should fix #5974

PR #6012, opened 2022-01-16
More seamless recoloring of cables.

### LINKED ISSUE #5999: Using the color applicator causes a network reboot even if connections don't change
### Describe the bug

Using the color applicator causes a network reboot even if connections don't change

### How to reproduce the bug

1. Place a cable attached to a powered network
2. Use a color applicator to change its colour
3. See WAILA/WTHIT reboot

### Expected behavior

If the cable doesn't change adjacent connections, the network shouldn't bother rebooting

### Additional details

_No response_

1.18

Fabric

### Crash log

N/A

shartte:
Do not replace the cable parts, instead just replace the part item and reset the grid color. Also sever existing, now incompatible connections if the node changes color.

Fixes #5999

### DIFF
--- a/src/main/java/appeng/parts/networking/CablePart.java
+++ b/src/main/java/appeng/parts/networking/CablePart.java
@@ -136,6 +136,7 @@ public class CablePart extends AEBasePart implements ICablePart {
 
                 getMainNode().setGridColor(getCableColor());
                 getHost().markForUpdate();
+                getHost().markForSave();
                 return true;
             }
         }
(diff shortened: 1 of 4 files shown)


src/main/java/appeng/parts/networking/CablePart.java
```diff
@@ -132,7 +132,10 @@ public boolean changeColor(AEColor newColor, Player who) {
                     return true;
                 }
 
-                this.getHost().replacePart(newPart, null, who, null);
+                setPartItem(newPart);
+
+                getMainNode().setGridColor(getCableColor());
+                getHost().markForUpdate();
```
Technici4n, line 138: Should this also markForSave?

shartte: Good find, yes it should

Issue #6004, opened 2022-01-15
Numbers on item storage monitors are almost impossible to see.


dan96kid:
### Describe the bug

Item count numbers on item storage monitors are difficult or almost impossible to see when using any cable color that causes the monitor to have a light colored background.
![image](https://user-images.githubusercontent.com/14021093/149599407-49a5b5ed-dd1a-43e8-9cd0-ec8cce490076.png)

### How to reproduce the bug

1: Place an item storage monitor on a colored cable, such as white.
2: Set the newly-placed monitor to display something stored in the connected ME system

### Expected behavior

The item count number is easily visible no matter what color the cable behind the monitor is attached to is.

### Which minecraft version are you using?

1.16

### On which mod loaders does it happen?

Forge

### Crash log

https://pastebin.com/E633vmfH

PR #6014, opened 2022-01-16
Fix for #6004 for the 1.16.5 version of AE2

dan96kid:
I originally tried to backport the official fix for #6004 but found it too complicated. I opted to instead turn the number black.


shartte:
I am not a fan of not actually porting the real fix, tbh. But since it's "good enough", I'll allow it.

shartte (approved):

PR #6021, opened 2022-01-22
Double the number of inputs and outputs for processing patterns

### LINKED ISSUE #5995: Reconsider allowing processing patterns with more than 9 unique items
### Describe the feature

(Extracted from #5987)
Combine processing patterns in an anvil/smiting table/matter condenser/inscriber to create processing patterns with more than 9 different items. In #3540, 4 years ago have been given excellent points about the theme by the issue’s author that I will not repeat myself here (please read there). Again, as this is a highly requested feature and ae2 have reconsidered those before I would like to give it another try now 4 years later to add some present perspective to invalidate some justifications for denial in the past, since I can’t do in the original feature request as it is already blocked.
In the old days few mods had this lack... but as the world of modded Minecraft has evolved and now there are many crafting mechanics that have support to use more than 9 unique item (Create's Mechanical Crafter, Draconic's Fusion Crafting, Botania's Runic Altar, Mythic Botany’s Rune Ritual, Occultism’s rituals, Astral Sorcery’s altars, Extended Crafting’s tables...) to solve this limitation is possible to use the work around of (micro)crafting to split that kind of recipes in smaller ones but that's bad thing because it loses all the magic and sense of scale of the endgame crafting and turns into a rabbit hole of a 3x3 crafting chain. It’s not like modpacks authors didn't want us to automate these recipes it’s just that they preferred the grandeur of giant crafting than having to subdivide it into minors just to be automatable. So, they are limited to using this type of recipe only for those who are one-time craft or can be bulk crafted. See the ATM star ATM6 for example, or any other endgame recipe for popular tech modpacks in versions after 1.16 (All the Mods, Seapolis, FTB oceanblock, Enigmatica, Skyopolis). 
imageimageimageimageimageimageimageimage
So my proposal is this: when in Pattern Encoding Terminal’s Process Mode if you don't set any output items and still hit the encoding recipe button you would be returned with a Partial Processing Pattern (that has only encoded input items) it could be combined with a Processing Pattern resulting in another Processing Pattern which one will have the Partial Processing Pattern inputs added to the original Processing Pattern, obviously this could be repeated and to create giant Processing Patterns and without needing any UI change to have a NxN matrix in P.E.T. (which would be a bad design choice in my opinion). At the end is just another nice mechanic that would remove the annoying 9 unique items limitation from modded Minecraft and if for some reason some modpack dev don't want it to be used it’s just a matter of removing the recipe of combining the patterns or the Partial Processing by itself.
imageimage

Done above.

### Additional details

_No response_

shartte:
This implements 18 input slots and 6 output slots for processing patterns.
Remaining limitation: only the first output is considered as the primary output and is requestable. (@Technici4n Is there a technical reason for this?)

The scroll-bar on the left allows scrolling down 3 rows.
grafik

Fixes #5995

- [x] Move "clear" button for processing mode to align with grid
- [x] Decide whether we want to highlight the primary output slot better
- [x] Actually implement and the encoding process of larger patterns
- [x] Check what happens when switching between crafting<->processing mode
- [x] Test recipe transfer with 10+ ingredients and 4+ outputs
- [x] Ensure clearing will actually clear all slots, not just the first 9


Technici4n:
> Remaining limitation: only the first output is considered as the primary output and is requestable. (@Technici4n Is there a technical reason for this?)

No, we could make multiple outputs primary if we wanted to but it probably needs to be clear from a UI perspective.

PR #6041, opened 2022-01-30
Fixes #6031: allow addons to register behavior for custom AEKeyTypes

### LINKED ISSUE #6031: API improvements for adding custom AEKeyTypes
In [my mod](https://github.com/ramidzkh/ae2-mekanism-addons/) to add Mekanism's 4 chemicals as `AEKeyType`s, I have found that the existing API isn't good enough for such a task while being on-par with the standard item and fluid keys

1. Lack of world interactions
It's impossible to actually have a custom type in the system, such as with a storage bus. The `StackWorldBehaviors` class looks promising but most of the dependent interfaces are sealed off, along with `StackWorldBehaviors` itself being non-API. 

2. Pattern provider return inventories and interfaces are hardcoded for the `ItemVariant` and `FluidVariant` storages, so they cannot be made to accept other lookups through the API

3. Lack of menu interactions
It's not possible through the API to declare how a container can be emptied or filled into the system through the UI provided by terminals, pattern providers and interfaces

4. Lack of support in pattern providers
Pattern providers are hardcoded to fast fail on non-standard `AEKeyType`s, for the system not knowing how to push non-standard `AEKeys`. Related to point 2

Technici4n:
Support is still missing for the ME Chest and Matter Condenser, but that should be trivial to reimplement in terms of the `GenericInternalInventory`.


src/main/java/appeng/api/behaviors/ContainerItemStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+import net.minecraft.world.item.ItemStack;
+
+import appeng.api.config.Actionable;
+import appeng.api.stacks.AEKey;
+import appeng.api.stacks.AEKeyType;
+import appeng.api.stacks.GenericStack;
+import appeng.menu.me.interaction.StackInteractions;
+
+/**
+ * Strategy to interact with the secondary keys held by container items (such as buckets or tanks).
```
shartte, line 17: secondary?

PR #6043, opened 2022-01-31
Optimize StorageService tick performance

Technici4n:
Optimize storage service tick performance in the following ways:
- Lazily update cache if there is no registered listener. This will ensure that the cache is not rebuilt too often for chained storage subnetworks.
- Reduce use of `KeyCounter` since I don't think the nested maps are that performant.

### DIFF
--- a/src/main/java/appeng/me/service/StorageService.java
+++ b/src/main/java/appeng/me/service/StorageService.java
@@ -29,6 +29,9 @@ import com.google.common.base.Preconditions;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.SetMultimap;
 
+import it.unimi.dsi.fastutil.objects.Object2LongMap;
+import it.unimi.dsi.fastutil.objects.Object2LongOpenHashMap;
+
 import appeng.api.networking.IGridNode;
 import appeng.api.networking.IGridServiceProvider;
 import appeng.api.networking.security.ISecurityService;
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/me/service/StorageService.java
```diff
@@ -71,25 +83,57 @@ public StorageService(ISecurityService security) {
...
-        for (var entry : previousStacks) {
-            long newAmount = cachedAvailableStacks.get(entry.getKey());
-            for (var watcher : interestManager.get(entry.getKey())) {
-                watcher.getHost().onStackChange(entry.getKey(), newAmount);
+        var previousStacks = cachedAvailableStacks;
+        cachedAvailableStacksBackBuffer.clear();
+        storage.getAvailableStacks(cachedAvailableStacksBackBuffer);
+        cachedAvailableStacks = cachedAvailableStacksBackBuffer;
+        cachedAvailableStacks.removeZeros();
+        cachedAvailableStacksBackBuffer = previousStacks;
```
shartte, line 104:
```suggestion
        var previousStacks = cachedAvailableStacks;
        cachedAvailableStacks = cachedAvailableStacksBackBuffer;
        cachedAvailableStacksBackBuffer = previousStacks;
        
        cachedAvailableStacks.clear();
        storage.getAvailableStacks(cachedAvailableStacks);
        cachedAvailableStacks.removeZeros();
```

Technici4n:
Seems to be working fine on the AOF5 public servers, so I'm going to merge this. It's not fully clear how much better this is, but the lazy updates should definitely help _a lot_.

PR #6069, opened 2022-02-13
"Split damage at 99%" fuzzy mode is confusing

NuclearMonk:
Honestly, not many better ways to write this tooltip that ar not just the technical description of how it works but this should do it. 
Still would need to be translated in the respective langs files tho.


Technici4n (approved):
Sorry for the delay in reviewing this... 😅

PR #6091, opened 2022-03-02
Chinese Localization for 10.1.0 beta2

DoremySwee:


l-ricardo:
Wouldn't it be the case to update the ReadMe localization instructions for this repository?

shartte:
Yes, you are right. We just didn't get to it yet. 
The translations from Crowdin have now automatically shown up here: https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/6093/files

PR #6097, opened 2022-03-05
Update to 1.18.2 [FABRIC]

shartte:
Updates to 1.18.2

Removes two Meteorite related config options (biome blacklist & disabling generation). Instead, use the ae2:has_meteorites  Biome tag to control this. I.e. replacing it with an empty tag should disable meteorites completely.


uselessSABOTAGE:
Do we know when this will be available?

PR #6113, opened 2022-03-20
Add GridHelper.createGridConnection with direction

artheus:
This is missing, for creating a connection where the direction can be specified. For cables to show that they are connected in a certain direction.

### DIFF
--- a/src/main/java/appeng/api/networking/GridHelper.java
+++ b/src/main/java/appeng/api/networking/GridHelper.java
@@ -160,10 +160,23 @@ public final class GridHelper {
      * @param b to be connected gridnode
      */
     public static IGridConnection createGridConnection(IGridNode a, IGridNode b) throws FailedConnectionException {
+        return createGridConnection(a, b, null);
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


shartte:
Why do you need this?

These connections should be handled by in-world grid nodes automatically and not created directly.

artheus:
> Why do you need this?
> 
> These connections should be handled by in-world grid nodes automatically and not created directly.

This is something me and @Technici4n talked about earlier, it is for allowing mods depending on AE2 to be able to manually handle grid node connections to GridNodes of GridHost's that does not implement the `IInWorldGridNodeHost`.
In my mod, I am creating cables that should be able to carry multiple grid nodes in one single cable (based on Color), so without this change I will have to use the GridConnection class directly, which is not a part of the AE2 API.
And also, since there is a `createGridConnection` in the GridHelper, I think one should be able to provide a specific Direction for the connection as well.

Technici4n:
Closing due to discussion on discord from a few days ago, i.e. this is too risky to expose in API (sorry!). As long as you know what you're doing, it's ok to use the internal `GridConnection` creation method.

Issue #36, opened 2014-09-27
Extending security to Sub Networks


AlgorithmX2:
With the new implementation of channels, utilizing subnets is pretty much a given, (wich I love BTW) especially if you intend to cover your entire base. Placing a security terminal on each one seems a bit overkill.

What I'm suggesting would be to add a cable similar to quartz fiber that instead of allowing networks to share power would share the security preferences  provided by your main network's security terminal.

Originally Posted by JoelCrosby

This issue was migrated; For original details see http://ae-mod.info/Tracker/0361/

yueh:
The best option is as already mentioned an upgraded quartz fiber, which transmits power and acts as security proxy. Two different ones would be really to tedious.

Not sure of the channel requirements.

TheJulianJES:
:+1: I would an upgraded quartz fiber without channel useage or maybe a config option for channel useage.

myrddraall:
How has this not happened yet. either this need to be implemented, or the requirement for the security terminal to be used for wireless access needs to be removed (or a wireless security term for wireless only added) I make extensive use of sub networks and i have to choose between creating hundreds or security terminals or having wireless access

Technici4n:
This would be nice, but it's not going to happen before a long time I think since I'd need to review the security system entirely to understand how it works and how it can be adapted to this.

Technici4n:
I feel like #6407 makes this a lot less useful / necessary.

PR #6158, opened 2022-04-11
Security Fiber

Mari023:
implement Security Fibers, like mentioned in #36 

TODO
- [ ] recipe
- [ ] textures
- [ ] figure out how to make Security Fibers work when they are loaded in the world


Mari023:
closing since I won't work for the foreseeable future

PR #6164, opened 2022-04-13
Add ID search mode

flier268:


Technici4n (approved):
Not sure about the prefix, seems fine otherwise.

flier268:
圖片

src/main/java/appeng/api/stacks/AEKey.java
```diff
@@ -209,6 +209,8 @@ public final boolean matches(@Nullable GenericStack stack) {
 
     public abstract String getModId();
 
+    public abstract String getId();
```
shartte, line 212: This needs to be documented. What kind of ID is this? Does it include the prefix, or doesn't it? Is it the translation key?

shartte (changes requested):
Please see the comment about documenting what getId actually returns. It looks like the translation key, and *not* the item id.

flier268:
@shartte I had push a commit

shartte:
> @shartte I had push a commit

I am currently out of comission (sick), will get back to this PR once I recover (maybe on the weekend).

flier268:
> > @shartte I had push a commit
> 
> I am currently out of comission (sick), will get back to this PR once I recover (maybe on the weekend).

Take care of you and wish you a speedy recovery

flier268:
> > @shartte I had push a commit
> 
> I am currently out of comission (sick), will get back to this PR once I recover (maybe on the weekend).

Take care of you and wish you a speedy recovery

PR #6171, opened 2022-04-14
Should fix #6088

### LINKED ISSUE #6088: [BUG] Fabric 1.18 having issue with move items button visual
### Describe the bug

When you view a recipe with REI in crafting terminal it should show the missing ingredients, instead it is making the + button yellow tint only if you have a missing ingredient.

### How to reproduce the bug

Make an average disk system with crafting panel
Open the terminal and look up for a random craft that you dont have ingredients (only one or all) 
Instead of + (move items button) turning red and highlighting missing ingredients it just makes + button yellow tint

### Expected behavior

Show what I am missing

### Additional details

unknown-6

1.18

Fabric

### Crash log

https://paste.ee/p/myoae

shedaniel:
The issue is that the old `Context#errorRenderer` method only works if it is a failure, but that has been deprecated into `Context#renderer`

though I changed it to a failure anyways

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -17,7 +17,7 @@ yarn_version=build.1
 # Provided APIs                                         #
 #########################################################
 fabric_version=0.47.9+1.18.2
-rei_version=8.0.442
+rei_version=8.1.444
 wthit_version=4.7.2
 tr_energy_version=2.0.0-beta1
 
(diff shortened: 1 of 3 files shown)


Technici4n (changes requested):
I don't think this is correct. We want to show which items are missing, but still allow the transfer of the existing ingredients. However with this PR doesn't allow partial transfers anymore (previous logic allowed it).

E.g. place only diamonds in an ME cell, and then try to transfer the enchanting table recipe into a crafting terminal.

shartte:
Yeah, we need to be able to show the slot-overlay and button-tooltip while still allowing the transfer.
I think JEI calls this warning, but I may be mistaken.

PR #6172, opened 2022-04-14
Fix #465: add void card

### LINKED ISSUE #465: Void disk feature request
So when using an me system with automation there are often byproducts that you get that you don't want, like cobble from quarries. A void disk would be a special storage disk. You have to configure it's type in a cell bench, but when  you do any of the types on the disk are just destroyed when they are put in the me system.

This would allow people to automate whatever they like and not have to set up more and more exporters depending on the speed of the automation to export into voiding blocks. And it's simple enough that no new mechanics really have to be added.

Technici4n:
While the feature is working code-wise, we need another name, another texture and another recipe. Annihilation cores specifically turn items into energy (M->E in the ME network), however here items are already in energy form. Therefore it doesn't make sense.

Opening this to gather some feedback around name/texture changes, and also because I'll be away for one or two days.


EnderProyects:
Destruction card, made with lava bucket or with lava itself in the molecular assembler, the texture could be a trash bin, and it could also be use in export bus to destroy excess of filtered items trying to be inserted in the thing attached to or in the import bus when they can't be imported but are filtered

skrrt:
Considering it's already a lump of voided material if you think about it, why not craft it with a Matter Ball?

PR #6177, opened 2022-04-15
Fix #5290: Add Fluix tools

### LINKED ISSUE #5290: Add fluix tools
Upgraded certus tools, slighly less durability to diamond tools. 

However with built in Fortune to counter #5282. Most likely just level 1.

62832:
Craftable via the Inscriber from regular nether/certus quartz tools and a fluix crystal, in a similar style to netherite tools via smithing table. All tools have Fortune I applied to them by default, except for the sword, which has Sharpness I applied instead.

Latter functionality required the inscriber recipe serialiser to be adapted in order to support NBT data for output items.


src/main/java/appeng/core/definitions/AEItems.java
```diff
@@ -60,6 +60,7 @@
 import appeng.items.tools.BiometricCardItem;
 import appeng.items.tools.MemoryCardItem;
 import appeng.items.tools.NetworkToolItem;
+import appeng.items.tools.fluix.*;
```
shartte, line 63: I am actually surprised the build didn't barf at you, spotless should not have allowed wildcard imports 🤔

src/main/java/appeng/datagen/providers/recipes/AE2RecipeProvider.java
```diff
@@ -49,6 +49,9 @@ public static JsonObject toJson(ItemStack stack) {
         if (stack.getCount() > 1) {
             stackObj.addProperty("count", stack.getCount());
         }
+        if (stack.hasTag()) {
+            stackObj.addProperty("data", stack.getTag().toString());
+        }
```
shartte, line 54: Does that even work in Vanilla? Just for the output, I presume?

src/main/java/appeng/datagen/providers/recipes/InscriberRecipes.java
```diff
@@ -160,6 +194,10 @@ public Result(String name) {
             @Override
             public void serializeRecipeData(JsonObject json) {
                 json.addProperty("mode", mode.name().toLowerCase(Locale.ROOT));
+                if (fluixTool) {
+                    output.enchant((output.is(AEItems.FLUIX_SWORD.asItem()) ? Enchantments.SHARPNESS
+                            : Enchantments.BLOCK_FORTUNE), 1);
+                }
```
shartte, line 200: Please don't do this in here. Pass the correctly enchanted output to the recipe result instead (just use a static helper to modify the item as needed)

shartte: I am still discussing whether it's appropriate to hand out Fortune 1 for the price of 1 fluix crystal or not on discord though.

src/main/java/appeng/recipes/handlers/InscriberRecipeSerializer.java
```diff
@@ -52,7 +55,15 @@ public InscriberRecipe fromJson(ResourceLocation recipeId, JsonObject json) {
...
-        ItemStack result = ShapedRecipe.itemStackFromJson(GsonHelper.getAsJsonObject(json, "result"));
+        JsonObject resultObj = GsonHelper.getAsJsonObject(json, "result");
+        ItemStack result = new ItemStack(ShapedRecipe.itemFromJson(resultObj));
+        if (resultObj.has("data")) {
+            try {
+                result.setTag(TagParser.parseTag(GsonHelper.getAsString(resultObj, "data")));
+            } catch (CommandSyntaxException e) {
+                throw new JsonSyntaxException("Invalid data tag:" + e);
+            }
+        }
```
shartte, line 66: (See above, I thought there was a way to directly deserialize JSON objects into NBT without going via string)

shartte:
Thanks for your PR! I am still debating on Discord whether the price is too low to get an assured Fortune I or not (also interesting question: how does this interac with common enchantment "strippers"? is this an assured source of fortune 1 they could strip and upgrade up to higher fortune levels by combining the books?).

I also have one question: Wouldn't this also automatically repair the tool durability? Is that something we should prevent in the inscriber? (i.e. by just generally not allowing to inscribe anything in mode PRESS that is damaged).

src/main/java/appeng/core/definitions/AEItems.java
skrrt: I think IntelliJ did that on its own for whatever reason when I got done importing all the new tool classes instead of just leaving them as one per line. I'll sort that out real quick.

src/main/java/appeng/datagen/providers/recipes/AE2RecipeProvider.java
skrrt: This is to save the NBT data _to_ the JSON, however in the meantime it occurred to me that I hadn't checked whether or not this could affect any other AE2 recipes. Might have to do some more on it anyways.

src/main/java/appeng/recipes/handlers/InscriberRecipeSerializer.java
skrrt: Not particularly, and I had to wrestle with that myself for a bit longer than I was expecting.

skrrt:
> I also have one question: Wouldn't this also automatically repair the tool durability? Is that something we should prevent in the inscriber? (i.e. by just generally not allowing to inscribe anything in mode PRESS that is damaged).

For the time being it would. It might be worth trying to replicate the behaviour of normal smithing table recipes somehow wherein existing enchantments, penalties and damage are retained when upgrading diamond tools to netherite.

skrrt:
Made some adaptations as suggested by @shartte along with a first attempt at replicating the behaviour of the smithing table, though for whatever reason this still gets stifled somewhere along the line such that the result forgets any previous damage and existing enchantment data.

Technici4n (changes requested):
I don't like the changes to the inscriber recipe system. I think it should have a way to enable "strict mode" where it will require that the input NBT matches some value exactly, and a way to specify output NBT. This allows arbitrary enchants or changes to the stack NBT. An issue is that the existing `Ingredient`s don't work for that, and fabric doesn't have a custom ingredient API.

skrrt:
I agree that the current approach could be reworked a bit further to support more general NBT data added onto an item as opposed to the currently overly-specific "tool upgrade only" solution. I'll see what I can do with that any time soon.

PR #6184, opened 2022-04-18
Fix #6183: level emitter iterating network contents multiple times

### LINKED ISSUE #6183: Huge Lag When Modifying the Network Structure
**Describe the issue**
In my server, I have created a huge AE network.
However, when I modify the network structure such as placing a fluix cable or removing a fluix cable,  there will be huge delay in the server side.
I have used the spark mod to profile it.

spark log: https://spark.lucko.me/gjVP1Ta3OZ

Is there a way to make the lag smaller?

**Environment**
Fabric 1.18.1 
10.1.0-beta.2

Technici4n:


src/main/java/appeng/parts/automation/StorageLevelEmitterPart.java
```diff
@@ -97,7 +99,11 @@ public void onStackChange(AEKey what, long amount) {
                 lastReportedValue = amount;
                 updateState();
             } else { // either fuzzy upgrade or null filter
-                updateReportingValue(getGridNode().getGrid());
+                long currentTick = TickHandler.instance().getCurrentTick();
```
shartte, line 102:
```suggestion
                // When using a fuzzy upgrade or no filter at all, the level emititer will actively scan the grid
                // We need to ensure we only do this once per tick in case any stack has changed.
                long currentTick = TickHandler.instance().getCurrentTick();
```

Issue #6137, opened 2022-04-05
 System Cycle Restart When Level Emitter Meets Toggle Bus


MidCoard:
### Describe the bug

The environment is the Fabric v10.1.0-beta.2 for minecraft 1.18.1

In one system, when a level emitter aimed at a toggle bus, the system will turn off and turn on again and again.
The video you can see.

And when i use the wrench to remove it, the serve just crashed. The crashed report is above

https://user-images.githubusercontent.com/24197013/161723185-8c740f89-8280-4ad7-8e0c-c3143d841099.mp4

.

### How to reproduce the bug

In one system, put a level emitter aimed at a toggle bus.

### Expected behavior

The system will not turn off and turn on again and again.

### Which minecraft version are you using?

1.18

### On which mod loaders does it happen?

Fabric

mindforger:
in a way that is expected to happen, since the toggle bus extends the network and forces an update, this could end in an endless loop, but what did you want to build there?

MidCoard:
a sub system that can turn on or turn off by a condition that the number of the item is bigger or smaller than a number. But the endless loop is never expected in the earlier version such as minecraft 1.16.5.

mindforger:
never "expected" or "happened" ? i can remember a few issues about wrongly or not updating emitters, maybe one of them broke your design by fixing another issue

this requires somebody that worked on the actual code to answer then

MidCoard:
Anyway, the endless loop is never happened when i use this design before. Thank you for your answer and there are still two more bugs. One is about the texture and the other one made the server crashed.

mindforger:
make new reports for them please bunched up reports are bad to handle

Technici4n:
Might be related to #5613, I'm not sure.

Technici4n:
So, it turns out that the level emitter goes off for a split second since any change in the connections causes a network reboot, which always takes at least 1 tick, and the network is unavailable while it reboots. I would say this is working as expected?

mindforger:
I do too, it might be inconveniant but it is gone for a reason i would say ... Leaving them in their last state might hurt other constructs as well. The workaround i exlained should do the trick in this case

MidCoard:
The level emitter should emit the redstone signal and the toggle bus should receive the signal and make itself pass. So, maybe in its function view, this may be the expected? However, I know what you mean. The two things both trigger the reboot. Redesigning them may cause a huge change of the logic. This may be a feature?

[...]

Issue #5603, opened 2021-10-26
9.0.0-alpha.7 fabric, ME Item Interface freezing


W1nt3rR:
### Describe the bug

ME Item Interface has slots to keep a set amount of items in there. It works until you change something on the network, that is, placing cables or providers or whatever. All interfaces which restock items will stop restocking and stop pulling items in that don't belong there
![2021-10-27_00 25 14](https://user-images.githubusercontent.com/11474127/138969714-eafdd1c4-e8d2-4756-88e8-96d9cc0fbb3e.png)

### How to reproduce the bug

Make a simple Network. Controller, Drive some cables and an interface
![2021-10-27_00 27 45](https://user-images.githubusercontent.com/11474127/138969935-19a44fa7-e028-4c91-8d11-94b8973f3edc.png)

Place an item in the system, and put that same item in an ME Interface to restock a certain amount of items.
![2021-10-27_00 29 08](https://user-images.githubusercontent.com/11474127/138970042-0ef1f794-202a-4362-a9aa-55e714f73616.png)

Place a hopper or some item pipe under the interface, so it pulls out items.
Now, quickly place some cables, or make any other change to the system.
![2021-10-27_00 39 53](https://user-images.githubusercontent.com/11474127/138971075-6ddca753-e909-4eab-adce-2b1eeab3176e.png)

Notice the ME Item Interface no longer restocking items nor taking excess items in the system.
![2021-10-27_00 34 27](https://user-images.githubusercontent.com/11474127/138970562-a87c34c7-8931-467d-8e84-82261edf744a.png)

Updating the Config slots fixes it temporarily.
![2021-10-27_00 41 32](https://user-images.githubusercontent.com/11474127/138971220-d75449b7-6621-4a88-bf3a-00468b55ecb5.png)

### Expected behavior

This creates a lot of issues in systems that rely on restocking items from the network, breaking them.
It Should continue to work even after changing the network in some way, and Not requiring the player to manually update every interface on the network.

### Additional details

This has been tested on All of Fabric 4 modpack aswell.
The pictures above are from my private modpack that I added AE2 to confirm it isn't AOF4 that's the issue, but the mod itself.

### Which minecraft version are you using?

1.17

### On which mod loaders does it happen?

Fabric

### Crash log

https://pastebin.com/Mpi9Dmkz

PR #6191, opened 2022-04-18
Fix #5613 and #6137: network flashing when booting status changes, and level emitter cycle restart (other approach)

### LINKED ISSUE #5613: Ad hoc network flashing
### Describe the bug

https://streamable.com/ppn36n

The left ad hoc network shouldn't be flashing like that.

### How to reproduce the bug

Place a line of cables with a creative cell and multiple terminals. Break a cable between some terminals, and watch as the terminals that remain powered flash for 1 tick.

### Expected behavior

No flashing.

### Additional details

_No response_

1.17

Fabric

### Crash log

n/a

Technici4n:
I changed when the channels are finalized (this is what sends the channel change events to grid node listeners) to be after the grid has finished booting so `isActive` checks properly succeed in there. There are two ways to filter out unwanted notifications while the grid is booting:
- Check `state != State.GRID_BOOTED`: this will only be called if either power or channel state changes. Checking `isActive` there will properly return the new active state. This is appropriate when activeness is only used for display purposes or decided whether some logic should be skipped.
- Check `getMainNode().hasGridBooted()`: this will be called if either the power or channel state changes, and also every time the grid finished booting. This is appropriate when the device was sleeping while booting and might need to be awakened. For example, for ME interfaces, this has to be used to avoid issues such as #5603.

### DIFF
--- a/src/main/java/appeng/parts/storagebus/StorageBusPart.java
+++ b/src/main/java/appeng/parts/storagebus/StorageBusPart.java
@@ -141,11 +141,13 @@ public class StorageBusPart extends UpgradeablePart
 
     @Override
     protected final void onMainNodeStateChanged(IGridNodeListener.State reason) {
-        var currentActive = this.getMainNode().isActive();
-        if (this.wasActive != currentActive) {
-            this.wasActive = currentActive;
-            this.getHost().markForUpdate();
-            remountStorage();
+        if (reason != IGridNodeListener.State.GRID_BOOT) {
+            var currentActive = this.getMainNode().isActive();
+            if (this.wasActive != currentActive) {
+                this.wasActive = currentActive;
+                this.getHost().markForUpdate();
+                remountStorage();
+            }
         }
     }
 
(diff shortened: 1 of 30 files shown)


src/main/java/appeng/parts/storagebus/StorageBusPart.java
```diff
@@ -141,11 +141,13 @@ public StorageBusPart(IPartItem<?> partItem) {
...
-        if (this.wasActive != currentActive) {
-            this.wasActive = currentActive;
-            this.getHost().markForUpdate();
-            remountStorage();
+        if (reason != IGridNodeListener.State.GRID_BOOT) {
+            var currentActive = this.getMainNode().isActive();
+            if (this.wasActive != currentActive) {
+                this.wasActive = currentActive;
+                this.getHost().markForUpdate();
+                remountStorage();
```
Technici4n, line 149: Should storage be unmounted when booting status changes?

shartte: If we unmount storage, would we also unmount drives?

Technici4n: Yeah, I think so.

Technici4n: It would probably cause unwanted level emitter updates though, wouldn't it?

shartte:
Yeah we should not do that I think

Technici4n ***@***.***> schrieb am Fr., 22. Apr. 2022, 13:44:

PR #6195, opened 2022-04-19
fix `ICraftingRequester#insertCraftedItems` doc

DAmNRelentless:
related to https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/6176

I found this when handling the crafting results myself and then returning the unwanted amount as `0` which resulted in duping my items.


Technici4n:
Maybe check for existing mistakes in other places so there doesn't need to be 10 PRs? :P

DAmNRelentless:
Yeah, that would be an idea but I am usually only finding those occasionally when working with them and trying to fix bugs for a few hours and then realizing it's a doc issue. :D

Technici4n:
All insert/extract functions that operate on AEKey and return long always return how much was transferred. ;)

PR #6197, opened 2022-04-20
Fix #6196: Add Inverter Card support to Import Bus

### LINKED ISSUE #6196: Import Bus should accept Inverter Card
### Describe the feature

The Import Bus should accept an Inverter Card to turn the accepted item list into a rejection list.

This request explicitly does _not_ ask for the Export bus to accept the inverter as it would result in performance issues, as discussed in Discord.

An example use case is with tech mods that have machines using fluids such as Tech Reborn's grinder. 
Using an export bus, you can fill the machine with fluid but as soon as you connect an import bus to collect the machine's output, it pulls all the fluid out. I, and other users, immediately made an inverter card to flip the Import Bus's accept list into a reject list to reject the fluid but found it was unsupported. 

### Additional details

This has been brought up before and rejected, but I believe the situation may have changed now as compatibility between mods has increased and with fluids being a part of ME networks.
Some previous issues: https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/2899, https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/5013 and https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/4071

skrrt:
When adding or removing an Inverter Card to the upgrades in an import bus, ~~reset the import strategy to invert the item filter~~ invert the transfer Context used by its import strategy accordingly.

### DIFF
--- a/src/main/java/appeng/api/behaviors/StackImportStrategy.java
+++ b/src/main/java/appeng/api/behaviors/StackImportStrategy.java
@@ -18,7 +18,7 @@ public interface StackImportStrategy {
 
     @FunctionalInterface
     interface Factory {
-        StackImportStrategy create(ServerLevel level, BlockPos fromPos, Direction fromSide, boolean inverted);
+        StackImportStrategy create(ServerLevel level, BlockPos fromPos, Direction fromSide);
     }
 
     static void register(AEKeyType keyType, Factory factory) {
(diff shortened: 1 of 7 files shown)


Wdavery:
Closes #6196

skrrt:
Forgot to note this does indeed have an API-breaking change as brought up on the Discord, adding a new boolean parameter to StackImportStrategy#Factory. This is a similar approach to the current PickupStrategy interface used by the annihilation plane which uses a boolean parameter to enable Silk Touch or not.

PR #6199, opened 2022-04-20
Fix #6198: Cache stack display names to improve terminal sorting performance

### LINKED ISSUE #6198: Client View Sorting by Name Causes Lag
### Describe the feature

I have a big AE network system in my server. 
And when I open the terminal of the system, I sometimes feel the obvious lag, which even makes the client stop.

One time, I use the spark mod to profile it.
This is the log: https://spark.lucko.me/1bwYv1a9Fg

From the log, the client has spent most of the time in the sorting.
I have read some of the code and done some debug work.

It seems like that when the items in the AE system changed, the server will cause all the clients who are in the view to update. Changes to each item will result in an update and sorting. So when the network system is changing all the time, the client will update and sort all the time. This maybe why there will be an obvious lag.

because of 20 ticks in one second, the AE network system can be updated up to 20 times in one second.

And also, I have added some code in the Repo class
image
the sout part to record the sorting time and count.

I have found one thing is that when i select the sorting mode by amount
CCA1F13385845622FE3478445D6C6171
the sorting time is 1ms or 0ms

but when i select the sorting mode by name or others
9998FBFE61039CB8AF247E1139FE7F94
the sorting time is 50ms

because of 20 ticks in one second, each update and sort has spent the whole time, so the client feel the obvious lag.

So I have looked the detailed part of the log and read the name comparator code. 
When name sorting, the getDisplayName method always create a new ItemStack to get its display name(in item part) .
image
the copyTag method and the new ItemStack instance have spent most of the time.

Solution: 

- 10 ticks for one update or it can be edited in the config file. 

- the AEItemKey class should have one field to store its display name to avoid calling the copyTag method and new ItemStack instance frequently.

the environment: fabric  1.18.1 

When there is a big AE network system and we want to sort by name, big client lag really makes the player sad :( 😭

### Additional details

_No response_

MidCoard:
I have created the displayName field in the AEItemKey as a cache. When sorting by name, the cache is used.

before:
9998FBFE61039CB8AF247E1139FE7F94

after:
4E154F2E6D4AC3854D329B650A554380

see details in #6198

### DIFF
--- a/src/main/java/appeng/api/stacks/AEItemKey.java
+++ b/src/main/java/appeng/api/stacks/AEItemKey.java
@@ -25,9 +25,11 @@ public final class AEItemKey extends AEKey {
     @Nullable
     private final CompoundTag tag;
     private final int hashCode;
+    private final Component displayName;
 
     private AEItemKey(Item item, @Nullable CompoundTag tag) {
         this.item = item;
+        this.displayName = toStack().getHoverName();
         this.tag = tag;
         this.hashCode = Objects.hash(item, tag);
     }
@@ -203,7 +205,7 @@ public final class AEItemKey extends AEKey {
 
     @Override
     public Component getDisplayName() {
-        return toStack().getHoverName();
+        return this.displayName;
     }
 
     @Override
--- a/src/main/java/appeng/api/stacks/AEItemKey.java
+++ b/src/main/java/appeng/api/stacks/AEItemKey.java
@@ -1,8 +1,10 @@
 package appeng.api.stacks;
 
-import appeng.api.storage.AEKeyFilter;
-import appeng.core.AELog;
-import appeng.util.Platform;
+import java.util.List;
+import java.util.Objects;
+
+import org.jetbrains.annotations.Nullable;
+
 import net.fabricmc.fabric.api.transfer.v1.item.ItemVariant;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Registry;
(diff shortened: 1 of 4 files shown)


src/main/java/appeng/api/stacks/AEItemKey.java
```diff
@@ -25,9 +25,11 @@ public final class AEItemKey extends AEKey {
     @Nullable
     private final CompoundTag tag;
     private final int hashCode;
+    private final Component displayName;
 
     private AEItemKey(Item item, @Nullable CompoundTag tag) {
         this.item = item;
+        this.displayName = toStack().getHoverName();
```
Technici4n, line 32: Obviously this needs to be moved below `this.tag = tag`. :P

Technici4n (changes requested):
Just one thing. :D

src/main/java/appeng/api/stacks/AEItemKey.java
MidCoard: I made some logical changes.

- First remove AEKey#getDisplayName implementation in AEFluidKey and AEItemKey, implement the method in AEKey. So the sub-class should super one displayName field.

- Second I have reserved no-arg constructor to avoid compatibility problems. The method that implemented this class before must implement the getDisplayName method.No need to worry about null return.

- Third, add Platform#getItemDisplayName to get item display name by itself and its tag. And change Platform#getFluidDisplayName, the argument is replaced by the Fluid instance

src/main/java/appeng/api/stacks/AEKey.java
```diff
@@ -242,7 +263,10 @@ public final boolean supportsFuzzyRangeSearch() {
         return getType().supportsFuzzyRangeSearch();
     }
 
-    public abstract Component getDisplayName();
+    public Component getDisplayName() {
+        assert this.displayName != null;
```
Technici4n, line 267: Please use `Objects.requireNonNull(this.displayName)`, java assertions are useless since they are never enabled in practice.

src/main/java/appeng/api/stacks/AEItemKey.java
Technici4n: Looks good, just one last thing to address.

Technici4n:
Make sure to run `gradlew spotlessApply` for the formatting check to pass.

MidCoard:
I have done formatting check

Technici4n:
Thanks a lot for the PR!

PR #6203, opened 2022-04-24
Fix #6194 by making the ME Annihilation Plane enchantable

### LINKED ISSUE #6194: Fortune Annihilation Plane
### Describe the feature

Just as there is a variant of the Annihilation Plane that uses Silk Touch to break blocks in front of it, it would be useful to have a variant of the plane that uses Fortune to assist in ore processing, especially post-1.18 with the addition of "raw ore" items for all remaining (metal) ore blocks.

Popular demand as per existing discussions on the Discord and suggestions from multiple people there.

### Additional details

_No response_

62832:
Following discussions on the Discord server regarding the addition of a Fortune annihilation plane, it was eventually decided that since the annihilation plane internally makes use of a fake vanilla pickaxe or other digging tool to break blocks, it should simply be enchantable in the same way as that vanilla tool would be.

https://user-images.githubusercontent.com/5098830/164952469-cab2f71a-91f5-4bc7-9ae2-7a74fb88edc5.mp4

https://user-images.githubusercontent.com/5098830/164952477-31191892-c7ca-4626-86cd-3573ab19e42f.mp4

As implemented, this approach allows for immediate support for any enchantments added by other mods which may affect resulting block drops, such as automatic smelting, allowing for specialised processing of items such as glass products. Furthermore, it supports enchantments of an arbitrary level without the need to hard-code anything into the plane as was previously done only for Silk Touch, e.g. a Fortune VII plane in a large modpack with mods like Apotheosis. Hence, the Identity Annihilation Plane has been made redundant and is subsequently ~~removed~~ deprecated by this PR.

As an initial balancing attempt, this comes with an energy multiplier of 8 times the sum of all applied enchantment levels - subject to change depending on balance in large packs, potentially to be switched to an exponential increase depending on specific enchantments.

Further enhancements to be considered include displaying the enchanted sheen on the face of the plane when placed as a cable part and a tooltip to indicate enchantability.

Closes #6194.


src/main/java/appeng/mixins/AnnihilationPlaneEnchantmentMixin.java
```diff
@@ -0,0 +1,20 @@
...
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
+
+import net.minecraft.world.item.Item;
+
+import appeng.core.definitions.AEParts;
+
+@Mixin(targets = "net/minecraft/world/item/enchantment/EnchantmentCategory$7")
```
shartte, line 12: This needs a class-javadoc that explains what the mixin does (why is it needed, basically).
This is important for porting to newer MC versions because no one will remember why this exists :-D

62832: Sorted.

Technici4n (approved):
Need to investigate that mixin issue in prod.

PR #6223, opened 2022-04-30
Fix #6221: Fix several blocks/parts not saving config changes

### LINKED ISSUE #6221: The ConfigManager do not call markDirty()
### Describe the bug

The ConfigManager does not call markDirty/setChanged for the part/block entity meaning it is not saved by default.

### How to reproduce the bug

Place a Pattern Provider
Change its mode to "Blocking Mode"
Log off the world
Log into the world
The mode as reverted back to default.

### Expected behavior

The mode should have been save to blocking mode.

### Additional details

_No response_

1.18

Forge, Fabric

### Crash log

N/A

shartte:
Fixes #6221


Technici4n (changes requested):
Needs to be rebased to account for the new ConfigManager in the pattern access terminal.

PR #6228, opened 2022-04-30
Fix #6210: Render Spatial Anchor overlay down to bedrock again

### LINKED ISSUE #6210: Spatial Anchor Overlay Doesn't properly render.
### Describe the bug

image

The Spatial Anchor Overlay within the overworld, doesn't render down to bedrock.

### How to reproduce the bug

Setup an ME System and Attach a Spatial Anchor, Select to display the overlay.

### Expected behavior

The Highlighted Chunk Markers render down to bedrock.

### Additional details

_No response_

1.18

Forge

### Crash log

Not  A Crash.

62832:
This could probably do with a better approach than simply hard-coding min/max Y-values if ever the cave system gets extended below Y = -64 again.


Technici4n:
Good enough, thanks!

PR #6251, opened 2022-05-06
add keybinds for Wireless (Crafting) Terminal and Portable Storage Cells

### LINKED ISSUE #4968: Add keybind to open Wireless Terminal from anywhere in your inventory much easier.
**Describe the issue**

Simply add a keybind to open to Wireless terminal. I'm suggesting this due to the fact that right now I'm playing a modpack, and the only thing swaying me to Refined Storage, even though I love AE much more, is the fact I know I will never be able to keybind it without the "Wireless Crafting Terminal" mod being updated. I'm amazed it's not be added to the base mod by now.

**Environment**

- Environment: N/A, this is a suggestion

Mari023:
- [ ] some classes are probably not in the right package
- [x] some parts of this should be part of the api
- LocatingService
- LocatingServices#register
- ~~maybe LocatingServices#registerPortableCell~~ this can't be in the api, but should still be accessible for addons
- [x] change the order of so LocatingServices from addons are called before ae2
- [x] the larger portable storage cells need hotkeys too
- [x] localize Keybinds

Closes #4968


Technici4n:
Is this ready for review?

Mari023:
yes

src/main/java/appeng/hotkeys/LocatingServicesImpl.java
```diff
@@ -0,0 +1,59 @@
...
+
+    static {
+        register(
+                new InventoryLocatingService(AEItems.WIRELESS_TERMINAL.asItem(),
+                        (player, i) -> AEItems.WIRELESS_TERMINAL.asItem().openFromInventory(player, i)),
+                "wireless_terminal");
+        register(
+                new InventoryLocatingService(AEItems.WIRELESS_CRAFTING_TERMINAL.asItem(),
+                        (player, i) -> AEItems.WIRELESS_CRAFTING_TERMINAL.asItem().openFromInventory(player, i)),
+                "wireless_crafting_terminal");
```
Technici4n, line 27: Shouldn't this one be under `wireless_terminal`? Since the wireless crafting terminal should just supersede the regular wireless terminal.

src/main/java/appeng/api/hotkeys/LocatingService.java
```diff
@@ -0,0 +1,14 @@
+package appeng.api.hotkeys;
+
+import net.minecraft.world.entity.player.Player;
+
+/**
+ * locates and opens Inventories from items
+ */
+public interface LocatingService {
```
Technici4n, line 8: Maybe we could go for a more general name? As far as I can tell, this can be used for any keybinding that can have multiple handlers. Maybe `HotkeyAction`?

src/main/java/appeng/client/Hotkeys.java
```diff
@@ -0,0 +1,32 @@
...
+import net.minecraft.client.KeyMapping;
+
+import appeng.hotkeys.LocatingServicesImpl;
+
+public class Hotkeys {
+
+    private static final HashMap<String, Hotkey> HOTKEYS = new HashMap<>();
+
+    static {
+        LocatingServicesImpl.REGISTRY.forEach((id, locatingServices) -> registerHotkey(createHotkey(id)));
```
Technici4n, line 17: I don't like this since it is dependent on mods registering their stuff before this class is loaded. Maybe this class could be combined with the hotkey action registry impl, and create the mappings as they are registered.

src/main/java/appeng/hotkeys/InventoryLocatingService.java
```diff
@@ -0,0 +1,44 @@
...
+        this.locatable = locatable;
+        this.opener = opener;
+    }
+
+    public InventoryLocatingService(Item item, Opener opener) {
+        this((stack) -> stack.getItem() == item, opener);
+    }
+
+    @Override
+    public boolean findAndOpen(Player player) {
```
Technici4n, line 24: We don't necessarily need to support them, but I'm curious how this would work with trinkets, if at all? I suspect trinket support could be added by an addon.

src/main/java/appeng/hotkeys/LocatingServicesImpl.java
```diff
@@ -0,0 +1,59 @@
...
+import appeng.api.hotkeys.LocatingService;
+import appeng.api.hotkeys.LocatingServices;
+import appeng.core.definitions.AEItems;
+import appeng.core.definitions.ItemDefinition;
+import appeng.items.tools.powered.PortableCellItem;
+
+/**
+ * registers {@link LocatingService}
+ */
+public class LocatingServicesImpl {
```
Technici4n, line 16: I think it would be nice to expose the ids we use in the API, so that e.g. a portable DISK could be opened with the `portable_item_cell` keybinding.

Technici4n (changes requested):
Quite a few comments, hopefully nothing too bad to fix. Sorry for the delay...

Could you also look into making the API thread-safe?

src/main/java/appeng/hotkeys/InventoryHotkeyAction.java
```diff
@@ -0,0 +1,44 @@
...
+    public InventoryHotkeyAction(Item item, Opener opener) {
+        this((stack) -> stack.getItem() == item, opener);
+    }
+
+    @Override
+    public boolean run(Player player) {
+        var items = player.getInventory().items;
+        for (int i = 0; i < items.size(); i++) {
+            if (locatable.matchItem(items.get(i))) {
+                opener.open(player, i);
```
Technici4n, line 28: Opener returns boolean, but it seems ignored here?

src/main/java/appeng/hotkeys/InventoryHotkeyAction.java
```diff
@@ -0,0 +1,44 @@
...
+
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.Item;
+import net.minecraft.world.item.ItemStack;
+
+import appeng.api.hotkeys.HotkeyAction;
+
+public class InventoryHotkeyAction implements HotkeyAction {
+
+    private final Locatable locatable;
```
Technici4n, line 11: I'd probably use `Predicate<ItemStack>` here.

src/main/java/appeng/hotkeys/HotkeyActionsImpl.java
```diff
@@ -0,0 +1,68 @@
...
+
+        registerPortableCell(AEItems.PORTABLE_FLUID_CELL1K, PORTABLE_FLUID_CELL);
+        registerPortableCell(AEItems.PORTABLE_FLUID_CELL4K, PORTABLE_FLUID_CELL);
+        registerPortableCell(AEItems.PORTABLE_FLUID_CELL16K, PORTABLE_FLUID_CELL);
+        registerPortableCell(AEItems.PORTABLE_FLUID_CELL64K, PORTABLE_FLUID_CELL);
+        registerPortableCell(AEItems.PORTABLE_FLUID_CELL256K, PORTABLE_FLUID_CELL);
+    }
+
+    public static void init() {
+
```
Technici4n, line 45: Would be nice to have a comment such as `// init static` at least.

src/main/java/appeng/hotkeys/HotkeyActionsImpl.java
```diff
@@ -0,0 +1,68 @@
...
+     * a convenience helper for registering hotkeys for portable cells
+     */
+    public static void registerPortableCell(ItemDefinition<PortableCellItem> cell, String id) {
+        register(new InventoryHotkeyAction(cell.asItem(), cell.asItem()::openFromInventory), id);
+    }
+
+    /**
+     * see {@link HotkeyActions#register(HotkeyAction, String)}
+     */
+    public static void register(HotkeyAction hotkeyAction, String id) {
```
Technici4n, line 58: Can you make this synchronized so I don't forget when porting to Forge?

Technici4n (changes requested):
Looks good, very nice and extensible system! Just a few comments and it will be good to go! Sorry for the delay in reviewing this...

src/main/java/appeng/hotkeys/InventoryHotkeyAction.java
Mari023: This is a leftover from a previous iteration

but it could be used to keep checking for more items if the first one fails to open

e.g. when having multiple wireless terminals for different systems at different places it would try opening the second one if the first one isn't in range

Technici4n: Hmmm, shouldn't the open call be `if(open(...)) return true;` then?

Technici4n:
For the package, I'd move this stuff to `appeng.api.features`. Might also want to have the register method in the `HotkeyAction` interface to avoid exposing many small classes.

For localization, you can add the entries manually to the datagen.

Technici4n (approved):
Looks good now, thanks a lot!

PR #6270, opened 2022-05-15
8.4.x 1.16.x

RobloMinerYT:
Pls adds ukrainan translate https://drive.google.com/drive/folders/1-Dn1LMbSVdRbuc_DQZkgX91BbvixsYeB applied energistics 2 8.4.7-1.16.5


Technici4n:
We will only accept new translations for 1.18.2 via crowdin.

PR #6277, opened 2022-05-18
Refactor CraftingUnitType for easier add-on extensibility

62832:
Up until now most add-ons attempting to add higher tiers of crafting storage have had to do so through heavy use of mixins or other hacks due to much of the code that makes use of the CraftingUnitType being fairly hard-coded. This PR should hopefully turn it into something more extensible to make use of more cleanly.


src/main/java/appeng/block/crafting/CraftingUnitType.java
```diff
@@ -0,0 +1,38 @@
...
+        return this == ACCELERATOR;
+    }
+
+    @Override
+    public boolean isStatus() {
+        return this == MONITOR;
+    }
+
+    @Override
+    public boolean isStorage() {
```
Technici4n, line 35: Could be a default implementation in the interface.

src/main/java/appeng/block/crafting/ICraftingUnitType.java
```diff
@@ -0,0 +1,11 @@
+package appeng.block.crafting;
+
+public interface ICraftingUnitType {
+    int getStorageKb();
```
Technici4n, line 4: Isn't this bytes and not kb?

62832:
> Don't you also want to expose the model rendering? That's still using an enum.

This is still a draft specifically for that reason. Once that's  figured out and done I'll mark it as ready.

src/main/java/appeng/block/crafting/ICraftingUnitType.java
62832: Whoops.

62832:
Most of what I wanted to abstract has been abstracted enough now to allow for relatively painless crafting CPU implementation without any use of Mixin whatsoever.

I still have qualms, however, with the approach taken for model rendering — which inexplicably still gives me trouble on my add-on's end when it comes to generating a model and so far causes display issues for textures that aren't the outer ring or the plus-shaped light on top. I especially hope for some suggestions and improvements for the rendering in particular.

src/main/java/appeng/blockentity/crafting/CraftingStorageBlockEntity.java
```diff
@@ -19,60 +19,11 @@
...
 
 import net.minecraft.core.BlockPos;
-import net.minecraft.world.item.Item;
 import net.minecraft.world.level.block.entity.BlockEntityType;
 import net.minecraft.world.level.block.state.BlockState;
 
-import appeng.block.crafting.AbstractCraftingUnitBlock;
-import appeng.core.definitions.AEBlocks;
-
 public class CraftingStorageBlockEntity extends CraftingBlockEntity {
```
Technici4n, line 25: Is this class still necessary?

src/main/java/appeng/blockentity/crafting/CraftingBlockEntity.java
```diff
@@ -216,16 +235,11 @@ public void onMainNodeStateChanged(IGridNodeListener.State reason) {
...
-        return false;
-    }
-
-    public boolean isStorage() {
-        return false;
-    }
-
     public int getStorageBytes() {
-        return 0;
+        if (this.level == null || this.notLoaded() || this.isRemoved()) {
```
Technici4n, line 239: Shouldn't this check be in `getUnitBlock()`?

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -153,7 +153,11 @@ void addBlockEntity(CraftingBlockEntity te) {
         } else if (te.isStatus()) {
```
Technici4n, line 153: Should probably just check for `instanceof CraftingMonitorBlockEntity`, I don't think `isStatus` should be in the interface.

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -153,7 +153,11 @@ void addBlockEntity(CraftingBlockEntity te) {
...
             this.status.add((CraftingMonitorBlockEntity) te);
         } else if (te.isAccelerator()) {
```
Technici4n, line 155: The else here prevents a CPU from acting both as a storage and an accelerator, which seems like an interesting feature for addons?

src/main/java/appeng/client/render/crafting/AbstractCraftingUnitModelProvider.java
```diff
@@ -0,0 +1,34 @@
...
+        this.type = type;
+    }
+
+    /**
+     * @return a list of any materials used for the model to be passed into
+     *         {@link CraftingCubeModel#getAdditionalTextures()}.
+     */
+    public abstract List<Material> getMaterials();
+
+    public abstract TextureAtlasSprite getLightMaterial(Function<Material, TextureAtlasSprite> textureGetter);
```
Technici4n, line 31: This doesn't seem like it needs to be here, it looks like an implementation detail of `CraftingUnitModelProvider`.

src/main/java/appeng/block/crafting/ICraftingUnitType.java
```diff
@@ -0,0 +1,46 @@
...
+     *         with a non-negative storage capacity in bytes is assumed to be a storage block.
+     */
+    default boolean isStorage() {
+        return this.getStorageBytes() > 0;
+    }
+
+    /**
+     * @return whether a given crafting unit type corresponds with a crafting <i>monitor</i>.
+     */
+    boolean isStatus();
```
Technici4n, line 40: I don't think these checks make sense to have there, nobody should be overriding them. The `isStatus` one in particular will probably cause issues if the BE doesn't extend `CraftingStatusBlockEntity` (see my comment in `CraftingCPUCluster`).

src/main/java/appeng/block/crafting/CraftingStorageBlock.java
```diff
@@ -24,7 +24,7 @@
 
 public class CraftingStorageBlock extends AbstractCraftingUnitBlock<CraftingStorageBlockEntity> {
```
Technici4n, line 25: This class can probable be removed.

Technici4n (changes requested):
Looks pretty good! Left a few remarks.

src/main/java/appeng/blockentity/crafting/CraftingStorageBlockEntity.java
62832: This class is what CraftingService#updateCPUClusters looks for to retrieve valid crafting CPUs, so it's probably worth keeping around still for the sake of that distinction.

src/main/java/appeng/block/crafting/CraftingStorageBlock.java
62832: If it comes down to removing CraftingStorageBlockEntity as above then this too shall go.

src/main/java/appeng/blockentity/crafting/CraftingStorageBlockEntity.java
Technici4n: I think it could be made to use `grid.getMachines(CraftingBlockEntity.class)` instead. From the implementation, this should also work for subclasses. Would also need to check that a cluster without storage doesn't assemble, but that should already be checked somewhere else.

62832:
Sorted.

src/main/java/appeng/me/cluster/implementations/CraftingCPUCluster.java
```diff
@@ -147,16 +146,18 @@ void addBlockEntity(CraftingBlockEntity te) {
...
-            this.storage.add(te);
-        } else if (te.isStatus()) {
+        if (te instanceof CraftingMonitorBlockEntity) {
             this.status.add((CraftingMonitorBlockEntity) te);
-        } else if (te.isAccelerator()) {
-            if (te.getAcceleratorThreads() <= 16) {
-                this.accelerator += te.getAcceleratorThreads();
-            } else {
-                throw new IllegalArgumentException("Co-processor threads may not exceed 16 per single unit block.");
+        } else {
```
Technici4n, line 151: Why else? That prevents having a storage monitor that also provides bytes and accelerator threads.

62832: Didn't think that combo made sense, but I guess it could be done lmao

PR #6278, opened 2022-05-19
Expose IAutoCraftingMenu in API, use player's menu in crafting terminal

Technici4n:


mcenderdragon (approved):
looks good - providing a generic object as the owner could be interesting so you can see what is autocrafting right now.

PR #6282, opened 2022-05-20
Fix #6281: Add void and equal distribution upgrade to fluid cells

### LINKED ISSUE #6281: Overflow Destruction Card cannot be added to Fluid Storage Cells
### Describe the bug

Fluid Storage Cells (and by extension, chemical storage cells from Applied Mekanistics) do not support the Overflow Destruction Card. The only card supported by them is the Inverter Card.  

They also do not support the Equal Distribution Card.

### How to reproduce the bug

1. Install AE2
2. Place down a Cell Workbench
3. Put a Fluid Storage Cell (any size) into it
4. Attempt to upgrade with Overflow Destruction Card

### Expected behavior

The cell should support the upgrade card.

### Additional details

_No response_

1.18

Forge

### Crash log

https://pastebin.com/JpMbrtBx

62832:


src/main/java/appeng/init/internal/InitUpgrades.java
```diff
@@ -77,15 +78,24 @@ public static void init() {
             Upgrades.add(AEItems.VOID_CARD, itemCell, 1, storageCellGroup);
         }
 
+        var fluidCells = List.of(
+                AEItems.FLUID_CELL_1K, AEItems.FLUID_CELL_4K, AEItems.FLUID_CELL_16K, AEItems.FLUID_CELL_64K,
+                AEItems.ITEM_CELL_256K);
```
Technici4n, line 83: still got an item cell there, is this intended?

Technici4n (changes requested):
Please also make sure to increase the number of upgrade slots in the config, otherwise it's quite useless. Looks good otherwise. ;)

62832:
Sorted. Fluid cells should have 3 upgrades corresponding to the amount of upgrades that can be applied to them now.

PR #6290, opened 2022-05-21
Fix #6289: Add powered/unpowered states to ME IO and Spatial IO ports

### LINKED ISSUE #6289: Different blockstates for the IO port depending on if it's powered
### Describe the feature

I'm making an emissives resource pack for AE2 and would like to make the little blue lights on the IO port emissive. However then they would also be emissive when the system is powered off, as the IO port only has 1 blockstate unlike something like the controller which has different textures depending on if the system is powered.

Many other ME system components look different depending on whether or not the system is powered

### Additional details

Good

Not so good

62832:
Might need confirmation on whether or not this is enough.


Technici4n (changes requested):
This is not going to work since it I don't see any server -> client syncing. Would it make sense to have a different model for each state to properly test this? (+ for a more logical look)

62832:
Presumably a new model would have some brighter lights on the base texture as a separate overlay for the sake of emissives, but I'd be lying if I said I knew as of now how to go about this.

PR #6296, opened 2022-05-22
Fix accelerated Export Bus Dupe Issues when Network shows items twice

### LINKED ISSUE #6294: [BUG]Copy item Bug!!!!! & the ME interface priority cannot work
### Describe the bug

You can swipe items through the input and output ports
And the me port priority is invalid
——————————————
你可以通过输入输出端口刷取物品
并且me端口优先级失效

### How to reproduce the bug

Add My Telegram and I'll send a video for you
[@](https://t.me/NatsumiXD)

### Expected behavior

priority works fine & Can't get extra items

### Additional details

_No response_

1.18

Fabric

### Crash log

NO

shartte:
Export buses were too trusting of the simulated extraction amounts.
I.e. double chest with two storage buses over-report available items and successfully allow simulated extractions with too many items.

Fixes #6294

### DIFF
--- a/src/main/java/appeng/parts/automation/StorageExportStrategy.java
+++ b/src/main/java/appeng/parts/automation/StorageExportStrategy.java
@@ -1,5 +1,8 @@
 package appeng.parts.automation;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import net.fabricmc.fabric.api.lookup.v1.block.BlockApiCache;
 import net.fabricmc.fabric.api.lookup.v1.block.BlockApiLookup;
 import net.fabricmc.fabric.api.transfer.v1.fluid.FluidStorage;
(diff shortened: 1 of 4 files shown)


src/main/java/appeng/parts/automation/StorageExportStrategy.java
```diff
@@ -55,26 +59,38 @@ public long transfer(StackTransferContext context, AEKey what, long amount, Acti
...
-                if (mode == Actionable.MODULATE) {
-                    StorageHelper.poweredExtraction(
-                            context.getEnergySource(),
-                            inv.getInventory(),
-                            what,
-                            wasInserted,
-                            context.getActionSource(),
-                            Actionable.MODULATE);
-                    tx.commit();
+        if (wasInserted > 0 && mode == Actionable.MODULATE) {
```
Technici4n, line 72: This seems to always be called with `MODULATE` by the way, not sure why we have a mode parameter there in the first place.

Technici4n, line 72: This seems to always be called with `MODULATE` by the way, not sure why we have a mode parameter there in the first place.

Technici4n: I opened a new issue for this, not going to break AppMek just for the sake of it.

PR #6377, opened 2022-06-18
Remove all wildcard imports

62832:


Technici4n:
Would be nice to have some sort of autodetection if you have the time, otherwise happy to merge this.

Maybe have a look at https://github.com/diffplug/spotless/issues/240#issuecomment-385206606?

62832:
> Maybe have a look at [diffplug/spotless#240 (comment)](https://github.com/diffplug/spotless/issues/240#issuecomment-385206606)?

Yeah, that seems to work well enough on my end. I'll throw that in here too.

Issue #6361, opened 2022-06-11
Remove the option to place down the meteorite compass


Technici4n:
It can be surprising/annoying for players, and it's kinda useless anyway. Will also come with a model redesign to make it more aligned with the real compass.

PR #6380, opened 2022-06-19
Certus Quartz Rework

### LINKED ISSUE #5687: Early game change: removal of crystal growth
### Tentative plan for 1.18 early game changes
The goal is to get rid of the current quartz growth mechanic, which is annoying and quite hard to automate correctly, and isn't very fun anymore at this point I think.

- Allow crafting "imperfect" budding certus quartz blocks which behave similarly to vanilla's budding amethyst blocks, except that they can be picked up and moved around, and they have a small chance of breaking after successfully growing a certus quartz crystal. (Maybe 20% chance to degrade, and it has to degrade 3 times to delete itself). A side-effect is that certus quartz is going to be infinite with the proper automation setup.
- Crystal growth accelerators will now accelerate the growth rate of budding amethyst and certus quartz.
- Fluix: I'm not too sure, I need to check if removing fluix dust entirely and replacing it by fluix crystals in the recipes would be possible. If that's ok, in-world crafting of fluix will require crystals and produce crystals.

I'd appreciate feedback or other ideas.

Technici4n:
# Major changes to early game (again)
Fixes #5687.

## Removed certus/fluix seeds and certus quartz ore
Self-explanatory. See below for the replacements.

## Budding certus quartz
Budding certus quartz is a new block that will grow certus quartz crystals, similar to how budding amethyst grows amethyst shards. It exists in the following variants:
- Flawless: will never deplete. Breaking with silk touch gives Flawed, breaking without silk touch gives a regular Certus Quartz Block.
- Flawed: has a 1/12 chance to deplete to Chipped after growing a crystal. Breaking with silk touch gives Chipped, breaking without silk touch gives a regular Certus Quartz Block.
- Chipped: has a 1/12 chance to deplete to Damaged after growing a crystal. Breaking with silk touch gives Damaged, breaking without silk touch gives a regular Certus Quartz Block.
- Damaged: has a 1/12 chance to deplete to regular Certus Quartz Block. Breaking always gives a regular Certus Quartz Block.

**Flawless is a lot easier to automate because it doesn't need replacement (see below), however it can only be moved by using Spatial IO or other modded block moving techniques (e.g. Create contraptions).**

Left to right: Flawless, Flawed, Chipped, Damaged, regular block.
image

## Certus quartz buds
They exist in 3 growth stages. Breaking them with silk touch gives the bud itself, breaking without silk touch gives certus quartz dust.
image

## Certus quartz clusters
The 4th and final growth stage. Breaking them with silk touch gives the cluster itself, breaking without silk touch gives 4 certus quartz crystals, increased by fortune.

## Recharging budding certus quartz
A "depleted" certus quartz block can be recharged using charged certus quartz:
- Certus Quartz Block + Charged Certus Quartz -> Damaged Budding Certus Quartz,
- Damaged Budding Certus Quartz + Charged Certus Quartz -> Chipped Budding Certus Quartz,
- Chipped Quartz Block + Charged Certus Quartz -> Flawed Budding Certus Quartz,
- (Flawless Budding Certus Quartz is not craftable, it can only be found in meteorites).

In order:
image

## Meteorite changes
Every meteorite has a 50% chance to be a fluix meteorite, and a 50% chance to be a certus quartz meteorite.
- Fluix meteorite: the sky stone chest is surrounded by 6 Fluix Blocks (as in 1.17 and 1.18).
- Certus quartz meteorite: the sky stone chest is surrounded by 6 Certus Quartz Blocks (either the regular one, or one of the 4 budding variants, picked randomly).

Since obtaining certus quartz requires finding a meteorite, the meteorite compass is now crafted by charging a regular compass in the Charger.
image

## Crystal growth accelerators
CGAs will now increase the random tick rate of adjacent blocks tagged with `#ae2:growth_acceleratable`. By default this only includes budding amethyst and the 4 budding certus quartz, but packmakers can configure this tag to support any tag. The acceleration rate is configurable.

Example automation setup with 4 budding blocks (requires additional autocrafting to turn the depleted blocks back into their flawless variant):
image

## Turning certus and fluix dust back into crystals
This is now possible using charged certus quartz in water:
image

In-world crafting for fluix now gives fluix crystals (as was the case in 1.16-).


Quezler:
feedback specifically requested here, would usually just post this in discord instead of reacting something so trivial:

the water in the jei preview looks odd, perhaps not "high" enough, or the perspective is just weird 🤔

Silfryi:
Having just six blocks in the larger meteorites feels _very small_ - may I suggest larger meteorites having bigger inner circles? it'd make it feel better, maybe topping out at a 5x5 "sphere" and not just 6 blocks

Shnupbups:
Very nice, it's good that even a classic mod like Applied Energistics can keep with the times and adapt to match and expand on new vanilla mechanics to modernize their gameplay loop and synergise with vanilla.

I would also recommend making the meteorites be essentially a Certus Quartz version of Amethyst geodes on the inside. Would greatly increase the amount of Certus Quartz per meteor, which I think is a good thing as it serves as a jump start before having to grow your own. Would look really neat. Would help tie it visually in with the Amethyst growth mechanic more so people get more of an idea that that's the way things are done now.

As for having a 50/50 of Certus Quartz and Fluix meteorites, I like the idea of having both but feel something's not quite right. I dunno, can't put my finger on it. Regardless, increasing the amount gained from them by making them full geodes would certainly serve to lessen the feeling of annoyance if you can't find the type you're after - because even if you're struggling to find a Fluix one, you got a ton of Certus on the way!

I think a visual redesign of the Meteorite Compass would also be nice, though that may just be stemming from the fact I've never really liked the 3d model look for it anyway. Recent vanilla code changes made as a result of Mojang adding the Recovery Compass now make it much easier to use essentially the same stuff they do, so making it visually more in line with the 2 vanilla compasses (or 3 if you include Lodestone Compasses) would be nice.

I disagree with ricardo in that I think the Silk Touch stuff is perfectly fine. It makes sense both for logic and balance, and really isn't as complicated as seeing it written down would have you believe.

Perhaps beyond the scope here, but with the meteorites now housing the only proper way to obtain your first Certus Quartz, is having them also contain the only way to obtain the presses redundant? Perhaps the presses should now be craftable, instead of found in a chest in the meteorites?
Would solve the issue of having to scour multiple meteorites to find the last press you need, would solve the issue of "where do we put a chest in something that looks like a geode?", and would definitely aid in a multiplayer setting.
Where previously other players would take the presses and go, making them hard for a new player to come by without a kind person donating them a copied press, instead a new player coming across a meteorite discovered by others previously is likely to find at least the Flawless Budding Certus remaining (due to incentive to leave them, like with budding amethyst) and be able to get a bit of Certus to get started (and maybe craft the presses themselves), whilst also not getting the huge boost of extra Certus Quartz that a new meteor would provide, helping to balance it out.
Just an idea, again possible out of scope for now and may not be how you want to take things.

In all, I very much approve of this little overhaul to the crystal growth mechanics. Nice call making the accelerators use a tag to determine what they can accelerate too, very nice.

Sea-Kerman:
I agree with Shnup, making them geode-like seems like a good idea.

Silfryi:
I disagree on the geode part, they're _meteorites_, and they wouldn't keep that same type of hollow structure, and certainly not after impact - I think having no airspace is good. I could see making the presses craftable, but I do still like finding them in the meteorites, I think that mechanic is quite good and fits well. 

I think that instead of dropping certus quartz blocks, they should drop the almost-block (damaged) certus quartz budding blocks unless they're damaged in which case they drop the block without silk touch. This is a good in between, you don't _need_ silk touch, but it's very helpful to have. Another thing you could do is have them drop a number of crystals when you break them without silk - fortune can get more, but 3 for flawless, 2 for flawed, 1 for chipped,

EnderProyects:
Could be implemented that the certus underground meteorites "transform" into a geode hollow one? The chest could appear in the floor of the geode as it would "fall down" ( lore: the certus with help of underground water has eaten the inner walls of the meteorite, transforming it into a geode, with more resources but difficult to find)

ChiefArug:
I agree with 6 blocks per meteor being waaay too few. I like the idea of them being hollow, and who is to say that sky stone isnt strong enough to hold its shape when it impacts?

To fit with how regular geodes are it would require an additional block as a layer between the skystone and crystal, maybe this could be the fluix? That is a lot of fluix though. Would probably result in AE2 adding a second decorative block. (on that note: whats happening to certus slabs, stairs and walls? I vote yeeting them because you cant make amethyst slabs so certus doesnt make much sense to be able to.).

If it stays solid maybe disperse extra blocks (in addition to the 6 around the chest) in the 'shell' of the meteor, which would require excavating the entire thing if you want to find them, or if someone already raided the meteor but didnt grab everything.

Random thought: getting the items out of the chest requires breaking at least one of the blocks, which i dont think is a good idea to encourage (early game players who dont have access to silk touch now won't want to get to the chest or they will waste stuff, or unkowing players will break a flawless one)

I am a little dissapointed that matter balls didnt make it into the budding block recipe at some point, they are a neat item with a unique crafting mechanic that is very underused imo. I think they would fit best in the recipe for the damaged one, as that is what enables people to make infinite budding blocks, without it they have to reuse the ones they found. This would gate a large scale certus farm behind a storage part and condenser (they actually have to start AE2 progression to farm its resources fully, instead of cheesing it with other mods).

I do kinda agree with changing how the presses are found. They have always been a big nuance for a lot of people (see: me spending hours covering a massive area looking for a logic press ([map](https://media.discordapp.net/attachments/855085761786675211/859427737079775252/unknown.png)). Im sure quite a few people can relate). 

If the chest must stay (which i can understand as i think it makes it easier for the compass to locate it), or the lore of finding them as alien technology is to stay then i suggest blank presses that are then combined with their respective resource to make the full press, that means you are garunteed to only need to find one or two meteorites. (This is getting off the topic of meteors but why not overhaul all of early game). Could also be implemented so you have a chance of finding blank presses, or can turn typed presses into blank ones.

Technici4n:
Thanks a lot for your feedback everyone, I greatly appreciate it! From what I understand, the changes seem generally ok, except for the design of meteorites and a few smaller things:

- **Meteorites:** I do indeed need to design them to make them a bit nicer, but I have a hard time understanding how the current generation code works, and modifying it without breaking everything. Something better is definitely planned. I like the idea of having some hollow cavity with some certus quartz and fluix + the chest. (Actually, I am going to remove fluix entirely now that crystals can once again be crafted directly).
- **Silk touch:** you only need one silk touch annihilation plane per automated budding quartz block (to replace the depleted block itself), which I think is fine. Not having silk touch might feel a bit "punishing" when trying to move budding blocks, but I think it's fair enough for now. We might tweak it later based on feedback from the alpha versions.
- **Presses/certus quartz without having to go hunting for meteorites:** this will be solved once we introduce some sort of AE2 villager. I am quite unlikely to implement it myself, but feel free to PR it if that's something that you would want to see.
- **Meteorite compass redesign:** yes this is definitely planned, but the first step is to remove the ability to place down the compass (it's a block at the moment), and then we will adjust the texture. All of this is planned for 1.19, but it won't be in this PR. See also #6361.
- **About in-world transformation:** I like the current recipes; while the idea of using matter balls is good, I think it might be overwhelming to automate for beginners. However, changing the recipes (for example in expert packs) is now possible! Thanks to @Quezler, this PR now introduces the `ae2:transform` recipe type that can be used to change and add in-world water transformation recipes. Here is an example to show how you could use matter balls as an in-world crafting ingredient:
```json
{
  "type": "ae2:transform",
  "ingredients": [
    {
      "item": "ae2:matter_ball"
    },
    {
      "item": "ae2:cell_component_1k"
    }
  ],
  "result": {
    "count": 2,
    "item": "ae2:cell_component_4k"
  }
}
```

Technici4n:
Going to merge this already, the meteorite redesign will come with #6399.

PR #6382, opened 2022-06-20
Transform recipe datapack support goes brrr

Quezler:
Challenged by @Technici4n on the enigmatica discord to implement datapack support for water transformation recipes with no prior knowledge of the ae2 codebase or ever having implemented generators for in java defined recipes before, so ideally you'll leave the reviewing to him since he knows the full context. ^-^ (in case a fellow member/contributor reads this first :O) 

Where do i even begin pitching this pr, i've basically scavenged the codebase and copy pasted stuff that looked like i might need, as well as somewhat follow pre-established structures in order to make it not stand out like a sore thumb too much.

Obviously there are (lots of) improvements to be made, e.g. me not being in-the-know about serializing an ingredient array of items to the network, so i took a page out of the book of `Mindustry` and just write the length first and then send the items one by one, and on the receiving end just get the length and unserialize that amount of items, it seems to work. :tm:

Functions that still return null, dunno what is important when, all i know is that this appears to work for single player.

Without a doubt this will require some finetuning by Technici4n, but at least the bulk of the work has been done for him.

### DIFF
--- a/src/main/java/appeng/datagen/AE2DataGenerators.java
+++ b/src/main/java/appeng/datagen/AE2DataGenerators.java
@@ -38,6 +38,7 @@ import appeng.datagen.providers.recipes.InscriberRecipes;
 import appeng.datagen.providers.recipes.MatterCannonAmmoProvider;
 import appeng.datagen.providers.recipes.SmeltingRecipes;
 import appeng.datagen.providers.recipes.SmithingRecipes;
+import appeng.datagen.providers.recipes.TransformRecipes;
 import appeng.datagen.providers.tags.BiomeTagsProvider;
 import appeng.datagen.providers.tags.BlockTagsProvider;
 import appeng.datagen.providers.tags.FluidTagsProvider;
@@ -78,6 +79,7 @@ public class AE2DataGenerators {
         generator.addProvider(true, new SmeltingRecipes(generator));
         generator.addProvider(true, new CraftingRecipes(generator));
         generator.addProvider(true, new SmithingRecipes(generator));
+        generator.addProvider(true, new TransformRecipes(generator));
 
         // Must run last
         generator.addProvider(true, localization);
--- a/src/main/java/appeng/datagen/providers/recipes/TransformRecipes.java
+++ b/src/main/java/appeng/datagen/providers/recipes/TransformRecipes.java
@@ -1,0 +1,113 @@
+/*
+ * This file is part of Applied Energistics 2.
+ * Copyright (c) 2021, TeamAppliedEnergistics, All rights reserved.
+ *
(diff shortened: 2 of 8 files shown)


Quezler:
Switched from a set to ingredients in the latest commit, only after i was done i realized it was called additionalitems because the charged certus quartz gets added independently because it is the trigger for all recipes, debating on whether or not to rename it back or to possibly even add it to the generated json for technical-correctness sake 🤔

DAmNRelentless:
Hey there! Lazier AE2 dev here.

When I designed the progression of my mod, I wanted to use the transform recipe because I really liked it and so I came up with the resonating materials that are used to craft the mod's ingredients. When implementing said progression, I quickly found out that this is a hardcoded process and was never implemented as an actual recipe type.
I pretty much rebuilt the functionality of the Charged Certus Quartz transformation so I rely on an Item Entity Mixin which could be avoided with an actual recipe type. At least the JEI category was usable externally so I didn't have to rebuild that.

By sharing this experience, I just want to mention that I would highly appreciate this transformation as a recipe type so I am grateful for the pull request. Good job!

Technici4n:
@DAmNRelentless Note that this PR will still require the charged certus quartz to be present regardless, hope that still suits your use case.

DAmNRelentless:
> @DAmNRelentless Note that this PR will still require the charged certus quartz to be present regardless, hope that still suits your use case.

Not really. I should have mentioned that I am referring to the answer from 62832 to make it more add-on friendly.

Technici4n:
Moving to a mixin might make it even easier for pack devs / addon devs to add new recipes. I'll let @Quezler finish what he was doing and then I'll turn it into a mixin I think. 🤔

Quezler:
i didn't really have any more plans other than trying to make the ingredients array an actual array of ingredients, by all means take over the state of the pr 🎁

DAmNRelentless:
I was only able to implement custom transforming because I mixed into the Singularity Item Entity which you also use for the Ender Dust (I didn't question the reason for that) so I didn't have to add my own item entity.

62832:
> I was only able to implement custom transforming because I mixed into the Singularity Item Entity which you also use for the Ender Dust (I didn't question the reason for that) so I didn't have to add my own item entity.

Now I'm intrigued and I've gotten another strange idea to tack onto this.

What if the recipe for Quantum Entangled Singularities were also reworked as such a Transform recipe with some boolean to indicate that an explosion is required? Granted, this would only make sense to implement as a bonus after some sort of generalisation has already been made to dropped item entities used in these recipes (i.e. not requiring strictly charged certus or ender dust or anything like that).

PR #6394, opened 2022-06-24
remove dummy container in MA block entity

DAmNRelentless:
This change allows more generic containers to get remaining items of a Crafting Pattern which makes it more accessible for other inventories, equal to the previous PR https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/6390.

The MA no longer uses the dummy inventory to keep track of the inputs to assemble the recipe output and the remaining items. Thus, several dummy syncing operations were removed.

### DIFF
--- a/src/main/java/appeng/crafting/pattern/AECraftingPattern.java
+++ b/src/main/java/appeng/crafting/pattern/AECraftingPattern.java
@@ -363,8 +363,17 @@ public class AECraftingPattern implements IAEPatternDetails {
         return output;
     }
 
-    public NonNullList<ItemStack> getRemainingItems(CraftingContainer container) {
-        return this.recipe.getRemainingItems(container);
+    public NonNullList<ItemStack> getRemainingItems(Container container) {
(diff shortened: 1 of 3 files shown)


src/main/java/appeng/crafting/pattern/AECraftingPattern.java
```diff
@@ -363,8 +363,17 @@ public ItemStack getOutput(Container container, Level level) {
...
+        var nonNullList = NonNullList.withSize(container.getContainerSize(), ItemStack.EMPTY);
+
+        for(int i = 0; i < nonNullList.size(); ++i) {
+            Item item = container.getItem(i).getItem();
+            if (item.hasCraftingRemainingItem()) {
+                nonNullList.set(i, new ItemStack(item.getCraftingRemainingItem()));
+            }
+        }
+
+        return nonNullList;
```
Technici4n, line 376: No, this is wrong. The entire point of this method is to correctly handle recipes with custom remainder logic. See also 78f5e6d.

Technici4n (changes requested):

PR #6403, opened 2022-06-25
Fixes #6398: Filling a bucket in an ME interface does not produce any sound

### LINKED ISSUE #6398: Filling a bucket in an ME interface does not produce any sound
Literally unplayable.

PiTheGuy:


src/main/java/appeng/items/misc/WrappedGenericStack.java
```diff
@@ -142,6 +151,11 @@ public boolean overrideOtherStackedOnMe(ItemStack itemInSlot, ItemStack otherSta
                 long amount = unwrapAmount(itemInSlot);
                 long inserted = heldContainer.insert(what, amount, Actionable.MODULATE);
 
+                if (what instanceof AEFluidKey && System.currentTimeMillis() > lastFillSound + 50) {
+                    FluidSoundHelper.playFillSound(player, (AEFluidKey) what);
```
Technici4n, line 155: Hi, thanks for the PR!

Looks good, however you should call `playFillSound` on the `heldContainer` to make it work with other key types too (e.g. chemicals in Applied Mekanistics).

Technici4n: One question now that I think of it: under which circumstances is the sound played multiple times? Cause it seems like a bucket would only be filled once per click?

PiTheGuy: When I was testing this, I was taking water from an interface (I also tried lava too), and for some reason, the `overrideStackOnMe` method was running twice when I picked up the water. This might be a separate bug, but I didn't want to mess anything up so I just worked around it.

Technici4n: I think it would be better to figure out why it's called twice; I don't think it's supposed to happen.

Technici4n: Could it be that it's called both on the client and the server? In which case checking if the player is a client player would probably do.

PiTheGuy: You were right. I changed it to a client check and it worked perfectly.

Technici4n: Ok, nice! Two things still: 😅
- Your check is going to crash on the server since `LocalPlayer` doesn't exist there. You should use `player.getLevel().isClientSide()` (or similar).
- You should also remove `FluidSoundHelper.playFillSound(player, (AEFluidKey) what);` since you're already playing the sound via the `heldContainer`.

PiTheGuy: Fixed 😄

src/main/java/appeng/items/misc/WrappedGenericStack.java
```diff
@@ -142,6 +143,10 @@ public boolean overrideOtherStackedOnMe(ItemStack itemInSlot, ItemStack otherSta
                 long amount = unwrapAmount(itemInSlot);
                 long inserted = heldContainer.insert(what, amount, Actionable.MODULATE);
 
+                if (what instanceof AEFluidKey && player.getLevel().isClientSide) {
```
Technici4n, line 146:
```suggestion
                // Check client to avoid duplicate sounds in singleplayer
                if (player.getLevel().isClientSide) {
```

The instanceof check is not necessary anymore, and the comment will help us understand why this check exists if we look at it again in a few months.

PiTheGuy: You're right about the isntanceof not being necessary. I didn't think about that. I also think the comment is a good idea,

PR #6406, opened 2022-06-26
Fix #6400: Formed Quantum Ring hitboxes don't match the model

### LINKED ISSUE #6400: Quantum Ring hitboxes don't match the model
Self-explanatory.

tylimbach:
I started by building from the master branch and creating an instance of minecraft 1.19 to test out the issue.  The hitboxes were fine on an unformed Quantum Ring block, but once it was formed into the correct shape for the ring, both the corner and edge piece hitboxes were too small. Standing on top of the shape, feet sunk into both blocks at differing heights.

After modifying the values for the shape of the formed and corner blocks, I rebuilt the jar and tested it in that same minecraft world. This fixed both blocks, now the player can stand on them and the feet properly rest on the top of each block.

### DIFF
--- a/src/main/java/appeng/block/qnb/QuantumRingBlock.java
+++ b/src/main/java/appeng/block/qnb/QuantumRingBlock.java
@@ -19,6 +19,7 @@
 package appeng.block.qnb;
 
 import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
 import net.minecraft.world.level.BlockGetter;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.level.material.Material;
@@ -33,7 +34,10 @@ public class QuantumRingBlock extends QuantumBaseBlock {
 
     private static final VoxelShape SHAPE = createShape(2.0 / 16.0);
     private static final VoxelShape SHAPE_CORNER = createShape(2.0 / 16.0);
-    private static final VoxelShape SHAPE_FORMED = createShape(0.0 / 16.0);
+    private static final VoxelShape SHAPE_CENTER = createCenterShape(2.0 / 16.0);
+
+    private static final double DEFAULT_MODEL_MIN = 2.0 / 16.0;
+    private static final double DEFAULT_MODEL_MAX = 14.0 / 16.0;
 
     public QuantumRingBlock() {
         super(defaultProps(Material.METAL));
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/block/qnb/QuantumRingBlock.java
```diff
@@ -32,8 +32,8 @@
 public class QuantumRingBlock extends QuantumBaseBlock {
 
     private static final VoxelShape SHAPE = createShape(2.0 / 16.0);
-    private static final VoxelShape SHAPE_CORNER = createShape(4.0 / 16.0);
-    private static final VoxelShape SHAPE_FORMED = createShape(1.0 / 16.0);
+    private static final VoxelShape SHAPE_CORNER = createShape(2.0 / 16.0);
+    private static final VoxelShape SHAPE_FORMED = createShape(0.0 / 16.0);
```
Technici4n, line 36: I think it would be a bit nicer to try to match the exact shape of the block. :)

Technici4n (changes requested):

src/main/java/appeng/block/qnb/QuantumRingBlock.java
tylimbach: Are you specifically referring to the middle blocks where there is a slight offset extruding out from each face, to make the hitbox match the extrusion?

Another side not, I noticed something odd in the QnbFormedBakedModel.class as shown below. The constants for the center and corner blocks are unused. I was wondering if they were intended to be used, because if implemented with those values it makes the corner blocks smaller, making the structure look much more like a ring. The constants from lines 71 and 72 are the only ones used in the whole file, the others are never referenced.

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/8f542e5047f120e8f27c861ec6400408f2cbc1e4/src/main/java/appeng/block/qnb/QnbFormedBakedModel.java#L71-L78

Technici4n: > Are you specifically referring to the middle blocks where there is a slight offset extruding out from each face, to make the hitbox adjust at the extrusion?

Yes

> Another side note, I noticed something odd in the QnbFormedBakedModel.class as shown below. The constants for the center and corner blocks are unused. I was wondering if they were intended to be used, because if implemented with those values it makes the corner blocks smaller, making the structure look much more like a ring. The constants from lines 71 and 72 are the only ones used in the whole file, the others are never referenced.

I will check, but I think it might have been from before the 2020 texture update.

tylimbach: Fixed the hitbox in recent commit. Now properly matches the extrusions on the model. Tested it by walking on top of the formed blocks and standing on the side extrusion of a formed block.

tylimbach (commented):
Tested changes to make sure hitbox matches the exact shape of a formed center block.

Technici4n:
Thank you!

PR #6462, opened 2022-07-27
Remove useless waste of computing power

top-sekret:
Reworked a core code fragment that modified models, which was slow due to three factors:

  1. For some reason it hooked into a getter, which was used multiple times by Balm and in turn invoked the logic that should be invoked once multiple times.
  2. It got a value from a map and then possibly replaced it with another value. This could have been done using compute, saving some map lookups. However, the even better was (that I implemented) was to just do it when the map entries were put for the first time.
  3. It copied a huge map (`src/main/java/appeng/init/client/InitAutoRotatingModel.java` line 99 before the fix).

This reduced the load time in a reproducible environment from 22 minutes to 3:30, which proves its impact.


top-sekret:
Well, I have *considered* the mod (Balm). In my opinion, as this vanilla function is a *getter* and is *public*, I see no reason why calling it multiple times should be wrong. The vanilla implementation just calls return, so other mods probably expect it to be a fast method.

I do not see how the redirect is wrong. Even conceptually it conveys exactly what happens - we redirect the process of model baking to modify the baked models.

Regardless of whether this patch gets accepted, I will continue using it, because the patched hook is conceptually and objectively better.

PR #6493, opened 2022-08-20
Enables the transfer of more than 9 parts of the formulation Ingredients.

vfyjxf:
Before:

dev(2)

After:

dev(1)

PR #6500, opened 2022-08-22
add highlight for craftable ingredients

vfyjxf:
This is a port of the unmerged feature in #6493, which was rejected due to the use of the old api。
JEI:

dev(6)

REI: 

dev(7)


vfyjxf:
Since I mainly want to add this feature in 1.18.2 first, please let me know if you would accept it if I implement it using the new api.

Technici4n:
Please submit your PRs to the `master` branch when applicable (here it's applicable :P ).

I'm fine with a backport to 1.18.2 forge yeah.

Technici4n:
Hmmm, I find the feature a bit weird though now that I actually look at the description. It's strange to highlight craftable outputs imo.

vfyjxf:
> Hmmm, I find the feature a bit weird though now that I actually look at the description. It's strange to highlight craftable outputs imo.

Because I don't think there is any difference between output and input, and of course people are more concerned about whether the input is craftable , maybe I should make some changes?

vfyjxf:
Well I've removed the highlight for the output.

Technici4n:
Ah, I like that feature. However I feel like most of the implementation could perhaps be shared with the other highlight for crafting recipes. If you want I can have a shot at it.

vfyjxf:
>  If you want I can have a shot at it.

Yes, I would like you to do that.

vfyjxf:
You could probably also make this a bit more powerful, such as supporting gas ingredients for mek, but I noticed that ae2 does not provide a generic AEKey comparator.

Technici4n:
Superseded by #6528, thanks for your initial work upon which I based my PR! 🙂

PR #6546, opened 2022-09-11
Fix #6418: Update Tooltip for Crafting/Processing and Substitution Buttons

### LINKED ISSUE #6418: Make Crafting / Processing mode button changing tooltips better
Based from some feedback:

> the button to swap between processing and crafting pattern mode feels backwards to me, it didnt make sense that it was telling me the current mode i was on. i assumed it meant press this button to go to that mode

Shadow-Devil:
Also remove unused Substitutions enum value.


Technici4n:
Perfect, thanks a lot!

PR #6548, opened 2022-09-11
Move IFacadeItem to an api package

agnor99:


agnor99:
As discussed [here(discord link)](https://discord.com/channels/922828438274986004/922829733132124200/1017045374864732180)
Technici4n proposed a facade interoperability between ae2 facades and EnderIO facades (I'm one of the devs and currently putting a lot of effort into conduits) , where both of them could be used to rightclick each other's Blocks and just be applied as if they where originally from that mod. To check, if it's an ae2 facade and getting the Blockstate it displays as, this Interface is needed as api

PR #6561, opened 2022-09-14
add option to properly hide slot via texturepack

Mari023:
currently only works with AppEngSlots that do NOT implement IOptional Slot
(Slots implementing IOptional slot override isSlotEnabled() with their own logic)

### DIFF
--- a/src/main/java/appeng/menu/AEBaseMenu.java
+++ b/src/main/java/appeng/menu/AEBaseMenu.java
@@ -76,6 +76,7 @@ import appeng.util.Platform;
 
 public abstract class AEBaseMenu extends AbstractContainerMenu {
     private static final int MAX_STRING_LENGTH = 32767;
+    private static final String HIDE_SLOT = "HideSlot";
 
     private final IActionSource mySrc;
     @Nullable
@@ -113,6 +114,7 @@ public abstract class AEBaseMenu extends AbstractContainerMenu {
         }
 
         this.mySrc = new PlayerSource(getPlayer(), this.getActionHost());
+        registerClientAction(HIDE_SLOT, String.class, this::hideSlot);
     }
 
     protected final IActionHost getActionHost() {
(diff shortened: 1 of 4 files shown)


src/main/java/appeng/menu/AEBaseMenu.java
```diff
@@ -323,6 +325,17 @@ private boolean isPlayerSideSlot(Slot slot) {
...
+        if (isClientSide()) {
+            sendClientAction(HIDE_SLOT, semantic);
+        }
+        for (Slot s : getSlots(SlotSemantics.getOrThrow(semantic))) {
+            if (s instanceof AppEngSlot slot) {
+                slot.setSlotEnabled(false);
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
Mari023: slots have to be disabled on the server to disable the shift click logic.

src/main/java/appeng/menu/AEBaseMenu.java
```diff
@@ -323,6 +325,22 @@ private boolean isPlayerSideSlot(Slot slot) {
...
+        if (isClientSide()) {
+            sendClientAction(HIDE_SLOT, semantic);
+        }
+        try {
+            for (Slot s : getSlots(SlotSemantics.getOrThrow(semantic))) {
+                if (s instanceof AppEngSlot slot) {
+                    slot.setSlotEnabled(false);
+                }
+            }
+        } catch (IllegalArgumentException ignored) {
```
Technici4n, line 340: That's awful, should be replaced by a proper nullable get.

src/main/java/appeng/menu/SlotSemantics.java
```diff
@@ -31,6 +33,7 @@ private SlotSemantics() {
     }
 
     private static final Map<String, SlotSemantic> REGISTRY = new ConcurrentHashMap<>();
+    public static final Set<String> HIDEABLE = new HashSet<>();
```
Technici4n, line 36: I still think it would be better to make it per-menu, unless that poses an issue to addons ? I wouldn't want to allow a malicious user to hide slots in the wrong menu because another menu is allowing that with the same semantic.

What about such a method in the menu class?
```java
/**
 * Return true if the client is allowed to hide slots with this semantic, false otherwise. You should return false unless you have checked that hiding the slots doesn't risk crashing the game or causing other severe issues.
 */
protected boolean canSlotsBeHidden(SlotSemantic semantic) {
    return false;
}
```

Technici4n (changes requested):

Technici4n (approved):
Looks good to me, I'll ask @shartte to double-check.

PR #6568, opened 2022-09-16
Make mysterious cube emissive and fix some model credits

Sea-Kerman:
change the mysterious cube model to use the same technique for emissives as the controller, and fix some model credits


Sea-Kerman:
ah, forgot to run datagen, how do I fix it?

PR #6634, opened 2022-09-30
Terminal QoL Improvements / Auto-Crafting Notifications

shartte:
- Track ownership of crafting jobs
- Send notifications about owend crafting job status to clients
- Pin items requested via auto-crafting to top-row of terminal
- Unpin after crafting has finished and terminal is closed
- Show toast notifications for completed crafting jobs when the terminal is closed / player has wireless terminal
grafik
grafik
grafik
grafik

### DIFF
--- a/src/main/java/appeng/client/gui/me/common/MEStorageScreen.java
+++ b/src/main/java/appeng/client/gui/me/common/MEStorageScreen.java
@@ -415,7 +415,7 @@ public class MEStorageScreen<C extends MEStorageMenu>
         this.currentMouseY = mouseY;
 
         // Render the pinned row decorations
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/client/gui/me/common/MEStorageScreen.java
```diff
@@ -423,6 +431,23 @@ public void drawFG(PoseStack poseStack, int offsetX, int offsetY, int mouseX,
         }
     }
 
+    private void renderPinnedRowDecorations(PoseStack poseStack) {
```
Technici4n, line 434: The pinned row also appears in the security terminal. 😅

PR #6638, opened 2022-09-30
Allow Memory Cards to Copy Upgrades & Filters

### LINKED ISSUE #1382: [Suggestion] Memory Cards inserting Upgrade-Cards
Found myself restructuring my base recently and make heavy usage of the Memory Card to copy the configurations of different busses. Unfortunately I have to manually insert the different Upgrade-Cards, before pasting the config.
So my suggestion now is, let the Memory Card insert the needed cards into the bus, taking them either  from the inventory or the Network Tool. If there are some missing cards abort the pasting and print some kind of error in chat telling the user what kind of card is missing.

I think this would be a neat little quality of life feature.

### LINKED ISSUE #6633: Memory Card doesn't Store/Restore Filter for Import/Export-Bus/Interfaces

shartte:
Added the ability to memory cards to:
- Save & Restore Upgrade Cards (prioritizes network tool)
- Partially restore settings to incompatible machines

Fixes #1382
Fixes #6633

### DIFF
--- a/src/main/java/appeng/items/tools/MemoryCardItem.java
+++ b/src/main/java/appeng/items/tools/MemoryCardItem.java
@@ -18,23 +18,14 @@
 
 package appeng.items.tools;
 
-import appeng.api.implementations.items.IMemoryCard;
-import appeng.api.implementations.items.MemoryCardMessages;
-import appeng.api.inventories.InternalInventory;
-import appeng.api.upgrades.IUpgradeableObject;
-import appeng.api.util.AEColor;
-import appeng.api.util.IConfigurableObject;
-import appeng.core.AELog;
-import appeng.core.localization.GuiText;
-import appeng.core.localization.PlayerMessages;
-import appeng.core.localization.Tooltips;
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/items/tools/MemoryCardItem.java
```diff
@@ -59,6 +75,200 @@ public MemoryCardItem(Item.Properties properties) {
...
+                    missingAmount -= overflow.getCount();
+                }
+
+                // Try getting them from the network tool or player inventory
+                for (var upgradeSource : upgradeSources) {
+                    var cards = upgradeSource.removeItems(missingAmount, potential, null);
+                    if (!cards.isEmpty()) {
+                        overflow = upgrades.addItems(cards);
+                        if (!overflow.isEmpty()) {
+                            player.getInventory().placeItemBackInInventory(overflow);
```
Technici4n, line 254: Not a fan of this as it silently moves the items around in the inventory afaict.

PR #6676, opened 2022-10-09
Add support for smithing table patterns

shartte:


Kangy1103:
Sorry but can you explain how to set this up please? I've attached an ME Pattern Provider to a smithing table and set the direction on the Pattern Provider. My smithing pattern is valid and inside said Pattern Provider but the table won't accept the items

image

Technici4n:
Crafting patterns require a molecular assembler even though they are made in a crafting table.

The same logic applies to smithing patterns.

Kangy1103:
Thank you! I feel daft for assuming it needed the actual table haha

PR #6689, opened 2022-10-12
fix(i18n): fix a typo in chinese localization

AnzhiZhang:


AnzhiZhang:
Sorry it looks like you are translating things on crowdin. But I cannot find where is that key on crowdin, please take this as a issue and fix it.

PR #6710, opened 2022-10-19
Abstract/refactor portable storage cells

62832:
Ideally accomplishes the following three:
- Allows for easier extensibility and support in add-ons for portable cells that use a custom cell inventory (i.e. not implementing `IBasicCellItem` in cases such as `AEKey`s with only one possible type)
- Allows simply instantiating the existing `StorageTier` record previously only used for portable cells in add-ons looking to add custom tiers
- Allows add-ons to define custom cell states with their own respective LED colours*

*TODO: Change `CellState` from an `enum` to a `record` just like storage tiers.


62832:
Most of what was particularly needed has been accomplished here now. The cell states suggestion will likely be the subject of its own PR later down the line.

PR #6723, opened 2022-10-25
Moved Portable Cells to use StackInteractions in a fully generic way

shartte:


src/main/java/appeng/items/tools/powered/AbstractPortableCell.java
```diff
@@ -194,14 +197,82 @@ public long insert(Player player, ItemStack stack, AEKey what, AEKeyType allowed
...
+        }
+
+        return false;
+    }
+
+    /**
+     * @return The key-type allowed in this cell. Can be null to allow any.
+     */
+    @Nullable
+    protected abstract AEKeyType getKeyType();
```
62832, line 275: Maybe setting this to `null` by default so as to not need to define this for cells with custom transfer behaviour altogether? (e.g. cells with multiple supported types)

62832 (approved):

src/main/java/appeng/items/tools/powered/AbstractPortableCell.java
62832: Fair enough.

PR #6749, opened 2022-11-02
fix crafting job remaining amount not being serialized

### LINKED ISSUE #6702: Crafting link is not marked as done after world restart
### Describe the bug

Leaving and rejoining the world while the system is crafting doesn't mark the crafting link as done when the job is over.

The CPU remains occupied.

### How to reproduce the bug

- have a simple system with a Pattern Provider, an MA, a single crafting CPU, an encoded pattern for planks and some storage
- request 32 planks for the system to craft
- wait until you see a craft popping up in the MA
- leave the world through the escape menu
- rejoin the world
- wait until the craft is over

Result:
- the crafted items will be stored in the system as they should
- opening the crafting CPU will not show ongoing tasks
- a normal terminal shows that one craft is still being performed
- you can't request a new crafting job because no CPUs will be available

I tested it on the latest master branches of Fabric and Forge with the same result.<br>
After debugging `CraftingCpuLogic#insert` as well as `CraftingLink#markDone`, the link is never marked as canceled or done which is likely the cause for this issue.

The only solution to fix this problem is to destroy and replace the crafting storage. At least, that is no dupe bug because the items are already inserted into the system and breaking the CPU will not drop any more items.

### Expected behavior

The crafting link should be marked as done and the task should be cleared which should make the previously used CPU available again. This issue also causes the oak planks to never get unpinned inside the terminal even if you take out all of them.

In my addon, I am relying on the crafting link being properly marked as done or canceled.

### Additional details

https://user-images.githubusercontent.com/16513358/195965203-7156234e-eba5-44e2-b2ad-3f843360f5f1.mp4

1.19

Forge, Fabric

### Crash log

https://pastebin.com/tqaZTC1M

DAmNRelentless:
fixes #6702

### DIFF
--- a/src/main/java/appeng/crafting/execution/ExecutingCraftingJob.java
+++ b/src/main/java/appeng/crafting/execution/ExecutingCraftingJob.java
@@ -142,6 +142,7 @@ public class ExecutingCraftingJob {
         }
         data.put(NBT_TASKS, list);
 
+        data.putLong(NBT_REMAINING_AMOUNT, remainingAmount);
         if (this.playerId != null) {
             data.putInt(NBT_PLAYER_ID, this.playerId);
         }


Technici4n:
what about the
```java
        // TODO 1.20: Remove
        if (data.contains(NBT_REMAINING_AMOUNT)) {
            this.remainingAmount = data.getLong(NBT_REMAINING_AMOUNT);
        } else {
            this.remainingAmount = this.finalOutput.amount();
        }
``` 
though? This shouldn't be needed (TM)

Issue #6732, opened 2022-10-29
ME Storage Bus Bugged with multiple similar unstackable items with same ID inserted


James103:
### Describe the bug

When multiple items of the same ID but different NBT are in a chest, and the chest is connected to an ME System via a Storage Bus, sometimes not all such items are visible in the grid. See the video linked below for more details.

### How to reproduce the bug

Per https://github.com/AllTheMods/ATM-8/issues/183:

> * Create AE2 System with Me Storage Bus attached to any valid container (in this case, a vanilla chest is used, but multiple others have the same issue).
> * Insert 3 (or more) of the same unstackable item with different metadata e.g. 3 iron chestplates, 2 with different enchantments and 1 unenchanted.
> * Observe glitchy behaviour, items disappearing, moving about, reappearing, etc.
> * [This video demonstrates the issue](https://www.youtube.com/watch?v=REpiasIoOKw)

### Expected behavior

All items visible to the ME system and its storage containers, including external ones connected via Storage Buses, should always be visible in the grid, in an order that is consistent across time.

### Additional details

Originally reported in the modpack "All the Mods 8", modpack version 0.1.7.

Can reproduce with just the following mods:
```
client-1.19.2-20220805.130853-srg.jar             |Minecraft                     |minecraft                     |1.19.2              |DONE      |Manifest: a1:d4:5e:04:4f:d3:d6:e0:7b:37:97:cf:77:b0:de:ad:4a:47:ce:8c:96:49:5f:0a:cf:8c:ae:b2:6d:4b:8a:3f
forge-1.19.2-43.1.47-universal.jar                |Forge                         |forge                         |43.1.47             |DONE      |Manifest: 84:ce:76:e8:45:35:e4:0e:63:86:df:47:59:80:0f:67:6c:c1:5f:6e:5f:4d:b3:54:47:1a:9f:7f:ed:5e:f2:90
appliedenergistics2-forge-12.8.3-beta.jar         |Applied Energistics 2         |ae2                           |12.8.3-beta         |DONE      |Manifest: NOSIGNATURE
```

### Which minecraft version are you using?

1.19

### On which mod loaders does it happen?

Forge

### Crash log

No crash (can provide logs if required)

PR #6758, opened 2022-11-07
Fix Storage Bus internal item tree and item name sorting in terminals.

### LINKED ISSUE #6737: Strange display/sort issue with Storage Bus
### Describe the bug

To put it briefly, the terminal is spastic when displaying items in storage containers connected via Storage Bus.

https://streamable.com/7l446j

In the video, there are no items moving into or out of the system. They're just sitting in barrels. I've tried using different sort types to no avail. 

### How to reproduce the bug

1. Setup a basic AE2 system. I set this up without a controller as I didn't need much, just to view items.
2. Attach external storage. 
3. Pump in a bunch of items. Seems to be worse with unstackables, like armor in this video.

I've replicated this by attaching Occultism storage, Functional Drawers (Armory Cabinet), or Sophisticated Storage (bunch of drawers attached by a controller). For what it's worth, these items are all from Apotheosis. I discovered this while testing loot tables by running the followin command on repeat to test the output. `/loot spawn 393.73 62.69 -150.40 loot minecraft:chests/village/village_armorer`

### Expected behavior

The items should remain in a static sort pattern without moving unless items are coming in or out of the system. 

### Additional details

Cannot replicate with disks. So it seems like it's just the Storage Bus. 

1.19

Forge

### Crash log

n/a

berryh:
Fixes #6737 and #6732.

### DIFF
--- a/src/main/java/appeng/api/stacks/FuzzySearch.java
+++ b/src/main/java/appeng/api/stacks/FuzzySearch.java
@@ -6,6 +6,7 @@ import java.util.SortedMap;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Preconditions;
 
+import net.minecraft.nbt.CompoundTag;
 import net.minecraft.world.item.ItemStack;
 
 import it.unimi.dsi.fastutil.objects.Object2ObjectAVLTreeMap;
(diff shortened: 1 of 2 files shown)


Technici4n:
No, outputting the whole NBT to a string is going to be too slow... Another solution will need to be found. Maybe I can have a look tonight.

Also, please submit PRs to the master branch.

berryh:
Fair enough, if you say so. Would this apply to both situations? As in, both to the comparator for the storage map as well as the comparator for item sorting? I've also played with doing it based on tooltip lines (as that's what the player sees so it probably feels the most intuitive) but that probably performs just as poorly then?

Technici4n:
Yeah, in both cases there can be performance issues if 1000s of similar items need to be compared, which is likely to happen in edge cases.

Technici4n:
You can see the fix at #6759, it was a subtle bug.

berryh:
Cool! Also, now more of an FYI then anything: I did some more testing yesterday with the changes (converted to a mixin) in this PR on a server with a fairly large/progressed world and a bunch of non-stacking items in storage buses. Frame-rate on the client-side remains the same as without, but getting items in and out of a terminal and crafting in a crafting terminal gets a bit more sluggish. It's not terrible or unusable, but definitely noticeable. You feel that it takes a little longer to insert or extract an item for example. The server's TPS is slightly affected as well, see the mod view for this Spark report: https://spark.lucko.me/CoPDy611wa

berryh:
Yeah, saw it as well. Don't think it's too bad, but certainly has an impact. I'd expect your fix to perform a lot better.

berryh:
The only thing left is the very odd ordering when using name-based sorting, but I'd say that's not a major problem or priority right now. Also, that does probably need a better performing way then sorting by tooltip lines.

See these tooltips for an example and note the position of the unenchanted Stone Pickaxe.
afbeelding
afbeelding
afbeelding

PR #6800, opened 2022-11-28
Reorganise and make public all custom recipe builders

62832:
Proper follow-up to #6752 to allow for actual use of the recipe builder methods and reorganise the builders themselves into their respective `recipes` packages instead of `datagen`.


Technici4n (approved):

Technici4n (commented):
I'd replace all `Item` parameters by `ItemLike` and call `.asItem()` internally where it's relevant.

62832:
Does that not cause any problems when also used for the output of a recipe rather than just the input?

Technici4n:
Look at what `ItemLike` is and you'll see that there's no problem

62832:
Should I also include any usage of `ItemStack` and replace that with separate item and count params?

PR #6807, opened 2022-12-05
"Lock Crafting" modes for the Pattern Provider

shartte:
Added a new "lock crafting" feature to pattern providers that allows for finer control via redstone or returning the crafted item. Useful in situations where machines to automate expect to be automated with vanilla redstone, for example.


Lordfirespeed:
Relevant documentation has yet to be updated

- [Applied Energistics Website](https://github.com/AppliedEnergistics/appliedenergistics.github.io/blob/source/content/features/auto-crafting.md)
- [Applied Energistics Guidebook](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/master/guidebook/ae2-mechanics/autocrafting.md)

PR #6831, opened 2022-12-15
Fix typos in translation

cubicbyte:
In Russian translation, "256K" components were translated as "16K"


62832:
If I'm not wrong these are supposed to be fixed via AE2's Crowdin page rather than through here.

cubicbyte:
> If I'm not wrong these are supposed to be fixed via AE2's Crowdin page rather than through here.

Oh I didn't know

PR #6842, opened 2022-12-17
Do not connect pattern providers to interfaces on the same network

### LINKED ISSUE #6820: Minor Pattern Provider Improvement
Pattern providers should not push items into interfaces that belong to the same network as the pattern provider itself.
This is already the case for other pattern providers to prevent items from immediately being sent back to the network, but the same issue exists for Pattern Providers adjacent to interfaces:

grafik

62832:
Fixes #6820.

### DIFF
--- a/src/main/java/appeng/helpers/iface/PatternProviderLogic.java
+++ b/src/main/java/appeng/helpers/iface/PatternProviderLogic.java
@@ -69,6 +69,7 @@ import appeng.core.localization.GuiText;
 import appeng.core.localization.PlayerMessages;
 import appeng.core.settings.TickRates;
 import appeng.helpers.ICustomNameObject;
+import appeng.helpers.InterfaceLogicHost;
 import appeng.me.helpers.MachineSource;
 import appeng.util.ConfigManager;
 import appeng.util.inv.AppEngInternalInventory;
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/helpers/iface/PatternProviderLogic.java
```diff
@@ -388,12 +389,14 @@ public GenericStack getUnlockStack() {
...
-        // Skip sides with grid connections to other pattern providers
+        // Skip sides with grid connections to other pattern providers and to interfaces connected to the same network
         var node = mainNode.getNode();
         if (node != null) {
             for (var entry : node.getInWorldConnections().entrySet()) {
                 var otherNode = entry.getValue().getOtherSide(node);
-                if (otherNode.getOwner() instanceof PatternProviderLogicHost) {
+                if (otherNode.getOwner() instanceof PatternProviderLogicHost
+                        || (otherNode.getOwner() instanceof InterfaceLogicHost
+                                && otherNode.getGrid().equals(mainNode.getGrid()))) {
```
Technici4n, line 399: Weird that this check didn't exist for pattern providers? Why shouldn't you be able to push to a pattern provider on a different network?

62832: I can't see myself why not, but I wasn't sure if there was any specific intent to disallowing any and all pattern providers from receiving other providers' inputs.

Technici4n: Hmmm this might have been an intended recent change. @shartte does that ring a bell?

PR #6885, opened 2023-01-07
Intern tags of AEItemKeys

Technici4n:


src/main/java/appeng/api/stacks/AEItemKey.java
```diff
@@ -302,13 +276,37 @@ public boolean equals(Object o) {
...
+                var weakRef = INTERNED.get(searchHolder);
+                InternedTag ret = null;
+
+                if (weakRef != null) {
+                    ret = weakRef.get();
+                }
+
+                if (ret == null) {
+                    // Intern a copy of the tag, since we cannot guarantee that we take ownership
+                    ret = new InternedTag(giveOwnership ? tag : tag.copy());
```
Technici4n, line 304: Can use searchHolder if `giveOwnership` is true

src/main/java/appeng/api/stacks/AEItemKey.java
```diff
@@ -89,7 +88,7 @@ public boolean equals(Object o) {
             return false;
         AEItemKey aeItemKey = (AEItemKey) o;
         // TagHolder instances are interned, compare by reference.
```
Technici4n, line 90: Not TagHolder anymore

PR #6902, opened 2023-01-18
[Forge] make processing pattern push items uncondensed and in order to the target inventory

gurmiguel:
Make Pattern Provider try to push items in order to the target inventory without condensing the ItemStacks first, only then, if there are remaining items from the recipe, it tries to jam items into valid slots.

### DIFF
--- a/src/main/java/appeng/me/storage/ExternalStorageFacade.java
+++ b/src/main/java/appeng/me/storage/ExternalStorageFacade.java
@@ -129,10 +129,17 @@ public abstract class ExternalStorageFacade implements MEStorage {
             int slotCount = handler.getSlots();
             boolean simulate = mode == Actionable.SIMULATE;
 
-            // This uses a brute force approach and tries to jam it in every slot the inventory exposes.
-            for (int i = 0; i < slotCount && !remaining.isEmpty(); i++) {
-                remaining = handler.insertItem(i, remaining, simulate);
-            }
+            int emptySlotsOnlyTries = 1;
+            do {
+                // This uses a brute force approach and tries to jam it in every slot the inventory exposes (first try
+                // to push to empty slots only).
+                for (int i = 0; i < slotCount && !remaining.isEmpty(); i++) {
+                    if (emptySlotsOnlyTries > 0 && !handler.getStackInSlot(i).isEmpty())
+                        continue;
+
+                    remaining = handler.insertItem(i, remaining, simulate);
+                }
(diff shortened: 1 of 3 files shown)


src/main/java/appeng/me/storage/ExternalStorageFacade.java
```diff
@@ -129,10 +129,17 @@ public int insertExternal(AEKey what, int amount, Actionable mode) {
             int slotCount = handler.getSlots();
             boolean simulate = mode == Actionable.SIMULATE;
 
-            // This uses a brute force approach and tries to jam it in every slot the inventory exposes.
-            for (int i = 0; i < slotCount && !remaining.isEmpty(); i++) {
-                remaining = handler.insertItem(i, remaining, simulate);
-            }
+            int emptySlotsOnlyTries = 1;
+            do {
```
Technici4n, line 133: This is a terrible idea, it will cause items to be fragmented in anything connected to a storage bus.

Technici4n:
Thanks for the contribution. The main thing I don't like with this is that separating the `IInput`s at the pattern level will also cause the crafting computation cost to increase a lot. (Each `IInput` needs to be resolved, this can easily combine at multiple levels and cause the size of the crafting tree to increase dramatically)

Technici4n:
This behavior would likely have to be opt-in, or done differently. However I feel like this would increase the configuration complexity of AE2 for a rather niche use case. So that's probably going to be a no from my side, sorry. I'm open to more feedback though.

gurmiguel:
I see, as I said, it's my first try to contribute, so it's kinda hard to know of possible side effects. In the meanwhile, I'll try to figure out some possible solutions to the issues you mentioned, like making it opt-in for each process or something like that.
Thank you for the feedback.

Starship-SpaceX:
The addition of this function can easily solve such problems, please be sure to add this function https://github.com/vectorwing/FarmersDelight/issues/26

Technici4n:
This will not be implemented in that way, this is fundamentally a farmer's delight problem and not an AE2 problem.

I am fine with pushing the items in the original order, I think it's a good idea. However we will NOT prevent them from stacking in the target.

Starship-SpaceX:
Well, the above problems are not unsolvable now, though.  If I rename the item, it prevents the item from stacking in the target container, or I choose to issue a set of items for each empty slot, I just wish there was an easier option that would allow me to directly address this type of issue

Technici4n:
Superseded by #7007. As explained above, this will only work if the target inventory prevents the items from stacking in some way, but there's nothing better I can do.

PR #6923, opened 2023-02-06
Use a different Shader get-method in `Blitter.blit()`

### LINKED ISSUE #6726: Support for Dark mode everywhere
### Describe the feature

Make the ui change color based on dark mode everyhwere settings

because ui be bright

### Additional details

https://www.curseforge.com/minecraft/mc-mods/dark-mode-everywhere

Lordfirespeed:
Closes #6726 

I'm not 100% on what effect this has, however in Vanilla, `getPositionTexShader` returns a `ShaderInstance` created using the same resource location as `getPositionTexColorShader`, the only difference is the "vertex configuration".

With the change implemented:
image

Clearly the widgets aren't perfect but that could be solved using a resourcepack etc.


62832:
That "resource pack" simply requires you to `runData` in your dev environment prior to starting the client.

Lordfirespeed:
> That "resource pack" simply requires you to `runData` in your dev environment prior to starting the client.

Could you elaborate? Sorry, I'm relatively new to the modding scene. 😅 
I'm confused about how that would help since DarkModeEverywhere applies the shaders at runtime, and you can toggle between different shades of light/dark

62832:
Ah, right. Yes, I imagine those could do with a bit of tweaking.

Technici4n:
Interesting! This seems to have some side-effects, however it gives me hints for an approach to fix this. Thank you! I will look further into this once Dark Mode Everywhere is available for 1.19.4.

Lordfirespeed:
> Interesting! This seems to have some side-effects, however it gives me hints for an approach to fix this. Thank you! I will look further into this once Dark Mode Everywhere is available for 1.19.4.

No problem! Glad to be helpful :)

Lordfirespeed:
@Technici4n
> Interesting! This seems to have some side-effects, however it gives me hints for an approach to fix this. Thank you! I will look further into this once Dark Mode Everywhere is available for 1.19.4.

Dark Mode Everywhere has now [released for 1.19.4](https://www.curseforge.com/minecraft/mc-mods/dark-mode-everywhere/files/4524796); here's the CurseMaven dependency (Forge):

```gradle
runtimeOnly fg.deobf("curse.maven:darkmodeeverywhere-574123:4524796")
```

Technici4n:
getPositionTex**Color**Shader allows multiplying the color of each vertex by some RGBA value. Removing it risks making all sorts of things rendering without the proper coloration. However I was keeping this PR open to remember to have a look at this eventually, not to merge it as is. Possibly we just need to add a parameter to how we create some blitters. ;)

@shartte regarding what was mentioned on Discord, I think it would be nicer to integrate with dark mode everywhere than make our own dark mode config that everyone will have to adjust separately.

Lordfirespeed:
@Technici4n Any updates? 
I'm about to start work on refactoring the DarkModeEverywhere shaders, so I should be able to implement a `TexColorPosition` shader once done with that - Would that help on your end?

Technici4n:
No progress on our side as we're busy with other things, but I'm not forgetting this. If the problem fixes itself magically then it's even better. 😄

Lordfirespeed:
Upon (brief) testing, I think the changes I've proposed work out of the box with AE2 on 1.18.2! Can't say for sure yet regarding 1.19 and 1.20 but I see no reason it won't work the same way.

- [1.20](https://github.com/Buuz135/DarkModeEverywhere/pull/41)
- [1.19.4](https://github.com/Buuz135/DarkModeEverywhere/pull/40)
- [1.18.2](https://github.com/Buuz135/DarkModeEverywhere/pull/39)

Took a bit of doing. But I discovered the 3 default shaders that DME was built on previously (Toasted Light, Less Perfect Dark, Perfect Dark) were actually three identical shaders with different parameters.

I parameterized those shaders into one `dark_position_tex` shader and added a new `dark_position_tex_color` shader. With all that done, I added a second mixin that changes the return value from `getPositionTexColorShader` (in addition to the original `getPositionTexShader`). 

That last bit is the key point - because it's the method that AE2 has been using all along!

Technici4n:
As far as I understood this was fixed with some changes in DME.

PR #7001, opened 2023-03-28
Fix #6995: Rework fluid cell type count

### LINKED ISSUE #6995: Rework fluid cell type count
They're way too low at the moment. It also doesn't make sense that portable fluids cells have more types than regular fluid cells.

62832:
Rudimentary fix for now, could do with more consistency for cells later down the line.

### DIFF
--- a/src/main/java/appeng/items/tools/powered/PortableCellItem.java
+++ b/src/main/java/appeng/items/tools/powered/PortableCellItem.java
@@ -47,12 +47,14 @@ public class PortableCellItem extends AbstractPortableCell implements IBasicCell
 
     private final StorageTier tier;
     private final AEKeyType keyType;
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/items/tools/powered/PortableCellItem.java
```diff
@@ -89,7 +89,7 @@ public int getBytesPerType(ItemStack cellItem) {
 
     @Override
     public int getTotalTypes(ItemStack cellItem) {
-        return 63 - this.tier.index() * 9;
+        return this.keyType == AEKeyType.items() ? 63 - this.tier.index() * 9 : 18;
```
Technici4n, line 92: That's a bit gross, would be better as a constructor parameter imo.

PR #7023, opened 2023-04-05
Hide level emitter redstone button when a crafting card is inserted

Sea-Kerman:
The button doesn't do anything if there's a crafting card anyway
ezgif com-gif-maker


Technici4n (changes requested):
That check looks inverted to me?

Technici4n:
Nevermind...

Issue #6864, opened 2022-12-30
Use "Action Bar" for Feedback Messages Instead of Chat


shartte:
Example:

![grafik](https://user-images.githubusercontent.com/1261399/210087069-ef5eda20-3b64-41d7-8105-f03d99e56303.png)

PR #7026, opened 2023-04-07
Use action bar for feedback messages

jpenilla:
Implements #6864


Technici4n (changes requested):
Hmmm, is it not possible to call `player.displayClientMessage(component, true)` for action bar messages?

Technici4n:
Just noticed, is it intended that you're PR'ing to `fabric/1.19.2` and not `master`?

jpenilla:
> Hmmm, is it not possible to call `player.displayClientMessage(component, true)` for action bar messages?

That should work, I just missed it because I looked at `ClientboundSetActionBarTextPacket` and worked backward. Do you want me to call `displayClientMessage` with true at all call sites, or keep the util method to make things more clear without checking what the `true` param means?

> Just noticed, is it intended that you're PR'ing to `fabric/1.19.2` and not `master`?

I had that branch checked out locally and there were conflicts with master :sweat_smile: will rebase shortly

Technici4n:
I think it's best to inline the helper method, otherwise we might have inconsistent code at some point (using the method vs the bool parameter)

Issue #49, opened 2014-09-27
[Wireless Terminal] Viewing signal strenght


AlgorithmX2:
While you are holding a wireless terminal in your hand, you can see the strenght of the signal above your hotbar.

Have a good day =)

Originally Posted by pcnorden

This issue was migrated; For original details see http://ae-mod.info/Tracker/0503/

Deshiba:
I was going to suggest something similar though less invasive. Having the fluix colored indicator turn "on" while in range and "off" while out of range of a WAP. Visible in both the hot-bar and your hand, like the compass/clock and probably several other modded items.

PR #7031, opened 2023-04-08
Add event to find terminal for the "Crafting job completed" toasts

### LINKED ISSUE #6898: "Crafting job completed" toasts do not work if the terminal is in a trinket / curio slot using ae2wtlib
originally reported on the ae2wtlib repo: https://github.com/Mari023/AE2WirelessTerminalLibrary/issues/183
however, this isn't really fixable in ae2wtlib alone.

ae2wtlib would need some way to hook into PendingCraftingJobs#hasNotificationEnablingItem and check for wireless Terminals in trinket / curio slots

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/a0336ccd2f73191319586564df6caa5be2ae13e6/src/main/java/appeng/client/gui/me/common/PendingCraftingJobs.java#L63-L77

Mari023:
Fix #6898

### DIFF
--- a/src/main/java/appeng/client/gui/me/common/PendingCraftingJobs.java
+++ b/src/main/java/appeng/client/gui/me/common/PendingCraftingJobs.java
@@ -1,19 +1,13 @@
 package appeng.client.gui.me.common;
 
-import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.UUID;
-import java.util.function.BiConsumer;
 
 import net.fabricmc.api.EnvType;
 import net.fabricmc.api.Environment;
-import net.fabricmc.fabric.api.event.Event;
-import net.fabricmc.fabric.api.event.EventFactory;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.player.LocalPlayer;
-import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 
 import appeng.api.client.AEKeyRendering;
@@ -22,6 +16,7 @@ import appeng.core.AEConfig;
 import appeng.core.AELog;
 import appeng.core.sync.packets.CraftingJobStatusPacket;
 import appeng.items.tools.powered.WirelessTerminalItem;
+import appeng.util.WirelessTerminalEvent;
 
 /**
  * Tracks pending crafting jobs started by this player.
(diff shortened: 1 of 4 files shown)


src/main/java/appeng/client/gui/me/common/PendingCraftingJobs.java
```diff
@@ -79,4 +86,12 @@ private static boolean hasNotificationEnablingItem(LocalPlayer player) {
 
     record PendingJob(UUID jobId, AEKey what, long requestedAmount, long remainingAmount) {
     }
+
+    public static Event<BiConsumer<List<ItemStack>, Player>> WirelessTerminalEvent = EventFactory.createArrayBacked(
```
Technici4n, line 90: Should be final and have upper case name.

I think this should be a separate class, as I was also planning on using it for #49.

Mari023: I am not quite sure on the name yet, maybe something more like `SearchInventoryEvent`, since it isn't actually specific to Wireless Terminals

src/main/java/appeng/core/AppEngClient.java
```diff
@@ -129,6 +130,7 @@ public AppEngClient() {
 
         INSTANCE = this;
         notifyAddons("client");
+        WirelessTerminalEvent.init();
```
Technici4n, line 133: You can replace this by a static init.

src/main/java/appeng/util/WirelessTerminalEvent.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+import java.util.function.BiConsumer;
+
+import net.fabricmc.fabric.api.event.Event;
+import net.fabricmc.fabric.api.event.EventFactory;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+
+public class WirelessTerminalEvent {
+    public static final Event<BiConsumer<List<ItemStack>, Player>> EVENT = EventFactory.createArrayBacked(
```
Technici4n, line 13: If you don't know what to write for the comment (I assume that's why it's a draft?), here's a suggestion:
```
Event fired when AE2 is looking for potential wireless terminal item stacks in a player inventory.
By default, AE2 only looks at the 36 usual slots of the player inventory, use this event to make AE2 consider more stacks.
The stacks don't need to contain a wireless terminal, AE2 will check after the event.
```

(you'll have to run spotless to format it)

Technici4n:
For the name, `SearchInventoryEvent` works well I think, yeah

src/main/java/appeng/util/WirelessTerminalEvent.java
Mari023: the main reason I converted it to a draft was because I hadn't tested it yet, and loading a world crashed because I forget to run datagen :D

Mari023:
> Is this something we should / want to use for menu hosts too?

Not really, since they need the location of the item

PR #7037, opened 2023-04-10
Fix issues relating to Annihilation Plane checking if it can store a block and then breaking it in two separate ticks (by doing it all in one tick)

jpenilla:
The problem was most obvious when you had a setup with X annihilation planes and 1 formation plane as the only storage; if multiple planes tried to pickup a block in the same tick, one will store it and the rest will drop the item on the ground, as they all succeeded on the simulated insert, but only one will be able to actually insert on the next tick. But the issue could be seen any time the networks ability to store the break result changes between the first and second break process tick (or if two loot table rolls gave different results, and only the first could actually be stored).

### DIFF
--- a/src/main/java/appeng/parts/automation/ItemPickupStrategy.java
+++ b/src/main/java/appeng/parts/automation/ItemPickupStrategy.java
@@ -91,7 +91,7 @@ public class ItemPickupStrategy implements PickupStrategy {
     }
 
     @Override
-    public Result tryStartPickup(IEnergySource energySource, PickupSink sink) {
+    public Result tryPickup(IEnergySource energySource, PickupSink sink) {
         if (this.isAccepting) {
             var blockState = level.getBlockState(pos);
             if (this.canHandleBlock(level, pos, blockState)) {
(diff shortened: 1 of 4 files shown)


src/main/java/appeng/parts/automation/ItemPickupStrategy.java
```diff
@@ -114,29 +114,18 @@ public Result tryStartPickup(IEnergySource energySource, PickupSink sink) {
...
 
         for (var item : items) {
             var inserted = storeItemStack(sink, item);
-            // If inserting the item fully was not possible, drop it as an item entity instead if the storage clears up,
-            // we'll pick it up that way
             if (inserted < item.getCount()) {
-                item.shrink(inserted);
-                Platform.spawnDrops(level, pos, Collections.singletonList(item));
+                // Shouldn't happen unless something messes with state from another thread?
+                throw new IllegalStateException("Failed to insert items after simulated insert succeeded");
```
Technici4n, line 128: Never trust simulation! This should not be changed.

Try picking up a chest that's connected to a storage bus. By the time we get here the chest is gone and the storage bus will reject the insertion. (I think)

Technici4n (changes requested):

PR #7043, opened 2023-04-13
 Allow AddOns to add their own TestPlots

### LINKED ISSUE #7041: Allow AddOns to add their own TestPlots
- [ ] currently, addons can't add their own classes to PLOT_CLASSES
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/31ff9daf96940f2af234df3014577324e409e9cc/src/main/java/appeng/server/testplots/TestPlots.java#L80-L86
- [ ]  it would be nice if addons could also add their items to kitOutPlayer
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/31ff9daf96940f2af234df3014577324e409e9cc/src/main/java/appeng/server/subcommands/SetupTestWorldCommand.java#L90-L96

Mari023:
fix #7041

I am not quite sure about the placement of the event

### DIFF
--- a/src/main/java/appeng/server/testplots/TestPlots.java
+++ b/src/main/java/appeng/server/testplots/TestPlots.java
@@ -17,15 +17,12 @@ import com.google.common.collect.Sets;
 
 import org.jetbrains.annotations.Nullable;
 
-import net.fabricmc.fabric.api.event.Event;
-import net.fabricmc.fabric.api.event.EventFactory;
 import net.minecraft.Util;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
 import net.minecraft.core.Holder;
 import net.minecraft.core.registries.BuiltInRegistries;
 import net.minecraft.resources.ResourceLocation;
-import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.util.RandomSource;
 import net.minecraft.world.entity.EntityType;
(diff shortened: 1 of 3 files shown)


src/main/java/appeng/server/testplots/TestPlots.java
```diff
@@ -77,17 +80,28 @@
...
-            P2PTestPlots.class,
-            MemoryCardTestPlots.class,
-            PatternProviderLockModePlots.class
-    };
+    private static final List<Class<?>> PLOT_CLASSES = new ArrayList<>();
 
     @Nullable
     private static Map<ResourceLocation, Consumer<PlotBuilder>> plots;
 
+    public static final Event<Consumer<ServerPlayer>> KIT_OUT_PLAYER_EVENT = EventFactory.createArrayBacked(
```
Technici4n, line 88: Please put it in a separate file in the same package.

src/main/java/appeng/server/testplots/TestPlots.java
```diff
@@ -146,6 +160,11 @@ private static Map<ResourceLocation, Consumer<PlotBuilder>> scanForPlots() {
         return plots;
     }
 
+    public static void addPlotClass(Class<?> clazz) {
```
Technici4n, line 163:
```suggestion
    public static synchronized void addPlotClass(Class<?> clazz) {
```

PR #7056, opened 2023-04-18
Add sounds to Facade placement/removal based on underlying block

dokastho:
### Summary

Play sounds when placing facades based on the facade's underlying block by invoking `minecraft.level.playLocalSound`. This will ensure the sound is played at the facade's `BlockPos` with the sound specified in the facade's `BlockState`

I think it would be best to move dissasembly sounds to WrenchHook::onPlayerUseBlock and add a general sound for facade or interface disassembly, like the sound an item frame makes when you remove an item from it. A general sound would be a better solution here because it makes sound feedback consistent for all wrenchable items as opposed to just facades.


src/main/java/appeng/items/parts/FacadeItem.java
```diff
@@ -79,6 +82,12 @@ public InteractionResult onItemUseFirst(ItemStack stack, UseOnContext context) {
             }
         }
 
+        BlockPos blockPos = context.getClickedPos();
+        Minecraft minecraft = Minecraft.getInstance();
+        BlockState blockState = this.getTextureBlockState(stack);
+        SoundType soundType = blockState.getSoundType();
+
+        minecraft.level.playLocalSound(blockPos, soundType.getPlaceSound(), SoundSource.BLOCKS, 1.0f, 1.0f, false);
```
Technici4n, line 90: This is not the right place, this should really fire in the code where the facade is placed.

Technici4n: I think inside `placeFacade` makes more sense, after the call to `addFacade`.

Technici4n (changes requested):

src/main/java/appeng/items/parts/FacadeItem.java
dokastho: Thanks, that code has been moved to placeFacade.

src/main/java/appeng/blockentity/networking/CableBusBlockEntity.java
```diff
@@ -346,7 +347,20 @@ public InteractionResult disassembleWithWrench(Player player, Level level, Block
...
+            if (facade != null) {
+                ItemStack masqItem = facade.getTextureItem();
+                BlockItem blockItem = (BlockItem) masqItem.getItem();
+                BlockState blockState = blockItem.getBlock().defaultBlockState();
+                SoundType soundType = blockState.getSoundType();
+
+                this.level.playLocalSound(blockPos, soundType.getBreakSound(), SoundSource.BLOCKS,
+                        1.0f, 1.0f, false);
+            }
+        }
```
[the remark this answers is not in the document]
dokastho: Your first comment is correct. I had misunderstood the purpose of that function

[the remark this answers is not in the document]
dokastho: To correct the issue I think it would be best to move dissasembly sounds to WrenchHook and add a general sound for facade or interface disassembly, like the sound an item frame makes when you remove an item from it.

src/main/java/appeng/hooks/WrenchHook.java
```diff
@@ -57,6 +60,10 @@ public static InteractionResult onPlayerUseBlock(Player player,
                     return InteractionResult.FAIL;
                 }
 
+                // Play a popping sound
```
dokastho, line 63: I think playing the 'Pop' sound that item frames make when you rclick would be appropriate here. Dissasembling with a wrench should make a sound for any given item, and this PR only specifies playing a sound for facades & not terminals, etc and therefore would only play a sound when dissasembling facades, not terminals etc

I'm open to suggestions on which sound should be played when dissasembling

PR #7059, opened 2023-04-19
Harmonize Import/Export Bus Filter Slot Count with Storage Bus

dokastho:
### Summary

Increase default number of filter slots from 1 to 9. This will be done by replacing the OptionalFakeSlot types with FakeSlot types in src/main/java/appeng/menu/implementations/IOBusMenu.java.

Old
image

Current
image
As seen in the image above, the UI needs to display the full 3x3 grid for the filter

image
It also needs to not accept capacity cards

### TODO

- [x] ~increase slot capacity from 1 to 9~
- [ ] ~remove capacity cards from eligible upgrades~
- [ ] ~update UI to reflect 9 slots instead of 1~
- [ ] Change base UI class to the same as the storage bus

---

### DIFF
--- a/src/main/java/appeng/menu/implementations/IOBusMenu.java
+++ b/src/main/java/appeng/menu/implementations/IOBusMenu.java
@@ -23,9 +23,6 @@ import net.minecraft.world.inventory.MenuType;
 
 import appeng.api.config.SecurityPermissions;
 import appeng.client.gui.implementations.IOBusScreen;
-import appeng.menu.SlotSemantics;
-import appeng.menu.slot.FakeSlot;
-import appeng.menu.slot.OptionalFakeSlot;
 import appeng.parts.automation.ExportBusPart;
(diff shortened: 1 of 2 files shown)


Sea-Kerman:
I was somewhat under the impression the plan was to make it like the storage bus, between 9 and 63

dokastho:
I see! I was unsure from the issue which UI was intended to be delivered. Not to worry! I'll update the issue with a note and adjust my changes accordingly. Thanks for your feedback!

Sea-Kerman:
You’d have to confirm with Tech or Gimpansor/Shartte but that’s what I think the intention was, yes.

Technici4n:
Yes the idea was to have the same number of slots as the Storage Bus.

Sea-Kerman:
how goes progress?

Technici4n:
Implemented by #7137.

PR #7062, opened 2023-04-21
Vibration chamber retexture

Sea-Kerman:
Also changes the crafting recipe to match, achievable in survival with just AE2 because you can crank a charger to get the fluix


Sea-Kerman:
Perhaps hold off on this a bit, I want to rethink the top texture with the pistons a bit

PR #7063, opened 2023-04-21
Sort crafting status/plan entries in GUIs

jpenilla:
- Sorts status entries by `active + pending`, then stored.

  - We add active and stored to avoid things jumping around when processing starts and stops, while still keeping the things the system is working on "most actively" in the largest amounts at the top.

- Plan entries are sorted by missing, to craft, and then stored.


jpenilla:
Maybe sorting by `active + pending` and then by stored would be better

Technici4n (approved):
Looks reasonable, I trust you that this is an improvement.

(I'd love a modernization of these screens to include a proper GUI size and search bar, if you feel like it... for a different PR in any case :D )

PR #7064, opened 2023-04-22
change 'useTerminalUseLargeFont' to 'terminalFontSize'

Cjsah:
When `useTerminalUseLargeFont` is set to `true`, the scaling factor of 0.85 is still blurry when the `GUI scale` is 1 or 2.
I changed `useTerminalUseLargeFont` to `terminalFontSize`. it accept a double(converted to float in the code) value
If possible, I can port it to other branches


Technici4n:
Hmmm, please check what's happening on the master branch, that one needs to be prioritized.

PR #7082, opened 2023-05-06
Allow switching between access points when a wireless terminal menu is open

jpenilla:
This allows the menu to stay open if you are moving between the range of 2+ APs while the menu is open, for example if you are in a mine cart, or are flying from one area to another with some momentum as you open the menu.


Technici4n:
Could probably use a comment or two, I'm not sure what this does at a glance.

PR #7083, opened 2023-05-06
Add overflow destruction card support to storage buses

jpenilla:


Technici4n (changes requested):
Please put that logic in the `StorageBusInventory`. It will also fix the `this.handler.getDelegate() instanceof CompositeStorage compositeStorage` check that prevents re-creating the storage every few ticks.

PR #7109, opened 2023-05-16
Allow crafting jobs to be requested from conversion monitors

### LINKED ISSUE #2821: Suggestion Auto Crafting Function with Conversion Monitors
It would be nice to be able to trigger a crafting operation using a Conversion Monitor. On a locked conversion monitor if you have none of the item in your storage but you have a crafting recipe for it instead of displaying "0" it would instead display "Craft", then when you left click, like you would to withdraw items, instead it would open up the crafting GUI.

62832:
Closes #2821
TODO: Figure out why canCraft doesn't update upon removal/reinsertion of patterns

### DIFF
--- a/src/main/java/appeng/api/networking/crafting/ICraftingWatcherNode.java
+++ b/src/main/java/appeng/api/networking/crafting/ICraftingWatcherNode.java
@@ -44,8 +44,9 @@ public interface ICraftingWatcherNode extends IGridNodeService {
     /**
      * Called when a crafting status changes.
      *
-     * @param craftingGrid current crafting grid
-     * @param what         changed key
+     * @param what changed key
      */
-    void onRequestChange(ICraftingService craftingGrid, AEKey what);
+    void onRequestChange(AEKey what);
+
+    void onCraftableChange(AEKey what);
 }
--- a/src/main/java/appeng/api/networking/crafting/ICraftingWatcherNode.java
+++ b/src/main/java/appeng/api/networking/crafting/ICraftingWatcherNode.java
@@ -48,5 +48,10 @@ public interface ICraftingWatcherNode extends IGridNodeService {
      */
     void onRequestChange(AEKey what);
 
+    /**
+     * Called when a craftable status changes.
+     * 
+     * @param what changed key
+     */
     void onCraftableChange(AEKey what);
 }
--- a/src/main/java/appeng/parts/reporting/AbstractMonitorPart.java
+++ b/src/main/java/appeng/parts/reporting/AbstractMonitorPart.java
@@ -35,6 +35,7 @@ import net.minecraft.world.phys.Vec3;
 import appeng.api.behaviors.ContainerItemStrategies;
 import appeng.api.implementations.parts.IStorageMonitorPart;
 import appeng.api.networking.IStackWatcher;
+import appeng.api.networking.crafting.ICraftingService;
 import appeng.api.networking.storage.IStorageService;
 import appeng.api.networking.storage.IStorageWatcherNode;
 import appeng.api.orientation.BlockOrientation;
@@ -63,6 +64,7 @@ public abstract class AbstractMonitorPart extends AbstractDisplayPart
     @Nullable
     private AEKey configuredItem;
     private long amount;
+    private boolean canCraft;
     private String lastHumanReadableText;
     private boolean isLocked;
     private IStackWatcher myWatcher;
@@ -102,6 +104,7 @@ public abstract class AbstractMonitorPart extends AbstractDisplayPart
         if (this.configuredItem != null) {
             AEKey.writeKey(data, this.configuredItem);
             data.writeVarLong(this.amount);
+            data.writeBoolean(this.canCraft);
         }
     }
 
@@ -118,9 +121,11 @@ public abstract class AbstractMonitorPart extends AbstractDisplayPart
         if (data.readBoolean()) {
             this.configuredItem = AEKey.readKey(data);
             this.amount = data.readVarLong();
+            this.canCraft = data.readBoolean();
         } else {
             this.configuredItem = null;
             this.amount = 0;
+            this.canCraft = false;
         }
 
         return needRedraw;
@@ -130,12 +135,14 @@ public abstract class AbstractMonitorPart extends AbstractDisplayPart
     public void writeVisualStateToNBT(CompoundTag data) {
         super.writeVisualStateToNBT(data);
         data.putLong("amount", this.amount);
+        data.putBoolean("canCraft", this.canCraft);
     }
 
     @Override
     public void readVisualStateFromNBT(CompoundTag data) {
         super.readVisualStateFromNBT(data);
         this.amount = data.getLong("amount");
+        this.canCraft = data.getBoolean("canCraft");
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


src/main/java/appeng/api/networking/crafting/ICraftingWatcherNode.java
```diff
@@ -44,8 +44,9 @@ public interface ICraftingWatcherNode extends IGridNodeService {
...
      * Called when a crafting status changes.
      *
-     * @param craftingGrid current crafting grid
-     * @param what         changed key
+     * @param what changed key
      */
-    void onRequestChange(ICraftingService craftingGrid, AEKey what);
+    void onRequestChange(AEKey what);
+
+    void onCraftableChange(AEKey what);
```
Technici4n, line 51: Could use javadoc

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
```diff
@@ -217,6 +239,23 @@ private void extractItem(Player player, int count) {
...
+        if (craftingWatcher != null) {
+            craftingWatcher.reset();
+        }
+
+        if (getDisplayed() != null) {
+            if (craftingWatcher != null) {
+                craftingWatcher.add(getDisplayed());
+            }
+
+            getMainNode().ifPresent(this::updateReportingValue);
```
Technici4n, line 255: This is already done in the super call. You can move the super call to the end. And add a comment saying why we update the crafting watcher first.

src/main/java/appeng/parts/reporting/AbstractMonitorPart.java
```diff
@@ -63,6 +63,7 @@ public abstract class AbstractMonitorPart extends AbstractDisplayPart
     @Nullable
     private AEKey configuredItem;
     private long amount;
+    private boolean canCraft;
```
Technici4n, line 66: This needs to be moved to `ConversionMonitorPart` imo, no need for it in the superclass.

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
```diff
@@ -69,27 +100,28 @@ public boolean onPartActivate(Player player, InteractionHand hand, Vec3 pos) {
             return true;
         }
 
-        if (!this.getMainNode().isActive()) {
```
Technici4n, line 72: Please undo all the formatting changes, they make the diff quite harder to review.

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
```diff
@@ -18,25 +18,37 @@
...
+import appeng.menu.ISubMenu;
+import appeng.menu.locator.MenuLocators;
+import appeng.menu.me.crafting.CraftAmountMenu;
 import appeng.parts.PartModel;
 import appeng.util.InteractionUtil;
 import appeng.util.Platform;
 import appeng.util.inv.PlayerInternalInventory;
 
-public class ConversionMonitorPart extends AbstractMonitorPart {
+public class ConversionMonitorPart extends AbstractMonitorPart implements ITerminalHost {
```
Technici4n, line 51: Why `ITerminalHost`? Would it not be enough to make `CraftAmountMenu` support any `ISubMenuHost` instead?

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
```diff
@@ -186,23 +226,65 @@ private void extractItem(Player player, int count) {
...
+    }
+
     @Override
     public IPartModel getStaticModels() {
         return this.selectModel(MODELS_OFF, MODELS_ON, MODELS_HAS_CHANNEL, MODELS_LOCKED_OFF, MODELS_LOCKED_ON,
                 MODELS_LOCKED_HAS_CHANNEL);
     }
 
+    @Override
+    public void returnToMainMenu(Player player, ISubMenu subMenu) {
```
Technici4n, line 266: Is the menu correctly closed on cancel? Otherwise calling `player.closeContainer` should work.

Technici4n (changes requested):

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
```diff
@@ -187,14 +198,12 @@ private void extractItem(Player player, int count) {
...
-                ItemStack newItems = itemKey.toStack((int) retrieved);
+                var newItems = itemKey.toStack((int) retrieved);
                 if (!player.getInventory().add(newItems)) {
                     player.drop(newItems, false);
                 }
 
-                if (player.containerMenu != null) {
-                    player.containerMenu.broadcastChanges();
-                }
+                player.containerMenu.broadcastChanges();
```
Technici4n, line 206: Why? Doesn't this cause a crash if the player is not in a menu?

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
```diff
@@ -112,6 +118,11 @@ public boolean onClicked(Player player, Vec3 pos) {
         }
 
         if (this.getDisplayed() instanceof AEItemKey itemKey) {
+            if (getAmount() == 0 && canCraft()) {
+                CraftAmountMenu.open((ServerPlayer) player, MenuLocators.forPart(this), itemKey,
+                        itemKey.getAmountPerUnit());
+            }
+
             this.extractItem(player, itemKey.getItem().getMaxStackSize());
```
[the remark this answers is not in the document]
62832: Didn't think it was necessary; if the item isn't craftable, we simply don't open the crafting request menu.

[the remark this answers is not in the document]
Technici4n: The problem is opening the menu and extracting the item. The item might be extractable even if it wasn't reported.

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
62832: The check was supposedly redundant due to `containerMenu` always being non-null at that point.

Technici4n: Why is it always non-null? `extractItem` is called when the player clicks the monitor to take an item, right?

src/main/java/appeng/parts/reporting/ConversionMonitorPart.java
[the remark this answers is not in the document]
62832: Is that not already covered by `extractItem` when the amount is non-zero?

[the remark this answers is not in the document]
Technici4n: I don't see why `extractItem` would check `getAvailableStacks`.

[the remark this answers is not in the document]
62832: Now that I think about it, would it be better to move the menu opening call to *within* `extractItem`?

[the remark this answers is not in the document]
Technici4n: Right, that makes more sense. :+1:

PR #7133, opened 2023-05-19
Remove BasicStatePart

62832:


src/main/java/appeng/parts/crafting/PatternProviderPart.java
```diff
@@ -71,6 +72,7 @@ public class PatternProviderPart extends BasicStatePart implements PatternProvid
 
     public PatternProviderPart(IPartItem<?> partItem) {
         super(partItem);
+        this.getMainNode().setFlags(GridFlags.REQUIRE_CHANNEL);
```
Technici4n, line 75: should be done in the logic probably

src/main/java/appeng/parts/misc/InterfacePart.java
```diff
@@ -77,6 +77,7 @@ public void onGridChanged(InterfacePart nodeOwner, IGridNode node) {
     public InterfacePart(IPartItem<?> partItem) {
         super(partItem);
         this.logic = new InterfaceLogic(this.getMainNode(), this, partItem.asItem());
+        this.getMainNode().setFlags(GridFlags.REQUIRE_CHANNEL);
```
Technici4n, line 80: same

src/main/java/appeng/parts/misc/ToggleBusPart.java
```diff
@@ -81,6 +81,7 @@ public ToggleBusPart(IPartItem<?> partItem) {
 
         this.getMainNode().setIdlePowerUsage(0.0);
         this.getMainNode().setFlags(GridFlags.PREFERRED);
+        this.getMainNode().setFlags(GridFlags.REQUIRE_CHANNEL);
```
Technici4n, line 84: this is wrong, setFlags completely overrides the flags I think. please go in creative mode and double check everything :P

src/main/java/appeng/parts/misc/ToggleBusPart.java
```diff
@@ -80,7 +80,7 @@ public ToggleBusPart(IPartItem<?> partItem) {
         super(partItem);
 
         this.getMainNode().setIdlePowerUsage(0.0);
-        this.getMainNode().setFlags(GridFlags.PREFERRED);
+        this.getMainNode().setFlags(GridFlags.PREFERRED, GridFlags.REQUIRE_CHANNEL);
```
Technici4n, line 83: that should be removed; toggle busses didn't require channels before

PR #7135, opened 2023-05-20
Fix #6621: Make it easier to place parts between cables

### LINKED ISSUE #6621: Make it easier to place parts [especially cable anchors and quartz fibers] between cables
### Describe the feature

Right now, you can only place a part in a direction based on the face of the bounding box you clicked on. In some situations the faces available for this are very small, and in some cases they're nonexistent [e.g. when placing directly between two cables of the same size in a straight line]. My suggestion is when clicking on a cable with a part, if the portion that is clicked is one of the 'arms' of the cable [i.e. not the center cube] and there is no part already in that direction, place the part in that direction instead of in the direction of the face that was clicked.

The arms already have separately outlined bounding boxes even for straight-line cables, and the placement preview shows where the part will go, so this should not be confusing. However, the fact that the placement preview does not appear when it is inside a block could be improved as well.

It can be difficult to break loops or disconnect subnets without entirely removing a cable

62832:
Closes #6621.

### DIFF
--- a/src/main/java/appeng/parts/PartPlacement.java
+++ b/src/main/java/appeng/parts/PartPlacement.java
@@ -11,12 +11,16 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.context.UseOnContext;
 import net.minecraft.world.level.Level;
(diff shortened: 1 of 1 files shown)


Technici4n:
Superseded by #7146.

PR #7137, opened 2023-05-20
Make IO Buses use Storage Bus UI

### LINKED ISSUE #6835: Harmonize Import/Export Bus Filter Slot Count with Storage Bus
There's no real reason that import/export buses have only a single filter slot as opposed to the 9 that the storage bus has by default. Same for formation plane filtering.

The UI should be harmonized, which has the side-effect of greatly increasing the default filter slot count to 9.

ramidzkh:
Closes #6835


Technici4n:
Thanks!

Issue #7042, opened 2023-04-13
Make cables and similar blocks not path-findable


AlexiyOrlov:
### Describe the feature

Mobs should avoid navigating through cables and such

### Reasons why it should be considered

Mobs should be able to move around these blocks without getting stuck. This is especially noticeable on flying mobs

### Additional details

I think you should override `isPathfindable` in Block class and return false

Minebrine66:
Is this issue being worked on? I'd like to try my hand at it if no one else is working on this.

Minebrine66:
So I've implemented the fix as @AlexiyOrlov described and I've done some testing. These are the behaviors I've noticed.

### Without the fix ###
Hostile mobs, such as zombies, will see through the cables and attempt to walk through the cables, only being stopped by the physical cables.
Allays will see a dropped item through the cables and attempt to fly through the cables, only occasionally being stopped by the physical cables due to the thinner cables not being as wide as a normal block.

### With the fix ###
Hostile mobs, such as zombies, will still see through the cables but will navigate around the cables to reach either the player or a villager. 
Allays don't seem to see dropped items through the cables. However, once guided around the cables, allays will pick up the dropped item and navigate around the cables to deliver them to the player.

The cables still function as normal as my fix didn't change how the cables worked. Is this the behavior that you are expecting?

AlexiyOrlov:
I think so

Technici4n:
Sounds good

Technici4n:
Fixed by #7153.

PR #7153, opened 2023-05-25
Prevent mobs from pathfinding through cables

Minebrine66:
Addresses issue [#7042](https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/7042).

### Summary ###
Imported ```pathfinder.PathComputationType``` library and overrode ```isPathfindable``` to ```false```. 

Modifed only ```src/main/java/appeng/block/networking/CableBusBlock.java```. 

### **Testing** ###
Already described in the issue but is reiterated here:

- ### Without the fix ###

Hostile mobs, such as zombies, will see through the cables and attempt to walk through the cables, only being stopped by the physical cables.
Allays will see a dropped item through the cables and attempt to fly through the cables, only occasionally being stopped by the physical cables due to the thinner cables not being as wide as a normal block.

- ### With the fix ###

Hostile mobs, such as zombies, will still see through the cables but will navigate around the cables to reach either the player or a villager.
Allays don't seem to see dropped items through the cables. However, once guided around the cables, allays will pick up the dropped item and navigate around the cables to deliver them to the player.

The cables still function as normal as my fix didn't change how the cables worked.

### DIFF
--- a/src/main/java/appeng/block/networking/CableBusBlock.java
+++ b/src/main/java/appeng/block/networking/CableBusBlock.java
@@ -59,6 +59,7 @@ import net.minecraft.world.level.block.state.properties.BooleanProperty;
 import net.minecraft.world.level.block.state.properties.IntegerProperty;
 import net.minecraft.world.level.material.FluidState;
 import net.minecraft.world.level.material.Fluids;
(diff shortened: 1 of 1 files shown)


Technici4n:
Cool, thank you!

PR #7184, opened 2023-06-05
Add freq tag check QNB restricted slot

### LINKED ISSUE #6922: add warning / error message if a quantum entangled singularity doesn't have a frequency
### Describe the feature

a message should be displayed both in the GUI of the quantum network bridge and in wthit / jade, telling the user the singularity has an invalid frequency

quantum entangled singularity taken from the creative inventory or REI / JEI are missing frequency NBT data. this will cause the quantum network bridge to not work, seemingly without a reason.

### Additional details

_No response_

floridaman:
Added freq tag check for QNB restricted item slot that will prevent the user from slotting in invalid entangled singularities such as ones pulled from the creative mode tab.

I will look into implementing some of the feature requests once I get some more direction. Does AE2 currently not distinguish entangled singularities in any way to the user?

Resolves #6922


Technici4n:
Should also check what happens with hopper insertions. More reasons to have the filter as a static helper method. 😉

Technici4n:
> Does AE2 currently not distinguish entangled singularities in any way to the user?

Do we not print the frequency in the tooltip? In any case it would be nice, and we could also add a red invalid tooltip.

Technici4n:
That's precisely why we should do it... Don't want people to try to use the one for JEI... 😉

Technici4n:
I don't think that's too big of a deal, we just need appropriate wording.

floridaman:
I implemented the inventory similar to how the vibration chamber was done. It works but seems messy. Any input would be appreciated.

src/main/java/appeng/items/materials/EntangledSingularityItem.java
```diff
@@ -0,0 +1,38 @@
...
+    @Override
+    @Environment(EnvType.CLIENT)
+    public void appendHoverText(ItemStack stack, Level level, List<Component> tooltipComponents,
+            TooltipFlag isAdvanced) {
+        super.appendHoverText(stack, level, tooltipComponents, isAdvanced);
+
+        if (hasChannel(stack)) {
+            tooltipComponents.add(GuiText.SerialNumber.text(stack.getTag().get("freq")).withStyle(Tooltips.GREEN));
+        } else {
+            tooltipComponents.add(GuiText.Missing.text("Channel").withStyle(Tooltips.RED));
```
Technici4n, line 31: How can this text be translated? You need to add a new `GuiText` entry, probably `GuiText.MissingChannel`.

PR #7187, opened 2023-06-07
Change terminalMargin from 25 to 0 to display full terminal size

TimmyChips:
In 12.9.1, more terminal sizes were added and vertical margins were added, making the terminal appear shorter than in previous AE2 versions. While you can change the `terminalMargin` in the config client file to `0` to remove these margins, this setting is not obvious for many users and makes the terminals unable to display at full height by default.

The "Small Centered" and "Medium Centered" show the same amount of rows, with both showing only two rows. The Full-Height option is the most drastic but is still unable to fully display the rows. Higher GUI settings make the margins very apparent; setting the GUI Scale to Auto shows only a couple rows for every single setting.

For users sake and parity to earlier versions of AE2, I feel like the `terminalMargin` setting should be set to `0.` The margins are not really needed anyways since you have more control of the height of the terminals with 12.9.1+.


Technici4n:
The margin exists for REI. Additionally, you picked the wrong branches for the PR. And any PR should always be targeted to the `master` branch.

Technici4n:
No problem! I assume you meant to open [an issue](https://github.com/AppliedEnergistics/Applied-Energistics-2/issues)... ;)

PR #7193, opened 2023-06-12
select amount screen issue for fluids

torhayes:


Technici4n:
As I said in the issue:
> Ideally we'd change the length of the text field based on whether it's an item / fluid. Your second picture is good for fluids, for items it would be best to have the full width.

PR #7267, opened 2023-07-16
Prevent Energy Storage Events from being fired multiple times

shartte:


src/main/java/appeng/blockentity/networking/EnergyCellBlockEntity.java
```diff
@@ -116,71 +114,64 @@ private void onAmountChanged() {
...
-            }
-
-            return 0;
-        }
-
-        if (this.internalCurrentPower < 0.01 && amt > 0.01) {
-            this.getMainNode().getNode().getGrid()
-                    .postEvent(new GridPowerStorageStateChanged(this, PowerEventType.PROVIDE_POWER));
+        var inserted = this.stored.insert(amt, mode == Actionable.MODULATE);
+        if (inserted > 0) {
```
Technici4n, line 148: That should only be done when modulating.

src/main/java/appeng/blockentity/networking/EnergyCellBlockEntity.java
```diff
@@ -116,71 +114,64 @@ private void onAmountChanged() {
...
-            this.internalCurrentPower = this.internalMaxPower;
+    @Override
+    public final double extractAEPower(double amt, Actionable mode, PowerMultiplier pm) {
+        return pm.divide(this.extractAEPower(pm.multiply(amt), mode));
+    }
 
+    private double extractAEPower(double amt, Actionable mode) {
+        var extracted = this.stored.extract(amt, mode == Actionable.MODULATE);
+        if (extracted > 0) {
             this.onAmountChanged();
```
Technici4n, line 162: Same here

PR #7297, opened 2023-07-30
Client config option to enable resorting of terminal

Christofmeg:
Added config option to Re-Sort the MEStorageScreen while holding SHIFT. 
This is to address large networks with high item throughput. 
Where pressing SHIFT might cause the player to experience a lot of lag. 
Therefore this option to enable it. (disabled by default)


Technici4n:
The lag was addressed in other ways (#7264), it's just not backported to 1.19.2.

Additionally, you should target `master`.

PR #7299, opened 2023-07-31
Add API method to get cell inventories from IChestOrDrive

62832:
Allows for individual cell contents to be queried directly from chests and drives by add-ons.


src/main/java/appeng/api/implementations/blockentities/IChestOrDrive.java
```diff
@@ -62,4 +63,9 @@ public interface IChestOrDrive extends IActionHost {
     @Nullable
     Item getCellItem(int slot);
 
+    /**
+     * Returns the inventory of the storage cell in the given slot or null.
```
Technici4n, line 67: Note that interacting with the inventory directly will bypass the chest/drive monitoring and thus not properly update LED colors etc. So I'd call the method `getInternalCellInventory` and add this disclaimer to the javadoc.

PR #7361, opened 2023-08-29
port the pattern access terminal features from ExtendedAE

GlodBlock:
- pattern provider in world highlight
image
image
- search input/output/pattern provider name
image
image
image
- highlight matched pattern
image


Mithi83:
This somehow combines two relatively separate features into one PR. In my opinion the features could have been separate PRs, since smaller PRs are easier to review. Maybe it would be a good idea to separate both aspects as it would be more likely to get either of them merged. Since I‘m not a developer of AE2 take my review as a potential user‘s feedback and not a technical review of your code.

The first part (world highlighting of pattern providers) seems useful, but I‘m not sure if this particular approach has the typical AE2 feel to it. I don‘t like the additional icon for highlighting as it is painted over the left border. In my opinion this should be inside the lighter-grey inner box and increase the width of that box. The highlighting in the world itself also seems to be rather unique and not found anywhere else in AE2 (correct me if I‘m wrong). I understand why someone would want to solve the problem of finding a particular PP in the world, but I‘m not sure if that is the best approach to do it.

The second part (searching for input/output/pp name and highlighting the found slot) seems very intuitive, straight forward, and useful. It is something that I‘ve missed in the past as well.

PR #7377, opened 2023-09-03
Add StorageCell::isEmpty method to prevent recursive storage of non-empty cells other than IBasicCellItem

62832:
Provides an API method that dictates whether a cell is empty or not in order for a non-empty cell with a custom `StorageCell` implementation to not be stored in other cells.

There is, however, the caveat that cells that do implement `StorageCell` on their own, if accepting `AEItemKey`s, will need to also implement their own checks for other non-empty cells within their insertion logic each time.

### DIFF
--- a/src/main/java/appeng/api/storage/cells/StorageCell.java
+++ b/src/main/java/appeng/api/storage/cells/StorageCell.java
@@ -40,6 +40,12 @@ public interface StorageCell extends MEStorage {
     double getIdleDrain();
 
     /**
+     * Return whether the cell does not currently contain anything. Use this to determine when your cell should be
+     * storable in or able to store other kinds of cells without recurring.
+     */
+    boolean isEmpty();
+
+    /**
      * Tells the cell to persist to NBT.
      */
     void persist();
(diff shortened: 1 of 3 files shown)


src/main/java/appeng/api/storage/cells/StorageCell.java
```diff
@@ -39,6 +39,12 @@ public interface StorageCell extends MEStorage {
      */
     double getIdleDrain();
 
+    /**
+     * Return whether the cell does not currently contain anything. Use this to determine when your cell should be
+     * storable in or able to store other kinds of cells without recurring.
+     */
+    boolean isEmpty();
```
Technici4n, line 46: This method is badly named, it should be something like `canFitInsideCell`. This is also an API breaking change as-is.

Maybe we should use `canFitInsideContainerItems` instead, but that might not have a stack aware overload (yet).

src/main/java/appeng/api/storage/cells/StorageCell.java
```diff
@@ -39,6 +39,11 @@ public interface StorageCell extends MEStorage {
      */
     double getIdleDrain();
 
+    /**
+     * Return whether the cell should be storable in other kinds of cells.
+     */
+    boolean canFitInsideCell();
```
Technici4n, line 45: Needs a default implementation returning true to preserve backwards compatibility. (This is an API class with foreign implementations).

Technici4n (changes requested):

Technici4n (approved):

PR #7479, opened 2023-11-24
check render before rendering it

GlodBlock:
some TESRs need `shouldRender` to check validity


GlodBlock:
mek's pipes. they do the null check in `shouldRender`

PR #7572, opened 2024-01-24
Curio Integration

### LINKED ISSUE #7525: Native Curio Integration
### Describe the feature

Add Curio support to the Wireless (Crafting) Terminal and Portable Cells

With Neo's Cap rework it is now possible to add Curio support without a compile time dependency on curio.

### Additional details

I am currently working on a PR for this feature.

Open questions:
- [ ] should the Network tool become a curio too?
(since it acts as a storage for Upgrade cards, and adds it's inventory to most ae2 GUIs)
- [ ] which slots should ae2 use?
(ae2wtlib just uses "Curio", which works as a wildcard for any slot)
- [ ] should ae2 add some curio slot on its own? and if yes, which one?
(adding slots can be done by the user with a datapack , or by other mods which do add slots on their own)
(ae2wtlib doesn't add a slot)

Mari023:
closes #7525

Implement Curio Integration without a Required Dependency on Curios, and without a compile time dependency on curios

`MenuItemLocator` has been renamed to `InventoryItemLocator`
added an interface (`MenuItemLocator`) which `InventoryItemLocator` and the new `CurioItemLocator` implement, which has methods for getting and replacing the item

Most Methods that previously accepted an Inventory slot now take a `MenuItemLocator` 

both Portable Cells and Wireless Terminals are now Curios and can be placed in any slot
no slots are added, you need some other mod (or datapack) that adds curio slots to use this


LostQuasar:
Assuming this gets approved I might toss up a pr for trinkets as well

Mari023:
> Assuming this gets approved I might toss up a pr for trinkets as well

trinkets is fabric only, and ae2 isn't doing fabric versions anymore

src/main/java/appeng/menu/locator/MenuLocators.java
```diff
@@ -29,15 +30,15 @@ public final class MenuLocators {
     static {
         register(BlockEntityLocator.class, BlockEntityLocator::writeToPacket, BlockEntityLocator::readFromPacket);
         register(PartLocator.class, PartLocator::writeToPacket, PartLocator::readFromPacket);
-        register(InventoryItemLocator.class, InventoryItemLocator::writeToPacket, InventoryItemLocator::readFromPacket);
-        MenuLocators.register(CuriosItemLocator.class, CuriosItemLocator::writeToPacket,
-                CuriosItemLocator::readFromPacket);
+        register(InventoryItemLocator.class, appeng.menu.locator.InventoryItemLocator::writeToPacket,
+                appeng.menu.locator.InventoryItemLocator::readFromPacket);
+        register(CuriosItemLocator.class, CuriosItemLocator::writeToPacket, CuriosItemLocator::readFromPacket);
```
Mari023, line 35:
```suggestion
        register(InventoryItemLocator.class, InventoryItemLocator::writeToPacket, InventoryItemLocator::readFromPacket);
        register(CuriosItemLocator.class, CuriosItemLocator::writeToPacket, CuriosItemLocator::readFromPacket);
```

src/main/java/appeng/menu/locator/InventoryItemLocator.java
```diff
@@ -17,16 +19,14 @@
  * Optionally also contains a block position and side in case the menu is to be opened by the item but for a clicked
  * host (i.e. network tool).
  */
-record InventoryItemLocator(
-        int itemIndex,
-        @Nullable BlockPos blockPos) implements MenuItemLocator {
+record InventoryItemLocator(int itemIndex, @Nullable BlockHitResult hitResult) implements ItemMenuHostLocator {
     @Nullable
     public <T> T locate(Player player, Class<T> hostInterface) {
```
Mari023, line 24: why did you move the implementation in CuriosItemLocator to the interface, but then keep the nearly identical implementation ere? (the only difference is the logging)

Issue #7399, opened 2023-09-13
Config: Disable Budding Decay


FooterMan15:
### Describe the feature

Add a Config feature for disabling Budding Certus decay.

### Reasons why it should be considered

Better Modpack Customization
Potentially Geode implementation

Sea-Kerman:
If you want that, just kubeJS up a recipe for flawless budding certus (which doesn’t decay).

FooterMan15:
Both

Sea-Kerman:
what exactly is your use case?

PR #7575, opened 2024-01-26
Config option spawn only flawless budding quartz in meteorites 

LostQuasar:
cloeses #7399 

Adds world gen config to spawn only flawless budding quartz in meteorites


LostQuasar:
Alternative solution provided by @Technici4n was to move the block set generation to a tag

PR #7605, opened 2024-02-11
Make energy card shapeless

TurnrDev:


TurnrDev:
I've got a few more of these coming :)

PR #7607, opened 2024-02-11
Fix build badge in readme

TurnrDev:
Can probably go to other branches too


TurnrDev:
Good catch! I updated the branch, and where it links to :)

PR #7619, opened 2024-02-13
make some methods protected instead of private in WirelessTerminalMenuHost

Mari023:
`linkStatus`: I change the link status in ae2wtlib.
`getLinkedGrid()`: I call this in ae2wtlib.
`rangeCheck()`: I run my check if it is linked to a quantum bridge here, and also I need to check this when I don't have a menu, but still need a connection to the me system
`updateLinkStatus()`: I moved the linkStatus calculation to its own method, since this allows me to call it without a menu (calling `onBroadcastChanges()` with a null menu would technically work, but also sometimes drain energy)


Mari023:
superseded by #7635

PR #7652, opened 2024-02-22
use getLinkStatus() in WirelessTerminalMenuHost#insert

Mari023:
when inserting in the terminal item using right click, use getLinkStatus() instead of the variable directly, so that the correct value is obtained when addons override getLinkStatus()


Technici4n (approved):

Technici4n:
Does this fix https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/7655?

PR #7666, opened 2024-02-24
Fix handling of container items that can contain multiple key types

Technici4n:
Waiting to be tested with gauge droppers once they are fixed in Mekanism.

Fixes https://github.com/AppliedEnergistics/Applied-Mekanistics/issues/25.


Technici4n:
Tested, it works!

PR #7677, opened 2024-02-25
Fix issue with energy source in terminals after terminal changes

### LINKED ISSUE #7676: crafting terminal can't craft
### Describe the bug

crafting an item doesn't work, and instead prints an error in the console

### How to reproduce the bug

1. open crafting terminal
2. craft an item

it might be important that more of the same item are available in the network

### Expected behavior

the item gets crafted

### Additional details

_No response_

1.20

Forge

### Crash log

https://gist.github.com/Mari023/c9cc3fc339f96729de23b30ac085bb58

shartte:
Fixes #7676


GlennSchner:
Any idea when this request might be merged, or a temporary fix we can do in-game?

Mari023:
> Any idea when this request might be merged, or a temporary fix we can do in-game?

this is already merged, it only affected 17.6.1-beta, the fix was released in 17.6.2-beta.
all older versions aren't affected, and all newer versions have this bug fixed

PR #7739, opened 2024-03-10
Fix rare QNB crash when leaving a single-player world

shartte:
Got an NPE when leaving a single player world because it was trying to destroy this QNB cluster (that had no center).


src/main/java/appeng/me/cluster/implementations/QuantumCluster.java
```diff
@@ -61,7 +61,7 @@ public QuantumCluster(BlockPos min, BlockPos max) {
 
     @SubscribeEvent
     public void onUnload(final LevelEvent.Unload e) {
-        if (this.center.getLevel() == e.getLevel()) {
+        if (this.center == null && this.center.getLevel() == e.getLevel()) {
```
Mari023, line 64: shouldn't this be `!=` instead of `==`?
won't this guarantee a NPE now?

PR #7751, opened 2024-03-12
Pattern providers now only push ingredients for crafting patterns to molecular assemblers, not other inventories

### LINKED ISSUE #7688: Make Crafting Pattern only works with Molecular Assembler
### Describe the feature

I think this is what we should have. Currently a Pattern Provider with Crafting Pattern can push items to any inventories.

Since we already have Processing Patterns, this *feature* feels redundant.
Although you can't encode a Crafting Pattern with no matching crafting recipe, things can still go wrong if you have both Crafting and Processing Patterns in one Provider.

### Additional details

_No response_

shartte:
Fixes #7688


Technici4n (approved):
Basically what I had in mind as well.

PR #7766, opened 2024-03-15
Move EncodedPatternItem to API

XFactHD:
This PR moves `EncodedPatternItem` to API to avoid large amounts of copy-paste code for custom patterns in addons.
I previously discussed with @Technici4n to move large parts of that class to an API interface such that pattern item tooltips are uniform and to make the pattern preview work for custom encoded patterns. After thinking about it a bit more, I opted to move the entire class instead as moving "only" the relevant parts to an interface would effectively leave `EncodedPatternItem` as an almost empty class. If you prefer the interface approach despite that (i.e. to avoid indirectly exposing `AEBaseItem` in the API), please let me know and I'll adjust it accordingly.

### DIFF
--- a/src/main/java/appeng/api/crafting/EncodedPatternItem.java
+++ b/src/main/java/appeng/api/crafting/EncodedPatternItem.java
@@ -16,7 +16,7 @@
  * along with Applied Energistics 2.  If not, see <http://www.gnu.org/licenses/lgpl>.
  */
 
-package appeng.crafting.pattern;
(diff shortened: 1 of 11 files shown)


XFactHD:
My custom item doesn't have to do anything special, it's currently just a kinda half-assed copy of `EncodedPatternItem`, so a simple factory pattern would work totally fine.

XFactHD:
Closing in favor of #7770

PR #7770, opened 2024-03-16
Rethink basic pattern items and expose a pattern item builder to Addons

shartte:


XFactHD (commented):
This solution is a lot saner than my bruteforce idea of just moving them item to API and it would be perfectly suitable for my needs. Is there any particular reason to use a separate interface for the pattern decoder instead of the existing `IPatternDetailsDecoder` which has to be registered to `PatternDetailsHelper` anyway?

The only additional thought I have (which my PR wouldn't have solved either) is the ability to control how the `InvalidPatternHelper` decodes the inputs and outputs from broken patterns. In my case, I only keep around the registry name of the primary input and the output. The secondary inputs are stored separately. The current implementation wouldn't be able to decode them at all and if I adapt my input and output to be a list tag, then it would still miss the secondary inputs. The additional methods needed for that could be added to `IPatternDetailsDecoder` if that class were re-used instead of a new interface.

XFactHD (approved):
Looks good to me, this will work very well for my needs :shipit:

PR #7776, opened 2024-03-19
Increase the generation rate of the crystal resonance generator to 20 AE/t

shartte:


Technici4n (approved):
Note that you'll have to send a config update PR for FC.

PR #7833, opened 2024-04-20
Pattern Access Terminal's search box will not auto focus

SpCoGov:
Why don’t all terminals except Pattern Access Terminal auto focus?

### DIFF
--- a/src/main/java/appeng/client/gui/me/patternaccess/PatternAccessTermScreen.java
+++ b/src/main/java/appeng/client/gui/me/patternaccess/PatternAccessTermScreen.java
@@ -171,9 +171,6 @@ public class PatternAccessTermScreen<C extends PatternAccessTermMenu> extends AE
 
         super.init();
 
-        // Auto focus search field
-        this.setInitialFocus(this.searchField);
-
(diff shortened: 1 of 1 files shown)


Mari023:
you have to enable `Auto-Focus on open` in the terminal settings for all other terminals

SpCoGov:
> you have to enable `Auto-Focus on open` in the terminal settings for all other terminals

I think we should standardize things a bit. Except for 'Pattern Access Terminal' being 'Auto-Focus,' none of the others are. So, I believe 'Pattern Access Terminal' shouldn't be 'Auto-Focus' either.

Mari023:
> I think we should standardize things a bit. Except for 'Pattern Access Terminal' being 'Auto-Focus,' none of the others are. 

the other terminals have auto focus, it just isn't enabled by default

SpCoGov:
> > I think we should standardize things a bit. Except for 'Pattern Access Terminal' being 'Auto-Focus,' none of the others are.
> 
> the other terminals have auto focus, it just isn't enabled by default

So I want to have PatternAccessTermianal disabled for Auto-Focus by default

Issue #7468, opened 2023-11-11
Incorrect crafting job progress bar (not consumed item)


RobbyLPHD:
### Describe the bug

As soon as there are reused items in an autocrafting recipe, the display goes above 100%. The remaining time display also disappears as soon as 100% is exceeded.
(Screenshot in Additional details.)

### How to reproduce the bug

Use an item in a crafting recipe that is not consumed. The display now goes over 100% for a larger recipe, as the recovered item probably counts towards the total number of crafted items. (Noticed when crafting Mystical agricultur essence with master infusion cristal.)

### Expected behavior

Normally 100% should not be exceeded.

### Additional details

![2023-11-11_23 51 46](https://github.com/AppliedEnergistics/Applied-Energistics-2/assets/112067765/c51c21a1-e6ce-4684-a1fb-1925b4851cad)

### Which minecraft version are you using?

1.20

### On which mod loaders does it happen?

Forge

### Crash log

No crash, just a display error

Mithi83:
This reminds me of this of this bug here: https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/7420

I think your observation that items that are used multiple times (like catalysts in chemistry) is the key point here. I would guess that during crafting the catalyst is counted each time a crafting step finished while the total number of expected crafting steps only counts it once. This would imply that the number of finished steps would be larger than the total number of expected steps at some point and the bar would exceed the maximum width.

There are two things that probably need be corrected here:
* Clamp the maximum width during rendering of the progress bar to 100%. This fixes the visual bug of rendering a too large progress bar. However the bar will reach 100% long before the crafting request is finished.
* Fix the discrepancy between expected step calculation and actual counting as described above. Then the problem of the previous change would be remedied.

For the second part I‘m not sure which of both behaviors is the correct one and which one needs adjustment.

Mithi83:
I have some slight problems to reproduce your problem, as I can't find any hints for compiling AE2 for forge.

The immediate problem with the bar which exceeds 100% probably comes from here (src\main\java\appeng\client\gui\widgets\CPUSelectionList.java):
```java
                // Draw a bar at the bottom of the button to indicate job progress
                var progress = (int) (cpu.progress() * (buttonBg.getSrcWidth() - 1) / Math.max(1, cpu.totalItems()));
```

[...]

PR #7839, opened 2024-04-25
Fix crafting job progress bar exceeding 100%

Mithi83:
This fixes a progress bar rendering more than the maximum width by clamping the percentage if the crafting job makes heavy use of non-consumed items.

Note that this is merely a cosmetic fix for an overly long progress bar. The bar will still reach 100% much faster than completing the job. This is a more fundamental issue that still needs to be fixed separately.

Partial fix for #7468.

---

Note that I could not in fact verify that it does no longer occur, because I have no setup with mods using non-consumed items. I still think that a cosmetic fix is better than nothing.


Mithi83:
Superseded by #8005. I decided to not include this visual fix in the new approach to allow future bugs to be visible.

PR #7840, opened 2024-04-27
fix mods.toml -> neoforge.mods.toml in build.gradle

Mari023:
this caused the variables not getting replaced, which among other issues causes ae2's version to be 0.0.0 ingame


Mari023:
fixed in debd8d1bb292bbcfb79ebeb166713b039d2e83f3

PR #7845, opened 2024-04-29
Fix german translation of 256k storage cell

Tischi86:
The current translation indicates that it is a 64k cell, which is incorrect.


62832:
Translations are done through [Crowdin](https://appliedenergistics2.crowdin.com/applied-energistics-2), not here.

PR #7847, opened 2024-04-30
Fix memory card tooltip not showing P2P frequency

Mithi83:
A missing translation text of gui.tooltips.ae2.P2PFrequency caused a messed up tooltip for a memory card with P2P information on it.

Memory-Card-Tooltip
Memory-Card-Tooltip-fixed

Some notes for the review: The actual text is a copy of waila.ae2.P2PFrequency. Other duplicates in the translation file make me think that the duplication is the way to go, rather than to change the code in MemoryCardItem.appendHoverText() to use that translatable string as that would be the alternative.

Sidenote: With the 1.20.6 changes the memory card again works correctly with P2P-Tunnels.


Mari023:
en_us.json is made using datagen, if there is something missing it means either the datagen code is wrong, or something is using the wrong string (or not using the constants at all).
also main is 1.20.6, so if this doesn't affect 1.20.6 you probably want to target a different branch

Mithi83:
I'm not familiar with the datagen way to checking for translatable strings. The code using that translatable string is in MemoryCardItem.java, function appendHoverText:

```java
        var p2pFreq = stack.get(AEComponents.EXPORTED_P2P_FREQUENCY);
        if (p2pFreq != null) {
            final String freqTooltip = ChatFormatting.BOLD + Platform.p2p().toHexString(p2pFreq);
            lines.add(Tooltips.of(Component.translatable("gui.tooltips.ae2.P2PFrequency", freqTooltip)));
        }
```

Mari023:
> I'm not familiar with the datagen way to checking for translatable strings.

the Strings are all explicitly registered to datagen (see https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/main/src/main/java/appeng/datagen/providers/localization/LocalizationProvider.java)

 the code should either re-use the existing constant, like this:
 ```diff
 var p2pFreq = stack.get(AEComponents.EXPORTED_P2P_FREQUENCY);
 if (p2pFreq != null) {
     final String freqTooltip = ChatFormatting.BOLD + Platform.p2p().toHexString(p2pFreq)
-   lines.add(Tooltips.of(Component.translatable("gui.tooltips.ae2.P2PFrequency", freqTooltip)));
+   lines.add(Tooltips.of(Component.translatable(InGameTooltip.P2PFrequency.getTranslationKey(), freqTooltip)));
}
 ```
 
 or manually add the key, like the others here: 
 https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/69c8d37832d4bcba7468ff73425f2bdc3b4fcba7/src/main/java/appeng/datagen/providers/localization/LocalizationProvider.java#L110-L151

Mithi83:
Thanks for your explanations! I've updated the branch according to your first suggestion.

PR #7871, opened 2024-05-17
incorrect translation was fixed

FriendlyCube:
incorrect translation was fixed


62832:

PR #7883, opened 2024-05-27
Search upgrade master

JoelMorrisey:
This was created based on #7311, instead of relying on the functionality of JEI, REI, etc. I modified the search functionality to search similar to RS, JEI, etc.

The reason behind this PR is the same reason as the previous PR (#7311):
>This can be useful because it is annoying when you try to search something with the modid and the name like it is possible in JEI. For example "@ae2 crystal". Without it, it tries to search a mod with the name "ae2 crystal".


src/main/java/appeng/client/gui/me/search/SearchPredicates.java
```diff
@@ -1,84 +1,51 @@
...
-
-    private static Predicate<GridInventoryEntry> createModNamePredicate(String searchText) {
-        var searchPattern = createPattern(searchText);
-        return entry -> {
-            var what = Objects.requireNonNull(entry.getWhat());
-            return searchPattern.matcher(Platform.getModName(what.getModId())).find();
-        };
-    }
+        if (orParts.length == 1) {
+            gridPredictions = getPredicates(searchString);
```
Mari023, line 17:
```suggestion
            gridPredictions = getPredicates(orParts[0]);
```

Otherwise you have an edge case with strings which end in `|` and contain no other pipe,
these search strings currently find no items. 

e.g. `redstone` finds redstone dust etc., but `redstone|` won't find anything

Mari023 (commented):
I played a bit with it, I like it.

src/main/java/appeng/client/gui/me/search/SearchPredicates.java
JoelMorrisey: Thanks, tested and pushed through a fix :)

JoelMorrisey:
> You did remove the tooltip cache in this PR, but did not replace it with any other caching. You are aware that we actually needed that to compensate for broken other mods, yes?

Sorry about that must have forgotten to add that logic back in when I was refactoring the code to use predicate objects.

Fixed it now.

Also realised I missed support for searchTooltips option, so added that in and switched it to false by default as for most search use cases I imagine the user will use `#` to search the tooltips specifically :)

PR #7989, opened 2024-06-30
Fix GUI for formation plane, import bus and export bus

### LINKED ISSUE #7963: Player inventory in Export Bus is Misaligned
image

Mithi83:
Fixes #7963.

All three use the storagebus.png but differ in the srcRect size. While storage_bus.json was updated during retexturing, the others were not leading to:
  - misaligned player inventory
  - missing a 2 pixel wide bottom border area


Mithi83:
Side note: I did a quick grep for other potential textures shared by multiple guis and this should be the only case with out of sync srcRect values.

Issue #7420, opened 2023-09-27
Progress bar exceeding the overlay when autocrafting anything


Thurion666:
### Describe the bug

When doing an autocraft, the progress bar mostly all the time overpasses the overlay, as shown in the image below.

![image](https://github.com/AppliedEnergistics/Applied-Energistics-2/assets/38357989/9e587ce0-fa3e-40bf-9ab2-a646b66068ee)

### How to reproduce the bug

Just autocraft anything in AE2

### Expected behavior

Progress bar not exceeding the left side.

### Additional details

It could be that i am using 1 pattern to 4 or more molecular assemblers, for maximum speed. But i am not sure, i didn't tested it.

### Which minecraft version are you using?

1.19

### On which mod loaders does it happen?

Forge

### Crash log

https://pastebin.com/HV4fpvM9

PR #8005, opened 2024-07-01
Fix crafting job progress bar exceeding 100%

Mithi83:
Fix for 1.21 for the bug reported independently in #7468 for 1.20 and #7420 for 1.19.

The bug stems from the discrepancy between the initial calculation that only accounts for recipe outputs and the actual execution of the job which adds recipe outputs as well as any remaining items in the crafting grid, called container items. When decrementing the counter for the remaining items, these container items must not be counted to ensure the counter only decrements down to 0 and not beyond.

Fixing this side of the problem, rather than the initial calculation has two reasons:
  - more accurate progress bar, because recipes with container items don't decrease the counter multiple times per recipe execution, which will become important for more complex crafting jobs with a mix of normal recipes and those with additional container items.
  - the actual number of container items is determined only at the execution stage and not during the initial calculation.

### DIFF
--- a/src/main/java/appeng/crafting/execution/CraftingCpuHelper.java
+++ b/src/main/java/appeng/crafting/execution/CraftingCpuHelper.java
@@ -102,7 +102,8 @@ public class CraftingCpuHelper {
             IPatternDetails details,
             ICraftingInventory sourceInv,
             Level level,
-            KeyCounter expectedOutputs) {
+            KeyCounter expectedOutputs,
+            KeyCounter expectedContainerItems) {
 
         // Extract inputs into the container.
         var inputs = details.getInputs();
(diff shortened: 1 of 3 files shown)


Mithi83:
Simple way to test this: Create a pattern for cable anchors and craft large quantities (hunderts are sufficient). The damaged knifes will trigger this bug.

The most critical part is in my opinion is CraftingCpuLogic.insert().

There is a separate bug that messes up crafting jobs that are not done when leaving and rejoining the world, which is why I couldn't properly test the NBT writing portion of my fix.

Technici4n:
Tomorrow :)

Technici4n:
I merged #8085 instead. Thanks for the analysis and for the fix idea!

Mithi83:
You‘re welcome. I was quite sure the analysis was correct but as always with unfamiliar code there might be a more elegant solution. Thanks for fixing the bug!

PR #8030, opened 2024-07-12
Fix EMI recipe widgets

Mithi83:
The new textures moved some coordinates around. Since the textures are reused by the EMI recipe widgets they need to be adapted accordingly.


Mithi83:
conderser old
conderser new

inscriber old
inscriber new

PR #8043, opened 2024-07-18
Fix resuming crafting recipe jobs after a world load

### LINKED ISSUE #8006: Crafting jobs don't resume after loading a saved world
### Describe the bug

A crafting job that is still running when leaving the world will not resume when loading that world again.

### How to reproduce the bug

- Create autocrafting setup
- Start a long running job (in my case 400 Iron Blocks)
- Click "Save and Quit to Title" while the job is still running
- Load that world back up
- Inspect the crafting status and notice no progress

### Expected behavior

Crafting resumes after loading the world.

### Additional details

_No response_

1.21

NeoForge

### Crash log

Selfcompiled main branch

Mithi83:
The ItemStack type used in EncodedCraftingPattern provides no hashCode() function and is thus not usable in hashCode() functions of objects containing it.

This leads to an effect where a running job using an EncodedCraftingPattern does not resume after loading the world. Providing an explicit hashCode() function for EncodedCraftingPattern solves this problem.

Fix #8006.


62832:
Considering this is a `record`, isn't this already supposed to have a built-in `hashCode` implementation?

PR #8080, opened 2024-07-31
Fix annihilation planes working when the network is offline

Technici4n:
We could do the reset anyway even the plane is not active, but it doesn't seem useful.


Technici4n:
Yes. We have a call to `refresh` in `onMainNodeStateChanged` which takes care of that.

PR #8112, opened 2024-08-05
Add craft all functionality

PiTheGuy:
Alt-clicking the recipe output slot in the crafting terminal now crafts the item until you run out of inventory space or materials, as opposed to shift clicking which only crafts up to 1 stack of the output item.


Technici4n:
I wonder if space+click would make more sense.

PR #8139, opened 2024-08-12
Optimized UI Drawing

shartte:


src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -777,21 +757,108 @@ protected boolean isHovering(Slot slot, double x, double y) {
...
 
-        var background = style.getBackground();
-        if (background != null) {
-            background.dest(offsetX, offsetY).blit(guiGraphics);
+        widgets.addBackgroundPanels(panels, getBounds(true));
+
+        panels.render(bgLayer, 0, 0xffffffff);
+
+        // Allow all composite widgets to participate in the sprite layer
+        widgets.drawBackgroundSpriteLayer(bgLayer, getBounds(true), new Point(mouseX - leftPos, mouseY - topPos));
```
sh1ftchg, line 781: There's only one mouse and creation of new unmanaged objects like a new `Point` is not really necessary is it?

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -777,21 +757,108 @@ protected boolean isHovering(Slot slot, double x, double y) {
...
                     0, 0,
                     generatedBackground.getWidth(),
                     generatedBackground.getHeight());
-            blitter.blit(guiGraphics, offsetX, offsetY);
         }
 
-        var background = style.getBackground();
-        if (background != null) {
-            background.dest(offsetX, offsetY).blit(guiGraphics);
+        widgets.addBackgroundPanels(panels, getBounds(true));
```
sh1ftchg, line 776: Store the result of `getBounds` - excessive cyclic calls are also ill advised as this is used not even 2 lines later.

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -777,21 +757,108 @@ protected boolean isHovering(Slot slot, double x, double y) {
...
+
+        panels.render(bgLayer, 0, 0xffffffff);
+
+        // Allow all composite widgets to participate in the sprite layer
+        widgets.drawBackgroundSpriteLayer(bgLayer, getBounds(true), new Point(mouseX - leftPos, mouseY - topPos));
+
+        // Group all slots by semantic. We make the assumption here
+        // that visually the slots belonging to the same semantic are contiguous.
+        for (var semantic : menu.getSlotBySemantic().keySet()) {
```
sh1ftchg, line 785: Store the `menu.getSlotBySemantic()` as a local variable (`lv`) as per the aforementioned advice. it's used a lot and recursive calls are still ill advised.

src/main/java/appeng/client/gui/AEBaseScreen.java
```diff
@@ -777,21 +757,108 @@ protected boolean isHovering(Slot slot, double x, double y) {
...
+                        backgroundHeight,
+                        color);
+
+                minX = Math.min(minX, backgroundX);
+                minY = Math.min(minY, backgroundY);
+                maxX = Math.max(maxX, backgroundX + backgroundWidth);
+                maxY = Math.max(maxY, backgroundY + backgroundHeight);
+            }
+
+            for (var slot : slots) {
```
sh1ftchg, line 826: Is two iterations necessary? Couldn't this pass be done in the previous loop on line 795 and again at line 824? The control flow graph only shows those 2 branch points? Also it's generally bad practice to separate alterations and work as it loses context of where/when it was modified. 

Plus you do the exact same `instanceof` and `isActive` logic in the aforementioned loop already... Why make it do double work - especially considering that aside from object creation (`new`) `instanceof` is the most costly java dynamic.

sh1ftchg (commented):
Nothing bad just some minor comments about performance pet peeves. Great work!

PR #8158, opened 2024-08-23
Improve controller textures

Mithi83:
As discussed in Discord, here are some ideas on how to improve the controller textures to avoid color mismatches in certain places. The basic idea is to expand on the (currently broken) feature for the 'inside' variant of controllers to have a and b variants of the blocks with different textures (colors swapped). It has been separated into 3 commits, each dealing with one variant of controller blocks (columns, blocks (i.e. corners) and inside).

I'm unfamiliar with what these changes will cause in terms of trouble for upgrading/downgrading existing worlds since the block states change. If columns and corners both will be merged some thought should be given to the question whether there should be a refactoring (moving the a or b information into a separate block state, as all CONTROLLER_TYPE will have a and b subtypes).

Some example pictures:
controller-texture-online
controller-texture-complex
controller-texture-offline
controller-texture-conflict
controller-texture-inside

Note that this approach cannot solve certain color mismatches, in particular corner edges and inner edges:
controller-texture-unsolvable

I'm posting this as a draft pull request as there should probably be some discussion.


heinrich26:
Just looked over your proposal briefly and i'm wondering, why you're not just implementing CTM or adding it as a dependency, thus eliminating the blockstates. 
You might even just add the CTM texture, keep the old behavior and users with CTM will enjoy your work (that's at least bundled with Chisel and OptiFine (there are more).

Mari023:
I wonder, all those remaining problems are on edges, are all edges problematic?
If so, couldn't you just mirror the texture on the y axis compared to those on the x axis?
that should fix all the sides.

Tho you'd still have the top and bottom broken with either the x or the y axis...
Tho I don't think a 3-way edge is fixable in general with the controller pattern...

Mithi83:
I'm pretty sure it is mathematically impossible to make it work in all cases. Basically you have two different variants of the texture, A and B. To make it look good you must ensure to always alternate between A and B, i.e. a controller side with texture A must be next to a controller side with texture B. Taking a look at a single controller block this is already impossible to achieve. Choose a face and put texture A on it. Then take an adjacent side, which according to the rules would have to be texture B. Now choose one of the two remaining sides that are adjacent to both previously chosen sides. It is impossible to satify the rule again, because choosing either A or B would be not allowed since there is already an adjacent side with that texture. You could argue that fixing a certain direction would be more beneficial rather than fixing the other direction due to that fact that most players look upon their controllers from a certain position (e.g. more people have the controller on the same y-level as the player, rather than having the controller under their feet for example), but I'm not sure that this would not make those edge cases look even worse.

That being said: My approach is even more simplistic. I do not change individual sides, I change the entire block to have texture A or B on all sides. With that I'd expect any edge within the same block being problematic.

Therefore this is just an improvement and not a complete fix, which I suspect to be impossible mathematically given the properties of the texture itself.

Mari023:
I am roughly aware of your how your fix works, this was just an additional suggestion to fix some more cases. 
That is what I meant with the 3 way edges being impossible, you'd need A B and C faces for that piece to work correctly mathematically.

I mostly meant that as an additional improvement to fix one of those two cases you circled.

And thinking about that again, it aren't just the corner pieces, one of the x/y/z pairs will always be broken.

PR #8045, opened 2024-07-18
Replace raw long counting with a Long2LongMap accumulator

cplir-c:
Attempts to fix the [fluid using 1 Bucket = 81k items issue](https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/7321).

I chose to use a Long2LongMap accumulator so that different AEKeyTypes with the same denominator could be summed in the same map slot, and avoid the internal-rounding issues that would be had if I divided all the values by their Unit denominators.

Issue #7321, opened 2023-08-07
Progress item number incorrect


Simba98:
### Describe the bug

When I use the wireless terminal to view cpu progress, the number shows even larger than the bytes of the cpu.

After looking the source code, this is due to one bucket = 81k items.

But the setting makes the progress meaningless, because the fluid will dominant the progress.

For some case, it use 1 bucket = 81k item.

![image](https://github.com/AppliedEnergistics/Applied-Energistics-2/assets/7941967/6cbc56ab-8872-4ab9-8c12-903eb88ac21e)
![image](https://github.com/AppliedEnergistics/Applied-Energistics-2/assets/7941967/34b4d809-f666-479c-9364-ca3afb3f3fb0)

### How to reproduce the bug

Make fluid as output

### Expected behavior

Maybe use 1bucket = 1item to balance.

### Which minecraft version are you using?

1.19

### On which mod loaders does it happen?

Fabric

### Crash log

NA

Simba98:
I guess this is due to https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/172a36daec8462bec4711e2660ac47c7d9021867/src/main/java/appeng/crafting/execution/ExecutingCraftingJob.java#L81

PR #8170, opened 2024-08-27
Implement ElapsedTimeTracker to support AEKeyType AmountPerUnit scaling

Simba98:
Implement #7321, and replace #8045.

This is by adding internal scaling to keep the tracker as an integer.

Possible issue: broken changes due to NBT data in the disk may be invalid.


Simba98:
Updated to newest commit

62832:
Please run the `spotlessApply` Gradle task for proper code formatting and for the automated build to pass.

Simba98:
> Please run the `spotlessApply` Gradle task for proper code formatting and for the automated build to pass.

Hi 62832, thanks for reminding me. I have applied the spotless format check and passed the checks.

src/main/java/appeng/crafting/execution/ElapsedTimeTracker.java
```diff
@@ -77,10 +81,10 @@ public long getElapsedTime() {
     }
 
     public long getRemainingItemCount() {
-        return this.remainingItemCount;
+        return Math.round(((float) this.remainingItemCount) / AEKEY_SCALE_FACTOR);
     }
 
     public long getStartItemCount() {
-        return this.startItemCount;
+        return Math.round(((float) this.startItemCount) / AEKEY_SCALE_FACTOR);
```
cplir-c, line 88: Why not just use integer division here?

Simba98: > Why not just use integer division here?

Integer Division equals Math.floor, making the 99/10 = 9 instead of 10 in this case.
I want to round to make it more precise in displaying, for showing progress, the trade-off is performance.

Simba98: > If you want to simulate math.round() with integer division you can add half of the AE_SCALE_FACTOR before dividing, right?

You are right. I have updated.

PR #8180, opened 2024-09-04
Fix wrong budding block tag

Mari023:
Found during Soaryns stream, where Direwolfs new budding blocks didn't work with the the growth accelerator

removed some outdated javadoc, and some tags / tag content already provided by neoforge
also uses the code reference for tags where it makes sense
(the population of AETags.METAL_INGOTS still uses resource locations even where neoforge provides convention tags, but I've kept those for readability)


Mari023:
superseded by #8224

PR #8184, opened 2024-09-06
Correction of RU translation

MRX8024:
At the moment, in Russian translation, all “dense” cables are called “not transparent”, and some are even called the same as ordinary ones.
I have corrected this situation.


Mari023:

MRX8024:
> Translations are done via [Crowdin](https://appliedenergistics2.crowdin.com/applied-energistics-2), not here.

Thanks for the tip! But on crowdin there are only 5 pages of Russian translation out of 21 English translations, and I can’t find where I can choose the IDs of items for further translation, could you help me?

PR #8215, opened 2024-10-17
Add all viewable items in MEStorageMenus to Emi inventory

### LINKED ISSUE #8214: EMI BOM (synthetic favourites) support
### Describe the feature

This mod add this functionality, but it's only for fabric.
https://www.curseforge.com/minecraft/mc-mods/ae2-emi-crafting-integration/files/5160115

https://github.com/emilyploszaj/emi/issues/696#issuecomment-2347974496
> EMI supports this, AE2 does not

The recipe tree and the generated bill of materials in EMI are some of the best QoL features in expert modpacks like GregTech. Having it before accessing an ME system is a treat, but losing it once you get ME access is quite sad.

### Additional details

_No response_

ramidzkh:
Also fixes a bug where items in the crafting terminal grid would disappear from the tree

Closes #8214

### DIFF
--- a/src/main/java/appeng/integration/modules/emi/AbstractRecipeHandler.java
+++ b/src/main/java/appeng/integration/modules/emi/AbstractRecipeHandler.java
@@ -51,6 +51,7 @@ abstract class AbstractRecipeHandler<T extends AEBaseMenu> implements StandardRe
         var slots = new ArrayList<Slot>();
         slots.addAll(menu.getSlots(SlotSemantics.PLAYER_INVENTORY));
         slots.addAll(menu.getSlots(SlotSemantics.PLAYER_HOTBAR));
+        slots.addAll(menu.getSlots(SlotSemantics.CRAFTING_GRID));
         return slots;
     }
 
--- a/src/main/java/appeng/integration/modules/emi/EmiUseCraftingRecipeHandler.java
+++ b/src/main/java/appeng/integration/modules/emi/EmiUseCraftingRecipeHandler.java
@@ -1,11 +1,14 @@
 package appeng.integration.modules.emi;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
 import net.minecraft.client.gui.screens.inventory.AbstractContainerScreen;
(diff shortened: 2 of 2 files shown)


theboo:
So i made a build against this PR and threw it on my family e10 server, and it seems to work fairly well as far as showing what items you have vs synthetic favorites. Does not take much time at all for emi to see what is available for the synthetic faves. My one issue is the "craft on/all to inventory/cursor" keybinds that are part of this do not seem to work while in the crafting terminal. This feature is really what makes the BOM/Synthetic system so powerful. Two gifs below show what happens when i craft using synthetic favorites in a vanilla crafting table vs what happens in ae2 using the same hotkey.
Vanilla table:
https://gyazo.com/8f22133858af7aa8434f02db5ccb168a
Crafting Terminal(notice how quickly EMI is able to recognize what is available in the ME system, but i cannot craft to either inventory with my hotkeys like previous gif):
https://gyazo.com/7fe814aa07e2079e6ed50e9ab2431b3c

theboo:
After using this PR on my family server for the whole weekend, i can say i didnt have any sort of noticeable performance issues(ofc this should be tested by others, and im sure it would be before it makes it to a release lol. I did not do any objective testing through spark profile or otherwise)... 

Missing the hot key support ended up not being as big of a deal as i thought(PR is still AMAZING without it), was only mostly annoying if crafting something just under a stack and you have the resources to craft more(emi hotkeys will craft the exact amount you need when using "craft all") so you end up either shift clicking to craft a stack which over-crafts what you need, or you just play cookie clicker for a bit to get the amount you need. Other than that, i think this PR is an amazing addition, could be something that is off by default with a toggle in the terminal settings if performance ends up being an issue.

Sorry if this input is not wanted here, will be my last comment on it unless asked for more input haha. Thanks for your time

TheEaglez:
Hi, I have a maybe a bit out there question. Is it possible to also add this to the 1.20.1 version? I was searching for this functionality specifically and was pretty happy to see that you already fixed it. But then couldn't find it anywhere in my config only to come to the conclusion it was only added to the 1.21.1 version of the mod. 

I totally get it if its not possible either time wise or limitations. Either way Thanks for the great mod :)

PR #8241, opened 2024-11-13
Update de_de.json

Hfllfijx:
Fixed small error


Hfllfijx:
im so dumb

Ommina:
Still, isn't the spirit of the change correct?

  "block.ae2.256k_crafting_storage": "16k-Fertigungsspeicher",
  "block.ae2.16k_crafting_storage": "16k-Fertigungsspeicher",

I don't know German in any capacity, but shouldn't the value for "256k_crafting_storage" be "256k-,,," instead of duplicating the 16K value?

Mari023:

for some reasons I couldn't find the translation in the GUI, I could however download the translations as a csv, edit it and uploaded it.
It should be fixed the next time changes from crowding get merged.

PR #8246, opened 2024-11-17
Remove most instances of @OnlyIn

wolfieboy09:
That annotation should NOT be used by modders at all.


wolfieboy09:
okay forget it

Issue #8274, opened 2024-12-11
Consistent network rebooting using a controller and infinite channels


bk5115545:
### Describe the bug

I'm running with `appliedenergistics2-forge-15.2.13` using the `"channels": "infinite",` config option. **This issue does not happen when only using an energy acceptor instead of a controller.**

When there is a controller on my network, the network reboots every time a cable is added or removed. When there is not a controller on the network, there is no reboot.

I want to use a controller because it decreases the passive AE drain.

This screenshot is a small network that exhibits this issue. My real network is much more complex and part of a modpack but I reproduced this issue running only AE2. Additionally, my network does not reboot when I make changes without a controller so I think I've isolated the issue to be related to the controller.

This network reboot issue is causing issues for my friends and I because it resets UIs and stops general auto-crafting until the network reboots. We can mostly work around it by setting `pathfindingStepsPerTick` to `128` however that causes some TPS lag in our world.

![image](https://github.com/user-attachments/assets/cd899c55-595e-46f4-b89e-325875de353c)

### How to reproduce the bug

Create a network using infinite channels and add a controller. Add a cable. The network will reboot.

This issue is more obvious with larger networks. You can set `pathfindingStepsPerTick` to `1` to simulate a 4x larger network than you build.

### Expected behavior

The network should not reboot as it doesn't need to without the controller.

### Additional details

Optionally, a config option to disable the increased AE drain on controller-less networks would also allow us to work around this issue.

### Which minecraft version are you using?

1.20.1

### On which mod loaders does it happen?

Forge

### Crash log

Does not crash

jpeaglesandkatz:
Yeah I noticed that too.. That is just silly and very unfortunate

bk5115545:
The network doesn't reboot when there is no controller while using infinite
channels; at least if it does, there is no distribution to terminals or
interfaces

bk5115545:
This code is pretty good...

I believe I've isolated the issue to an assumption in https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/main/src/main/java/appeng/me/service/PathingService.java#L117-L129 that assumes a controller means the network needs to be rebooted even when using infinite channels. A simple fix may be to add `&& this.channelMode != ChannelMode.INFINITE`.

[...]

PR #8279, opened 2024-12-13
(#8274) Optimize PathingService and GridConnection repath conditions

bk5115545:
There are several ways that a network can enter the "rebooting" state however only a few actually have consequence to the overall network state.

This change removes the need for the network to reboot during "simple" Grid merges while still enforcing rebooting for "complex" merges. A simple merge is one that has no bearing on the network pathing; imagine a long cable with no channel-consuming devices attached to it and then you add another cable. A complex network merge is everything else (combining grids, adding channel-consumers, removing channel-consumers, deleting grid connections).

Of note: half-items like P2P, Quartz Fiber, and all terminals (though not other cable types) always force a complex merge for some reason but seeing as the network needs to reboot to calculate their energy/channel consumption; this is preferred.

I performed testing using multiple networks both with and without controllers. using 1x and "infinite" channel modes.
I confirmed that power usage matched before and after this change by setting up identical networks, using the network tool to check both idle and channel consumption, and then adding more cables (old version triggers a reboot while this change does not); and then verified that both consumptions were identical.

I noticed some oddities around not needing to update pathing when placing AE powercells but I'm not sure these can pass channels and they don't consume one so I guess this is expected. EnergyService seemed to handle everything correctly.
`runGametest` also passes.

### DIFF
--- a/src/main/java/appeng/api/networking/IGrid.java
+++ b/src/main/java/appeng/api/networking/IGrid.java
@@ -106,6 +106,11 @@ public interface IGrid {
     Iterable<IGridNode> getNodes();
 
     /**
+     * @return true if the grid contains the provided node
+     */
+    boolean containsNode(IGridNode node);
+
+    /**
      * @return true if the last node has been removed from the grid.
      */
     boolean isEmpty();
--- a/src/main/java/appeng/api/networking/pathing/IPathingService.java
+++ b/src/main/java/appeng/api/networking/pathing/IPathingService.java
@@ -23,6 +23,7 @@
 
 package appeng.api.networking.pathing;
 
+import appeng.api.networking.IGridNode;
 import appeng.api.networking.IGridService;
 
 /**
@@ -63,4 +64,9 @@ public interface IPathingService extends IGridService {
      * @return The total number of channels currently used by this network.
      */
     int getUsedChannels();
+
+    /**
+     * Whether this PathingService is already aware of the provided node
+     */
+    boolean contains(IGridNode node);
 }
--- a/src/main/java/appeng/me/Grid.java
+++ b/src/main/java/appeng/me/Grid.java
@@ -202,6 +202,11 @@ public class Grid implements IGrid {
     }
 
     @Override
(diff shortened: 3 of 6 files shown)


Technici4n:
Placing a controller or a channel-consuming device is a relatively "simple" update but it requires a repath, yet it seems that with this PR no repath would be triggered?

Technici4n:
Ah it seems that it handles some of these cases with some extra logic in `addNode`. I would not be too confident that this works in all cases though.

bk5115545:
I wasn't sure of the correct way to find all those devices so I tried to
use GridFlags.

I'm also partial to allowing PathingService to decide when it needs to
repath rather than several other components unconditionally forcing one
hence delegating that logic into addNode but I can move it if desired

bk5115545:
Luckily, adding any device that connects to 2 different GridConnections is
handled as a complex merge so creating shorter routes still forces a repath

Technici4n:
I started looking into a cleaner fix: https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/8285. The idea is simply to optimize the algorithm so that we can perform way more steps per tick. Potentially all steps immediately every time.

bk5115545:
#8285 looks really promising so far. I'm also experimenting with creating an overlay, sparse directed graph of nodes that care about channels and traversing that instead of the whole grid - this approach greatly optimizes the prior BFS implementation to instead use dijkstra which is the fastest single-source pathing algorithm.

For now, I'm using https://github.com/jgrapht/jgrapht to avoid writing my own pathfinding and traversal algorithms.

I'm targeting a reboot-less network for most operations (maybe still have reboots for grid merges but not looking required at this point). Hopefully it shows up in a PR later but I'm having issues with something I've nicknamed "reboot stability" wherein a network may no longer be able to provide channels to all devices if it is rebooted.

If I can work this out, it will probably including removing the `repath()` method in `IPathingService` and so may require a MV to avoid breaking consumers.

bk5115545:
It does yes. Thanks for the fast improvements

PR #8285, opened 2024-12-16
Optimize channel computation

Technici4n:
- Optimize the channel computation to make it go from O(N^2) to O(N).
- Cleanup related code a little bit. Most importantly, choose the same meaning for `usedChannels` vs `lastUsedChannels` for both `GridConnection` and `GridNode`.
- Remove booting delay. Note that the API still supports it, and that we might want to add back some sort of delay in the future if necessary.
- Remove `pathfindingStepsPerTick` config option.


Technici4n:
> I think you mean it goes from O(N^2) to O(N) I hoper

Yes :smile:

Technici4n:
The new algorithm is linear time because there are at most  2 jumps. (In theory. I have to double-check that this is true in practice.)

As a side note, BFS and DFS are both linear time traversal algorithms. And BFS can find shortest paths in linear time, the catch is that it only works with unweighted edges.

bk5115545:
Is this something that can be backported to 1.20 (I don't know your guidelines on versioning)?

I'd really like to get it into Monifactory and I can do the backport work/make a PR if you're alright with that

bk5115545:
You're amazing. Thank you!

PR #8291, opened 2024-12-18
[Feature] Allows players to reverse the function of '#' and '$' in search

LittleCircleOO:
Upon investigation, there is a difference in how # and $ are handled by different mods:

| Prefix | #       | $       |
| ------ | ------- | ------- |
| JEI    | tag     | tooltip |
| REI    | tooltip | tag     |
| EMI    | tag     | tooltip |
| RS     | tooltip | tag     |

So a client-side setting is provided to allow players to reverse their functionality, which also automatically assigns defaults based on the item lists mod on first load.


Mari023:
Don't some recipe viewers also have the ability to use any arbitrary symbol for those functions? (I think I have seen this in either REI or EMI, possibly both) It might make sense to have an option to get those symbols from the recipe viewer

LittleCircleOO:
> Don't some recipe viewers also have the ability to use any arbitrary symbol for those functions? (I think I have seen this in either REI or EMI, possibly both) It might make sense to have an option to get those symbols from the recipe viewer

Unfortunately, neither of these recipe viewer mods has been given a configuration option, and currently they only have a ‘search with or without prefix’ option, and custom prefixes are not yet supported.
Related issues:
https://github.com/emilyploszaj/emi/issues/767
https://github.com/shedaniel/RoughlyEnoughItems/issues/1652

This pull request will be marked as a draft for now and will be re-enabled when the recipe viewer supports this feature.

Mari023:
> Another problem this pull request contains is that it only considers both # and $. If the recipe viewer supports custom prefixes in the future, is it necessary to support each prefix specifying symbols in advance?

no, that can be done once it becomes relevant (i.e. once they get the feature, or when it is known they get the feature relatively soon (ie. if there is an open PR for that feature))

Technici4n:
Let's just use the intuitive characters: https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/8433.

PR #8304, opened 2024-12-25
A change for zh_cn.json

MJQFHCMZ:
I don't think there's much to say about it. I can try to translate it into Italian, but someone always translates it phonetically, so I'm going to change the translated text.


MJQFHCMZ:
Okay.

PR #8324, opened 2025-01-12
Allow filtering config stack slots with containers holding 0 of a stack

62832:
This PR allows config slots of `CONFIG_STACKS` mode to be filtered with a minimum of 1 of a given stack retrieved from a container, even if the container may have a zero amount of that stack.

The rationale for this comes from https://github.com/62832/ArsEnergistique/issues/17 and the [resulting Mixin](https://github.com/62832/ArsEnergistique/blob/master/src/main/java/gripe/_90/arseng/mixin/ConfigInventoryMixin.java) used to allow players to filter ME Interface slots to Source, even if the Source Jar being used to filter it is empty. As the Ars Énergistique add-on is designed around a key type for which only one possible resource exists (Source), it is wholly unambigious that any Source Jar may contain or make use of Source regardless of whether it currently doesn't contain any of it.

The same would apply, should it ever be updated to 1.21, to the Applied Botanics add-on and empty Mana Tablets as a result of this being included upstream, and has already applied as such, though only also in the presence of Ars Énergistique.

### DIFF
--- a/src/main/java/appeng/util/ConfigInventory.java
+++ b/src/main/java/appeng/util/ConfigInventory.java
@@ -210,7 +210,7 @@ public class ConfigInventory extends GenericStackInv {
                 // in stack mode, amounts of 0 or less will:
                 // - filter the slot to a minimum of 1 of that stack, if empty
                 // - clear the slot, if not empty
-                stack = getStack(slot) == null ? new GenericStack(stack.what(), 1) : null;
+                stack = getStack(slot) == null && mode == Mode.CONFIG_STACKS ? new GenericStack(stack.what(), 1) : null;
             }
         }
         super.setStack(slot, stack);
--- a/src/main/java/appeng/util/ConfigInventory.java
+++ b/src/main/java/appeng/util/ConfigInventory.java
@@ -207,8 +207,10 @@ public class ConfigInventory extends GenericStackInv {
                 // force amount to 0 in types-only mode
                 stack = new GenericStack(stack.what(), 0);
             } else if (!typesOnly && stack.amount() <= 0) {
-                // in stack mode, amounts of 0 or less clear the slot
-                stack = null;
+                // in stack mode, amounts of 0 or less will:
+                // - filter the slot to a minimum of 1 of that stack, if empty
+                // - clear the slot, if not empty
+                stack = getStack(slot) == null ? new GenericStack(stack.what(), 1) : null;
             }
         }
         super.setStack(slot, stack);
--- a/src/main/java/appeng/util/ConfigInventory.java
+++ b/src/main/java/appeng/util/ConfigInventory.java
@@ -207,10 +207,15 @@ public class ConfigInventory extends GenericStackInv {
                 // force amount to 0 in types-only mode
                 stack = new GenericStack(stack.what(), 0);
             } else if (!typesOnly && stack.amount() <= 0) {
-                // in stack mode, amounts of 0 or less will:
-                // - filter the slot to a minimum of 1 of that stack, if empty
-                // - clear the slot, if not empty
-                stack = getStack(slot) == null && mode == Mode.CONFIG_STACKS ? new GenericStack(stack.what(), 1) : null;
+                if (mode == Mode.CONFIG_STACKS && getStack(slot) == null) {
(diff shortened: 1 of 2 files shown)


src/main/java/appeng/util/ConfigInventory.java
```diff
@@ -207,8 +207,10 @@ public void setStack(int slot, @Nullable GenericStack stack) {
                 // force amount to 0 in types-only mode
                 stack = new GenericStack(stack.what(), 0);
             } else if (!typesOnly && stack.amount() <= 0) {
-                // in stack mode, amounts of 0 or less clear the slot
-                stack = null;
+                // in stack mode, amounts of 0 or less will:
+                // - filter the slot to a minimum of 1 of that stack, if empty
+                // - clear the slot, if not empty
+                stack = getStack(slot) == null ? new GenericStack(stack.what(), 1) : null;
```
Technici4n, line 213: This looks problematic for config inventories in `STORAGE` mode (currently only the bottom slots of an interface) where it should materialize one source. Can you 1) check that it is indeed problematic for those, and 2) make this only apply for the `CONFIG_STACKS` mode?

62832: Point 2 can be done, point 1 seems to be redundant as empty container items like these are already filtered out by the `AEBaseMenu::handleEmptyHeldItem` method and as a result this current method is never actually called for the contained zero-amount stack.

src/test/java/appeng/util/ConfigInventoryTest.java
```diff
@@ -77,15 +77,15 @@ class StacksMode {
         ConfigInventory inv = ConfigInventory.configStacks(1).supportedType(AEKeyType.items()).build();
 
         @Test
-        void stacksWithAmountZeroAreDropped() {
+        void stacksWithAmountZeroFilterToOne() {
             inv.setStack(0, ZERO_STICK);
-            assertNull(inv.getStack(0));
+            assertEquals(ONE_STICK, inv.getStack(0));
```
Technici4n, line 82: The tests should also verify that calling `setStack` a second time does indeed remove the stack.

src/main/java/appeng/util/ConfigInventory.java
```diff
@@ -207,8 +207,10 @@ public void setStack(int slot, @Nullable GenericStack stack) {
                 // force amount to 0 in types-only mode
                 stack = new GenericStack(stack.what(), 0);
             } else if (!typesOnly && stack.amount() <= 0) {
-                // in stack mode, amounts of 0 or less clear the slot
-                stack = null;
+                // in stack mode, amounts of 0 or less will:
+                // - filter the slot to a minimum of 1 of that stack, if empty
+                // - clear the slot, if not empty
+                stack = getStack(slot) == null && mode == Mode.CONFIG_STACKS ? new GenericStack(stack.what(), 1) : null;
```
Technici4n, line 213: A bit cleaner I think, and with a comment giving an example:
```suggestion
                if (mode == Mode.CONFIG_STACKS && getStack(slot) == null) {
                    // filter the slot to a minimum of 1 of that stack, if currently empty
                    // This allows setting a filter using empty containers that are limited to one resource, e.g. empty mana tablet.
                    stack = new GenericStack(stack.what(), 1);
                } else {
                    // in stack mode, amounts of 0 or less clear the slot
                    stack = null;
                }
```

Technici4n (changes requested):

src/test/java/appeng/util/ConfigInventoryTest.java
```diff
@@ -77,15 +77,18 @@ class StacksMode {
...
             inv.setStack(0, ZERO_STICK);
             assertNull(inv.getStack(0));
         }
 
         @Test
-        void stacksWithNegativeAmountsAreDropped() {
+        void stacksWithNegativeAmountsFilterToOne() {
             inv.setStack(0, new GenericStack(STICK_KEY, -1000));
-            assertNull(inv.getStack(0));
+            assertEquals(ONE_STICK, inv.getStack(0));
```
Technici4n, line 91: You could have amended this test too, but if you're lazy I will merge the PR without.

Technici4n (approved):

PR #8325, opened 2025-01-13
Backport #8322 to Forge 1.20.1

StarskyXIII:
As I'm merely replicating the implementation by embeddedt, I hope there won't be any problems.


62832:
I was unaware that this was actually a feature in Forge as well prior to NeoForge, but the more you know.
https://docs.minecraftforge.net/en/latest/rendering/modelextensions/facedata

PR #8335, opened 2025-01-18
Do not allow items to be inserted into the Matter Condenser when the output slot is full

### LINKED ISSUE #8311: Matter Condenser deletes items above storage cell limit if output is full
### Describe the bug

When a matter condenser has a 1k Storage Component in it and is set to produce matter balls (probably also works with singularities, haven't tested this), and 25600 Items (the amount needed for 100 of matter balls) are inserted without the output being cleared, the Matter Condenser will produce 64 matter balls and continue to store energy. When the energy reaches the maximum of the Storage Component, all further items will still be accepted, but will just be destroyed.

### How to reproduce the bug

- Put a 1k Storage Component into a Matter Condenser
- Set the Matter Condenser to produce Matter Balls
- Pump 25600 Items into it without taking anything from the output
- Take the entire output out of the machine
- The Output will be 96 Matter Balls (1.5 Stacks)

### Expected behavior

The Matter Condenser should not accept any more items than the Storage Component can hold, or this behavior should be documented in the Guide. The Guide currently states "Matter balls take 256 items or buckets each, so a 1k ME Storage Component (which provides 8192 bits of capacity) is more than enough."

### Additional details

I had this problem in the Modpack All the Mods 10

1.21

NeoForge

### Crash log

https://gist.github.com/adaexec/ee3c819a7b70828731cbafd29c28d2e0

62832:
Closes #8311.

### DIFF
--- a/src/main/java/appeng/blockentity/misc/CondenserBlockEntity.java
+++ b/src/main/java/appeng/blockentity/misc/CondenserBlockEntity.java
@@ -110,28 +110,25 @@ public class CondenserBlockEntity extends AEBaseInvBlockEntity implements IConfi
 
     private void fillOutput() {
         var requiredPower = this.getRequiredPower();
-        var output = this.getOutput();
-        while (requiredPower <= this.getStoredPower() && !output.isEmpty() && requiredPower > 0) {
-            if (this.canAddOutput(output)) {
+        while (requiredPower <= this.getStoredPower() && !getOutput().isEmpty() && requiredPower > 0) {
+            if (this.canAddOutput()) {
                 this.setStoredPower(this.getStoredPower() - requiredPower);
-                this.addOutput(output);
+                this.addOutput();
             } else {
                 break;
             }
         }
     }
 
-    private boolean canAddOutput(ItemStack output) {
-        return this.outputSlot.insertItem(0, output, true).isEmpty();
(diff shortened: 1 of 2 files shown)


Mithi83:
I was tempted to tackle this bug too but ultimately didn't because this changes the semantics of the matter condenser a bit. The old implementation was something along the line of it being a fancy trash can which _could potentially_ give you something in return if you can manage to pull it out fast enough. The emphasis was on _trash can_ with a small perk, i.e. its main feature was to void items as fast as they could be put inside of it.

Your PR changes that to something like this: The matter condenser is primarily used for generating a resource (matter balls or singularities) from trash. This can now stall if extraction is too slow to prevent the build up and this could propagate down the entire supply line and thus lead so some unintended side effects.

I've not taken a closer look at the code yet, but have you considered the cases that the matter condenser is set to trash can mode and may or may not have any storage? Will that stall eventually as well?

62832:
> I've not taken a closer look at the code yet, but have you considered the cases that the matter condenser is set to trash can mode and may or may not have any storage? Will that stall eventually as well?

If the Matter Condenser is set purely to its trashing mode then there's never an output to consider; `getOutput()` always returns an empty `ItemStack` in that case and `canAddOutput()` always returns `true` as a result.

PR #8337, opened 2025-01-19
Remove a redundant `PatternProviderMenu` constructor

62832:
Just for the sake of doing away with [more excessive add-on mixins](https://github.com/62832/MEGACells/blob/master/src/main/java/gripe/_90/megacells/mixin/PatternProviderMenuAccessor.java).


src/main/java/appeng/menu/implementations/PatternProviderMenu.java
```diff
@@ -41,8 +41,7 @@
 public class PatternProviderMenu extends AEBaseMenu {
 
     public static final MenuType<PatternProviderMenu> TYPE = MenuTypeBuilder
-            .create((id, inv, host) -> new PatternProviderMenu(id, inv, host),
-                    PatternProviderLogicHost.class)
+            .create(PatternProviderMenu::new, PatternProviderLogicHost.class)
```
kylev, line 44: Nice find!

PR #8338, opened 2025-01-19
Fix map colours for decorative (Quartz, Fluix, Sky Stone) blocks

### LINKED ISSUE #8236: Many AE2 blocks don't have map color
### Describe the bug

Including certus block (and it's varints), budding certus blocks, fluix block, sky stone block (and it's varints), ME interface, ME pattern provider.

### How to reproduce the bug

Fill a large area with any block mentioned above and create a map to see the map color.

### Expected behavior

Certus blocks should have blue map color, fluix blocks should have purple map color and so on.

### Additional details

_No response_

1.21

NeoForge

### Crash log

https://gist.github.com/YocyCraft/e4b76b7ba21731f374f89ec4cadf21d2

62832:
Closes #8236.

### DIFF
--- a/src/main/java/appeng/core/definitions/AEBlocks.java
+++ b/src/main/java/appeng/core/definitions/AEBlocks.java
@@ -112,14 +112,15 @@ public final class AEBlocks {
     private static final List<BlockDefinition<?>> BLOCKS = new ArrayList<>();
     private static final Properties QUARTZ_CLUSTER_PROPERTIES = defaultProps(MapColor.COLOR_CYAN,
             SoundType.AMETHYST_CLUSTER).forceSolidOn().strength(1.5f).requiresCorrectToolForDrops();
-    private static final Properties QUARTZ_PROPERTIES = stoneProps()
+    private static final Properties QUARTZ_PROPERTIES = stoneProps().mapColor(MapColor.COLOR_CYAN)
             .strength(3, 5).requiresCorrectToolForDrops();
-    private static final Properties SKYSTONE_PROPERTIES = stoneProps()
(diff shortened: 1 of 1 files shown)


src/main/java/appeng/core/definitions/AEBlocks.java
```diff
@@ -112,14 +112,15 @@ public final class AEBlocks {
...
             .strength(3, 5).requiresCorrectToolForDrops();
-    private static final Properties SKYSTONE_PROPERTIES = stoneProps()
+    private static final Properties SKYSTONE_PROPERTIES = stoneProps().mapColor(MapColor.COLOR_BLACK)
             .strength(5, 150).requiresCorrectToolForDrops();
     private static final StateArgumentPredicate<EntityType<?>> NEVER_ALLOW_SPAWN = (p1, p2, p3,
             p4) -> false;
-    private static final Properties SKY_STONE_CHEST_PROPS = stoneProps()
+    private static final Properties SKY_STONE_CHEST_PROPS = stoneProps().mapColor(MapColor.COLOR_BLACK)
             .strength(5, 150).noOcclusion();
+    private static final Properties FLUIX_PROPERTIES = QUARTZ_PROPERTIES.mapColor(MapColor.COLOR_PURPLE);
```
XFactHD, line 123: This will mutate the quartz properties instead of deriving a separate properties instance with purple map color from it

XFactHD (changes requested):

src/main/java/appeng/core/definitions/AEBlocks.java
62832: I hadn't accounted for that for whatever reason, my bad.

PR #8358, opened 2025-01-31
Mysterious Cube and Quartz Glass fixes for FramedBlocks

Technici4n:
- Remove unnecessary offset in Mysterious Cube model.
- Add cull faces in Mysterious Cube model.
- Also check for current appearance in Quartz Glass. This basically ensures that we only connect to adjacent blocks if they would connect with us, and it matters when the current block is not Quartz Glass, but rather something more complex.


XFactHD (approved):
:shipit:

PR #8370, opened 2025-02-11
Correct the translation of Crystal Resonance Generator

Rain-Flying:


Rain-Flying:
Why are there some missing strings on Crowdin?

PR #8411, opened 2025-03-02
Fix groovy syntax for build version generation.

kylev:
In Groovy closures (different from Java lambdas), the parameter goes inside the curly braces.
https://groovy-lang.org/closures.html#_syntax

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -123,11 +123,11 @@ if (ext.tag) {
 } else {
     // This essentially tries to implement delay expansion of the project version based on a value source to
     // still support configuration caching.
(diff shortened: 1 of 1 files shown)


kylev:
Looks like Gradle has some compatibility added to its DSL, but this gives a syntax error in VS Code and breaks the "builds green out of the box" experience.

kylev:
Doesn't prevent me from building from the command line. The syntax error makes the VS Code gradle plugin fail to initialize (so the nice clicky buttons for tasks people are used to won't be available).

HeatherComputer:
Do we know how it affects IDEA?

I'd personally like to see this merged though (and should we consider mentioning this to the gradle plugin devs?)

PR #8448, opened 2025-03-29
1.21.5

shartte:


AceLikesGhosts:
Heyo! Sorry to bug but what's the current state of this PR? Is it roughly working (enough to throw on a server?) or is it still an early stage draft? 
Thank you for making the mod btw (:

Technici4n:
The port here is purely to anticipate vanilla changes and might not be released at all. As long as it's not released things will break and we will not provide support, so please don't put it on a server.

Issue #8467, opened 2025-04-16
Spatial IO doesn't carry POI


Cassy-Lee:
### Describe the bug

Spatial IO seems doesn't storage POI (Point Of Interest) and break nether portal functionality.

### How to reproduce the bug

1. Create a nether portal.
2. Storage the nether portal using Spatial IO.
3. Execute command: "locate poi minecraft:nether_portal"

### Expected behavior

Locate command won't find any nether portal poi where the nether portal used to be.

### Additional details

POI was added in 1.14 to accelerate block searching.
Vanilla updates POI in net.minecraft.server.level.ServerLevel::onBlockStateChange (easy fix but not recommended as it will break deliberately made no-poi portal).

### Which minecraft version are you using?

1.21

### On which mod loaders does it happen?

NeoForge

### Crash log

https://pastebin.com/0aiyGPXk

Cassy-Lee:
This can be treat in the same way as the block entities.

PR #8468, opened 2025-04-16
Fix POI not updated (#8467)

Cassy-Lee:
Tested in run config "Client"


Cassy-Lee:
> I am slightly concerned about calling this for every block individually though since for a 128x128x128 spatial plot, this results in 4 million calls.

I changed it to the same as block entity moving, it should be much faster now...
I can't find any public method to list POI by chunk or section, the `getInChunk` was only visible for debug.

PR #8495, opened 2025-05-07
Adding Checking for Claims

cookta2012:
Added Checking for permissions from other mods

### DIFF
--- a/src/main/java/appeng/util/Platform.java
+++ b/src/main/java/appeng/util/Platform.java
@@ -29,6 +29,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.mojang.authlib.GameProfile;
 
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.NotNull;
 
 import net.minecraft.ChatFormatting;
 import net.minecraft.client.Minecraft;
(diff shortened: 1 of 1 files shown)


cookta2012:
Closes #8496

cookta2012:
It is a side effect of some claim protections capture the BlockEvent.BreakEvent and BlockEvent.EntityPlaceEvent for a little more of a granular level Flan and other mods like it target the events because there are times you want to prevent breaking not placing or placing and not breaking, but because mayInteract is not checked by those mods. Normally when you place a block you are supposed to fire a EntityPlaceEvent, same with BreakEvent for breaking.

cookta2012:
Matter of fact, there are some servers that just just check mayInteract. mayInteract is a vanilla spawn protection primarily

cookta2012:
https://github.com/McJtyMods/RFToolsBuilder/blob/1.20/src/main/java/mcjty/rftoolsbuilder/modules/builder/blocks/BuilderTileEntity.java#L1125

cookta2012:
https://github.com/cc-tweaked/CC-Tweaked/blob/mc-1.20.x/projects/common/src/main/java/dan200/computercraft/shared/turtle/upgrades/TurtleTool.java#L318
they check in another way by attempting to do it as the player.

cookta2012:
@shartte thoughts...

Issue #8508, opened 2025-05-17
RFC: Multi-controller networks


iczero:
### Describe the feature

Currently, ME networks are limited to a single controller structure. All channels must eventually terminate at the main "hub". ME P2P tunnels are commonly used to send large amounts of channels elsewhere. If players do not want to deal with the channel system, they can enable infinite channels in config.

I would like to propose a new mode where ME networks may have multiple ME controllers. This mode would be disabled by default in config, similar to the infinite channels mode. It would represent a middle-ground between infinite channels and the "hub" mode. Multiple controllers can be connected to each other using dense cable, at which point that cable becomes a "trunk" connection with all channels occupied. Invalid trunk configurations shut down the network in the same way that having multiple controllers normally would.

![Image](https://github.com/user-attachments/assets/01d58434-8b1f-4ed9-abfb-0203f0eb8e95)

### Reasons why it should be considered

In Ethernet networking, many switches are typically present in a single broadcast domain. These switches may be connected to each other, and end devices connected to any of these switches may communicate with each other. PCIe switches operate in a similar manner.

From a gameplay perspective, this allows more decentralized networks to be feasible. For example, if a player wanted to migrate their house somewhere else, they could do so by building a second controller at the new house, and then bridge it with the existing controller. Centralized networks are highly reliant on ME P2P tunnels, but configuring and keeping track of these tunnels is quite annoying. I personally do not find the gameplay of wiring up a hundred or so P2P tunnels very fun.

An implementation is available here: https://github.com/iczero/Applied-Energistics-2/tree/multi-controller (https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/8509)

cplir-c:
How did you account for the changing network structure when chunks are unloaded and loaded again? I saw an error for invalid trunk connection on a quick peek, but didn't look closely.

Technici4n:
Having a single controller is a design decision to force the player to think about network layout. Making p2p configuration nicer is a separate thing, and we agree that it would be good to do.

handicraftsman:
Would absolutely love this feature, especially in layouts where you have several bases of different players talking to each other.

iczero:
Regarding https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/8509#issuecomment-2890411653:

The feature flag was added because this was designed to be a prototype. In the future, a different type of controller could be used, perhaps named as a "router", and enabled by default.

[...]

PR #8509, opened 2025-05-17
Multi-controller networks initial implementation (#8508)

iczero:
Initial implementation of the mechanics described in #8508.

Controller trunk links are discovered during the BFS stage of network pathing. A valid tree should always be found if trunk connections only have two controllers.


Technici4n:
Hi, we will not accept this feature because it goes against AE2's current game design. That's why you made it disabled by default. However we have no interest in adding more feature flags which mean that eventually everyone is playing a slightly different flavor of the mod depending on their pack.

PR #8541, opened 2025-06-08
Fix Russian translation for crafting time

### LINKED ISSUE #8532: ae2 bug timer localization
### Describe the bug

In the Russian localization, you translated "gui.ae2.ETAFormat", which is why the system cannot render it correctly.
Image

This parameter is incorrect only in the Russian localization. ru_ru.json
Image

### How to reproduce the bug

Image

Avtokraft Interface

### Expected behavior

"gui.ae2.ETAFormat": "Час:мин:сек" ->"gui.ae2.ETAFormat": "HH:mm:ss"

### Additional details

_No response_

1.20

Forge

### Crash log

-

Mithi83:
Fixes #8532


Mithi83:
You might want to cherrypick that to the 1.20 fabric branch as well.

PR #8550, opened 2025-06-11
Optimized watcher update and ME IO Port Fill/Empty operations

### LINKED ISSUE #8539: Frequent Hash collision and redundant lookups in `StorageService.updateCachedStacks()`
### How to reproduce the performance issue

1. Install AE2 `19.2.11` and Spark `1.10.124-neoforge-1.21.1` via Modrinth.
2. Create a new Superflat world with The Void preset in creative mode.
3. Set up an ME network with more than 32 ME Drives, all full of 1k ME Storage Components.
4. Connect an ME Level Emitter with the filter set to Bone Meal and the threshold set to 1.
5. Populate the ME network with a lot of Netherite Hoe with random `damage` and random `repair_cost`.
6. Set up an apparatus that constantly exports and imports the Bone Meal, and put a Bone Meal into the ME network.
7. Start the Spark profiler by `/spark profile`.
8. See what happens.

### Spark profile link

https://spark.lucko.me/pJo1dIk6o7

1.21

NeoForge

### Additional details

When I was profiling and debugging the AE2, I noticed that some Netherite hoes share the same hash value.
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/452216e8228ea5d979887d353aa228e7166e71d5/src/main/java/appeng/api/stacks/AEItemKey.java#L60

I dived deep into this `ItemStack.hashItemAndComponents` function, and I found out that the vast majority of `Reference2ObjectFunction<K, V>.hashCode()` implementations only does a simple hash calculation of each entry.
So simple that it causes a lot of hash collisions among Netherite hoes, even with just a few differences in terms of `damage` and `repair_cost`.
Such a high probability of hash collision causes a lot of slow lookups in hash maps, which leads to even slower, detailed `ItemStack` comparisons. In the Spark profile, `appeng.api.stacks.AEItemKey.equals()` calls in `it.unimi.dsi.fastutil.objects.Object2LongOpenHashMap.getLong()` shows how frequent the hash collisions are.
Even without the hash collision issue, hash calculation on the whole components would be significantly slower than the one on the component patches.

I also noticed that `appeng.me.service.StorageService.updateCachedStacks()` has 3 loops. One for notifying the watchers the updates in currently available stacks, one for notifying the watchers the removed stacks, and the last one for updating everything in `cachedAvailableAmounts`.
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/452216e8228ea5d979887d353aa228e7166e71d5/src/main/java/appeng/me/service/StorageService.java#L109-L145

`cachedAvailableAmounts.put(entry.getKey(), entry.getLongValue())` returns the old value, so we don't need to look up the `cachedAvailableAmounts` more than once.

The second loop calls `cachedAvailableStacks.get(what)` for every single items that the `StorageService` observed in the ME storage in the previously tick. I would rather create a `ObjectSet<AEKey>` for this purpose, and only iterate the items that the `cachedAvailableStacks` doesn't contain.

So I forked the AE2 and created an almost-PR-ready optimization prototypes [here](https://github.com/MineCake147E/Applied-Energistics-2/tree/main-optimizations).
* Introduces a custom `ItemStack` hash calculation algorithm that
  * Outputs a 64-bit hash instead of 32-bit hash.
  * Suffices the existing requirements, like being oblivious to the registration order of the component patches' elements.
* Separates the watcher notification logic from the `StorageService.updateCachedStacks()` entirely, since it is never needed while `interestManager.isEmpty()` is true, and `cachedStacksNeedUpdate` only becomes true when the `interestManager.isEmpty()` is true.
* Optimizes the now separate watcher notification logic (`StorageService.notifyWatchers()`) by leveraging the `ObjectOpenHashSet<AEKey>`, which is cached in the `StorageService` instance.

The Spark profile after optimization: https://spark.lucko.me/90I4MugraO

MineCake147E:
- Separated watcher update from `StorageService.updateCachedStacks()`.
  - Optimized the now separate watcher update code.
  - Watcher updates now only runs in `StorageService.onServerEndTick()`
- Replaced the `ItemStack` hash calculation in `AEItemKey` with a new one with more entropy, dramatically reducing the chance of the slow lookup in hash table, and `ItemStack.isSameItemSameComponents` calls.
  - Added `AEKey.hashCode64()` for improved-entropy hashes.
  - `FuzzySearch.KeyComparator` now uses `AEKey.hashCode64()` for finer tiebreaker.
- Replaced the `FluidStack` hash calculation in `AEFluidKey` with a new one with more entropy, reducing the chance of the slow lookup in hash table, and `FluidStack.isSameFluidSameComponents` calls.
- Optimized ME IO Port Fill/Empty operations (~10x faster in some scenario).
  - Added several filtering APIs for improved filling performance.

Resolves #8539


Technici4n (changes requested):
Thanks for sending the PR. Some notes:
- The IO port changes should be split off to a separate PR. But the patch is too long... I wouldn't accept anything that increases the code size so much.
- Widening `hashCode` to a `long` should not make a noticeable difference.
- The new hash code computation sounds like it could make a big difference, but besides its size, do you have some examples of problematic cases that this fixes? I am thinking of say an example where two stacks are different but have the same hash code.

MineCake147E:
> * The new hash code computation sounds like it could make a big difference, but besides its size, do you have some examples of problematic cases that this fixes? I am thinking of say an example where two stacks are different but have the same hash code.

For example, with old algorithm, `{id:"minecraft:netherite_hoe",components:{"minecraft:damage":643,"minecraft:repair_cost":644}}` and `{id:"minecraft:netherite_hoe",components:{"minecraft:damage":472,"minecraft:repair_cost":449}}` have the same hash code `0x6116_8C69`.
This is just one example of thousands of different colliding combinations of `damage` and `repair_cost` among `minecraft:netherite_hoe`.
Another example of hash collision is `0x6CCF_3935` from `{id:"minecraft:bow",count:1,components:{"minecraft:damage":311,"minecraft:enchantments":{levels:{"minecraft:power":3,"minecraft:unbreaking":3}}}}` and `{id:"minecraft:bow",count:1,components:{"minecraft:damage":340,"minecraft:enchantments":{levels:{"minecraft:power":4,"minecraft:unbreaking":3}}}}`, both being survival-obtainable item via fishing.
Even with the same item with different damage, enchantments, and repair cost, the hash collides so easily. That alone signifies how frequently the items' hash code collide each other.
But nearly all the collisions are gone with the new hash algorithm.

> * The IO port changes should be split off to a separate PR. But the patch is too long... I wouldn't accept anything that increases the code size so much.

I updated the code accordingly.
I stress-tested the update logic using IO port, but IO port was hiding the performance difference of watcher update logic.
So I had to optimize that for a better stress-testing experience.
I will post a separate PR for that matter, after simplifying the code a little bit.

MineCake147E:
> Is there maybe a way that you can provide the list of items in the system? I would like to get this fixed in NeoForge if possible, with a targeted fix. A `/ae2 grids export` of the big grid should contain all the items and be a useful reference. 🤔

Here is the one with many colliding Netherite hoes. I hope that helps.
[ae2_grid_6_20250613_1832.zip](https://github.com/user-attachments/files/20723918/ae2_grid_6_20250613_1832.zip)

Technici4n:
Thanks a lot for this data. Thanks to it, I managed to find a strategy for a simpler hashCode fix. The goal is to implement it in NeoForge directly, see: https://github.com/neoforged/NeoForge/pull/2332.

PR #8561, opened 2025-06-14
Reduce collisions in AEItemKey hashing

Technici4n:
The main trick we play is how we hash the component patch map entries:
instead of XOR'ing the hash codes of the keys and the values,
we multiply the values by 31 ^ id, where id is the registered id of the data component.

Partially addresses https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/8539 and https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/8550.


Technici4n:
Fixed in NeoForge.

PR #8590, opened 2025-07-11
Balance annihilation plane enchantment energy cost

### LINKED ISSUE #8188: Annihilation Plane has chance to get stuck on hard blocks when enchanted with unbreaking
### Describe the bug

Oddly specific but if an annihilation plane is enchanted with unbreaking every few block placements it will get stuck and take up to about 30 seconds to mine it.

The enderchest also does this (Maybe to do with hardness?). Haven't tested with many other blocks.

Without unbreaking enchantment it is flawless. My guess is if the roll/chance to not use energy is true something causes this to happen.

### How to reproduce the bug

Connect annihilation plane enchanted with unbreaking to the network, and keep placing obsidian. You will see the block sometimes sticks around for a long time. 

### Expected behavior

Breaks the blocks consistently.

### Additional details

I am on 1.20.1
I know this is a nitpick but I was trying to just quickly farm obsidian by collecting lava

1.20

Fabric

### Crash log

https://pastebin.com/A6jrWDew

Mithi83:
The current implementation had two problems:

Enchanting an annihilation plane with any level of efficiency (and nothing else) would reduce the required energy to zero due to multiplying with the 0 in levelSum directly.

Enchanting an annihilation plane with unbreaking would increase the energy cost by a factor of 8 per level of unbreaking, but only leading to an average saving of 50% (lv1), 33% (lv2), or 25% (lv3), effectively increasing the average cost rather than reducing it.

Fixes #8188

### DIFF
--- a/src/main/java/appeng/parts/automation/ItemPickupStrategy.java
+++ b/src/main/java/appeng/parts/automation/ItemPickupStrategy.java
@@ -271,8 +271,8 @@ public class ItemPickupStrategy implements PickupStrategy {
                 useEnergy = randomNumber == 0;
             }
             var levelSum = enchantments.entrySet().stream().map(Map.Entry::getValue).reduce(0, Integer::sum)
-                    - efficiencyLevel;
-            requiredEnergy *= 8 * levelSum * efficiencyFactor;
+                    - efficiencyLevel - unbreakingLevel;
+            requiredEnergy *= (levelSum > 0 ? 8 * levelSum : 1) * efficiencyFactor;
         }
 
         return useEnergy ? requiredEnergy : 0;


src/main/java/appeng/parts/automation/ItemPickupStrategy.java
```diff
@@ -271,8 +271,8 @@ protected float calculateEnergyUsage(ServerLevel level, BlockPos pos, List<ItemS
                 useEnergy = randomNumber == 0;
             }
             var levelSum = enchantments.entrySet().stream().map(Map.Entry::getValue).reduce(0, Integer::sum)
```
62832, line 273: Just a minor nitpick, but while you were here you could have probably condensed this unnecessary stream mapping to `enchantments.values().stream().reduce[...]`. I'm not sure what I was thinking at the time going for the round-about entry set allocation.

Mithi83: This was the way the code was when you wrote it. Shartte changed it in the 1.20.5 port in 70838d74f81917ece64335db2aca3cb3a604b780
Your suggestion doesn't compile in the current code base, as enchantments is no longer a `Map<Enchantment, Integer> ` but rather became `ItemEnchantments`.

62832: Oh yeah, I completely forgot about that. Thanks for reminding me. 😅

PR #8671, opened 2025-09-22
Update es_es.json and create es_mx.json

ArrivedBog593:


Mari023:
We only accept translations on Crowdin. https://appliedenergistics2.crowdin.com/applied-energistics-2

PR #8706, opened 2025-11-03
Automatic pattern terminal restock

Satherov:
Automatically pulls a blank pattern from the ME System's Storage when the Pattern Terminal runs out

https://github.com/user-attachments/assets/a2fbd52f-24c4-4d66-9242-b764d41901e0


Mithi83 (commented):
I think this patch is wrong after a quick review, without having tested it yet.
In my option you forgot to handle the case that you can‘t extract a blank pattern from the system. You remove the `set(ItemStack.EMPTY)` from line 296 entirely. That should probably be moved into an `else` in line 309.

PR #8716, opened 2025-11-10
(26.1.2) Use IBlockExtension#getCloneItemStack instead of new ItemStack(Block)

Frinn38:
This PR implements a TODO in PatternContainerGroup by using NeoForge `IBlockExtension#getCloneItemStack` instead of a simple `new ItemStack` for finding the item to display in the pattern access terminal.

This gives the targeted block more context (level, pos) so the returned item should be more correct, as currently blocks that rely on their BlockEntity to return an item are displayed incorrectly.


Technici4n (approved):
Sounds good, thanks

src/main/java/appeng/api/implementations/blockentities/PatternContainerGroup.java
```diff
@@ -112,9 +114,9 @@ public static PatternContainerGroup fromMachine(Level level, BlockPos pos, Direc
             }
         } else {
             // Try to wrestle an item from the adjacent block entity
-            // TODO On Forge we can use pick block here
             var targetBlock = target.getBlockState().getBlock();
-            var targetItem = new ItemStack(targetBlock);
+            var targetItem = targetBlock.getCloneItemStack(target.getBlockState(),
+                    new BlockHitResult(Vec3.atCenterOf(pos), side, pos, false), level, pos, null);
```
andrewsf: At least 3major mods ([KubeJS](https://github.com/KubeJS-Mods/KubeJS/blob/2601/src/main/java/dev/latvian/mods/kubejs/core/mixin/IBlockStateExtensionMixin.java#L30), EnderIO, [IntegratedDynamics](https://github.com/CyclopsMC/IntegratedDynamics/blob/47fd4c319a388ebd4af2561d107386bdaaafee46/src/main/java/org/cyclops/integrateddynamics/block/BlockCable.java#L254)) have code expecting a non-null player and have seen crashes from null passers ([Just Dire Things](https://github.com/Direwolf20-MC/JustDireThings/issues/501)). NeoForge does not document the player paramter as nullable.

Frinn38: As suggested on the AE2 discord I used a FakePlayer instead of null here, I think it should be ok for mods expecting a non-null player.

PR #8723, opened 2025-11-19
Fix: Prevent Pattern Provider item loss on simultaneous slot insertions

emroiST:
Modified GenericStackItemStorage.insertItem() to automatically search for available slots when the requested slot cannot accept the full stack. This prevents item loss when external mods (e.g., Industrial Upgrade ejector) insert multiple items targeting the same slot number.

The method now:
1. Attempts insertion into the specified slot first (maintains compatibility)
2. Iterates through remaining slots if insertion is incomplete
3. Distributes items across available slots automatically

Backward compatible - only enhances behavior in failure cases.

### DIFF
--- a/src/main/java/appeng/helpers/externalstorage/GenericStackItemStorage.java
+++ b/src/main/java/appeng/helpers/externalstorage/GenericStackItemStorage.java
@@ -48,6 +48,19 @@ public class GenericStackItemStorage implements IItemHandler {
 
         var inserted = (int) inv.insert(slot, what, stack.getCount(), Actionable.ofSimulate(simulate));
 
+        if (inserted < stack.getCount()) {
+            var remaining = stack.getCount() - inserted;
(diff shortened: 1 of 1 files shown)


Mari023:
I don't fully understand the problem yet, but I am a bit confused about this fix.

1. Some mod inserts some items into e.g. a slot in a pattern provider
2. the slot is already partially filled, so it can't insert the whole stack.
3. AE2 correctly reports the leftover that could not be inserted.
4. It is now on the other mod to ensure the leftover stack is not voided.

I fail to see how this should be AE2's responsibility here.
It seems more like this is a problem with Industrial Upgrade not handling the leftover correctly (or not handling it at all).

it also seems like this could never fully fix the problem, only mitigate it at bit:
e.g. if the inventory is already nearly full, and it can't find more free slots, it would still void items.

shartte:
Yeah no, not going to do that. Inserts into specific slots will not suddenly start inserting into all other slots.

Issue #8592, opened 2025-07-13
Duplication Glitch with Storage Drawers Mod in 1.21


Nifiore:
### Describe the bug

Technically not really duplication in the strictest sense, but this bug allows the player to convert any item into any other item, provided they have at least 1 of the thing they want more of. Notably, this happens with storage drawers but not functional storage. So this may be something on their end and I will be making a report there too. 

### How to reproduce the bug

This bug only requires AE2 and Storage Drawers

Fill a 1x2 or 2x2 storage drawer with a large amount of any item in any slot except the first slot. Let's call this item 1.  I used stone. 
Attach an import bus to that drawer and set it to blacklist only item 1 and fill it with acceleration cards. 
Hook the network up to some type of storage so the import bus will actually pull items. I used a second drawer.
Place 1 item that you want more of in a drawer slot before that of item 1. 

The minimum condition for this to work seems to me that the import bus attempts to pull more of item 2 than is in the drawer and starts pulling item count from item 1 instead, resulting in the creation of more item 2.

I have provided some pictures to help. In this set of pictures, I placed a single diamond block in the top left slot of the 2x2 drawer and it imported 96 diamond blocks into the 1x1 drawer while decreasing the quantity of stone by 95. I am @relas. on discord if you'd like to contact me there too. 

<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/823c0253-2a19-4df6-b6ef-e5fd545232c1" />
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/6405e14a-964e-491f-93c0-5323c7950120" />
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/6e3f6e01-078f-4a0f-8524-1f38d40eab4f" />
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/5c48c8f8-b269-46ed-bb12-29011f73c80a" />
<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/4c0eea48-7f55-4433-8ba3-919df8ef4799" />

### Expected behavior

It should've only imported 1 diamond block and not changed the quantity of stone in the 2x2 drawer 

### Which minecraft version are you using?

1.21

### On which mod loaders does it happen?

NeoForge

### Crash log

https://pastebin.com/HQKif1Tv

Technici4n:
[...]

PR #8734, opened 2025-12-04
Fix: item duplication

Frostbite-time:
The `extractFromHandler` method currently assumes the item type in a given slot does not change between extractions. However, for some containers with auto-sorting or a dynamic number of slots, this can lead to item duplication.

This issue has already caused AE2 to duplicate items when import bus work for my mod’s container (see Frostbite-time/BeyondDimensions#19). It also caused the duplication reported in #8592.

This PR is a simple fix,  just one line,  and is very easy to review.


Technici4n:
I would move the check below `totalExtracted += extracted` to avoid checking in the first iteration. Otherwise seems fine

Frostbite-time:
Sure, I just added a commit to do that

Technici4n (approved):

PR #8735, opened 2025-12-04
Improve pattern provider grouping

### LINKED ISSUE #8732: Molecular Assembler mislabeled in Pattern Access Terminal if it has uneven number of speed upgrades
### Describe the bug

Molecular Assembler appears as Pattern Provider when the Pattern Provider has multiple Molecular Assembles connected to it and at least one of them has different amount of speed upgrades.

Setup: tower, 4 assemblers per provider, 8 providers.

Image

In pattern access terminal with no speed upgrades in the tower:

Image

With a single speed upgrade in an assembler on the lowest level:

Image

### How to reproduce the bug

1. Install an array of molecular assemblers and pattern providers in a way that each provider has > 1 assemblers. Eg checkerboard or tower.
2. Add one (1) speed upgrade.

### Expected behavior

1. Pattern access terminal properly shows the machines as a molecular assembler.
2. Able to shift click patterns in the terminal into the molecular assembler group.

### Additional details

AE2: 19.2.17
Modpack: ATM10 10-5.2
NeoForge: 21.1.115

1.21

NeoForge

### Crash log

-

Mithi83:
If different group entries are available still use the more specific icon if it is the same for all group entries.

Fixes #8732

### DIFF
--- a/src/main/java/appeng/helpers/patternprovider/PatternProviderLogic.java
+++ b/src/main/java/appeng/helpers/patternprovider/PatternProviderLogic.java
@@ -746,12 +746,19 @@ public class PatternProviderLogic implements InternalInventoryHost, ICraftingPro
             return groups.iterator().next();
         }
 
+        // If nothing is adjacent, just use itself
+        var icon = this.host.getTerminalIcon();
         List<Component> tooltip = List.of();
         // If there are multiple groups, show that in the tooltip
         if (groups.size() > 1) {
             tooltip = new ArrayList<>();
             tooltip.add(GuiText.AdjacentToDifferentMachines.text().withStyle(ChatFormatting.BOLD));
+            // If all groups share the same icon use that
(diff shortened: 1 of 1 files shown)


Mithi83:
Example with different combinations:

1. Furnace + Molecular Assembler -> Generic Pattern Provider (as before), verbose tooltip (as before)
2. 4 Molecular Assemblers with 5 Acceleration Cards each -> Molecular Assembler (as before), slim tooltip (as before)
3. 4 Molecular Assemblers with no upgrades -> Molecular Assembler (as before), no tooltip (as before)
4. 4 Molecular Assembler with 0, 2, 3, and 5 Acceleration Cards respectively -> Molecular Assembler (this has changed), verbose tooltip (as before)

image

For the review: I'm not sure if `if (!icon.getId().equals(group.icon().getId()))` is the best way to check for that equality, feel free to suggest an improved check.

PR #8756, opened 2026-01-02
Secondary Outputs are now listed and available to craft

Fernthedev:
This seems way too easy and simple, so I'm expecting something to be off that I'm not seeing as of writing this. I don't understand the entire machine that is AE2 to look for these bugs though.

Additionally, I imagine `PatternProviderLogic.onPushPatternSuccess` needs to be modified somehow to support this too, but I don't fully understand its purpose or how it needs to be reworked (if required) for this to work.

### DIFF
--- a/src/main/java/appeng/me/service/helpers/NetworkCraftingProviders.java
+++ b/src/main/java/appeng/me/service/helpers/NetworkCraftingProviders.java
@@ -200,7 +200,6 @@ public class NetworkCraftingProviders {
                 for (var output : outputs) {
                     methods.craftableItemsList.add(output.what(), 1);
 
-
                     methods.craftableItemsList.add(output.what(), 1);
 
                     var patternsForKey = methods.craftableItems.computeIfAbsent(output.what(),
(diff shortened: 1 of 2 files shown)


Technici4n:
Making them secondary was a conscious design decision, just going back to all outputs being primary is not a good solution.

Mithi83:
Having a primary output is useful for steering AE2 to use a specific recipe for a specific product.

Example: recipe 1 (A -> X + Y) and recipe 2 (B -> Y). With primary outputs you can encode pattern 1X as (A -> X + Y) and pattern 1Y (A -> Y + X). Suppose you want to craft a bunch of Y, AE2 now has 2 options, using pattern 1Y or pattern 2. Suppose further that A is a very valuable and rare resource, whereas B is abundant. You‘d want to use B whenever you craft Y and not waste A. To achieve that you could just remove pattern 1Y. If you craft X AE2 knows it can use Y from 1X in further crafting steps in the crafting plan, but it will not waste A just to craft Y.

It is basically a tradeoff between ease-of-use and the ability to fine-tune your recipes.

Fernthedev:
That's fair, but what about the possibility of making a recipe have a N-primary outputs then? Obviously this PR isn't that but could it be considered to have a UI mechanism to state that a recipe can output more than 1 item as its done here? Not that I'm suggesting to do it either anytime soon, but I figure I should ask for future reference.

PR #8764, opened 2026-01-05
Perf: optimize crafting calculation for multi-branch patterns

Gabriel-Hiss:
### Changes:
- **Exponential batch processing**: Multi-branch crafting now uses exponential probing (1, 2, 4, 8...) instead of processing one item at a time, reducing complexity from O(n) to O(log n)
- **Object pooling**: Added ThreadLocal pooling for `ChildCraftingSimulationState` to eliminate repeated allocations
- **Pattern caching**: Cache `getCraftingFor()` results in `CraftingCalculation` to avoid duplicate lookups
- **Reduced sync overhead**: Increased `PAUSE_CHECK_INTERVAL` from 100 to 1000
- **Optimized binary search**: CRAFT_LESS strategy now uses exponential search to find upper bound before binary search
- **applyDiff optimization**: Skip entries with zero delta

### Performance Impact

For crafts with multiple pattern branches:

| Metric | Before | After |
|--------|--------|-------|
| Multi-branch iterations | O(n) | O(log n) |
| Object allocations | O(n) | O(1) |
| Pattern lookups | 2x per node | 1x per node |

### DIFF
--- a/src/main/java/appeng/crafting/CraftingCalculation.java
+++ b/src/main/java/appeng/crafting/CraftingCalculation.java
@@ -122,40 +122,25 @@ public class CraftingCalculation {
         }
 
         if (strategy == CalculationStrategy.CRAFT_LESS) {
-            // Exponential search to find upper bound, then binary search.
-            long upperBound = 1;
-            ICraftingPlan lastSuccessful = null;
-
-            while (upperBound < requestedAmount) {
-                var plan = runCraftAttempt(false, upperBound);
-                if (plan != null) {
-                    lastSuccessful = plan;
-                    upperBound *= 2;
-                } else {
-                    break;
+            // Try crafting less if possible using binary search.
+            long successfulAmount = 0;
+            ICraftingPlan successfulPlan = null;
+            for (long increment = Long.highestOneBit(requestedAmount); increment > 0; increment /= 2) {
+                long testAmount = successfulAmount + increment;
+                if (testAmount < requestedAmount) {
+                    var plan = runCraftAttempt(false, testAmount);
+                    if (plan != null) {
+                        // Success! :)
+                        successfulAmount = testAmount;
+                        successfulPlan = plan;
+                    }
                 }
             }
 
-            if (lastSuccessful == null) {
-                return runCraftAttempt(true, requestedAmount);
+            // Found a successful plan! :)
+            if (successfulPlan != null) {
+                return successfulPlan;
             }
-
-            long low = lastSuccessful.finalOutput().amount();
-            long high = Math.min(upperBound, requestedAmount - 1);
-            ICraftingPlan bestPlan = lastSuccessful;
-
-            while (low < high) {
-                long mid = low + (high - low + 1) / 2;
-                var plan = runCraftAttempt(false, mid);
-                if (plan != null) {
-                    bestPlan = plan;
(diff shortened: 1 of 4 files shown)


Gabriel-Hiss:
i will submit some videos of the before and after

Technici4n:
Hi, thanks for sending a PR. Exponential probing for multiple branches seems like a good idea. However, I am not convinced by the rest. In particular, the "optimized binary search" just makes the code longer but does not optimize anything? Your PR description looks very AI generated, I sure hope the code isn't as well.

Gabriel-Hiss:
https://github.com/user-attachments/assets/1af0b86e-2ee5-4122-b872-61d0f7a14e8b

Gabriel-Hiss:
i'm not a native english speaker and i asked chatgpt help to generate a good pr summary

Gabriel-Hiss:
Let's suppose the value that we are trying to find is 100 then the exponential goes on till 64 which is a pass then 128 is a fail then we do a binary search to get the exact value It is literally a reverse exponential search

if we remove the binary we will probably lose some "possible crafts"

Technici4n:
A 40% improvement is unfortunately not that good. Do you have a profile of the calculation code?

src/main/java/appeng/crafting/CraftingCalculation.java
```diff
@@ -104,25 +122,40 @@ private ICraftingPlan computePlan() throws InterruptedException {
...
-            ICraftingPlan successfulPlan = null;
-            for (long increment = Long.highestOneBit(requestedAmount); increment > 0; increment /= 2) {
-                long testAmount = successfulAmount + increment;
-                if (testAmount < requestedAmount) {
-                    var plan = runCraftAttempt(false, testAmount);
-                    if (plan != null) {
-                        // Success! :)
-                        successfulAmount = testAmount;
-                        successfulPlan = plan;
-                    }
```
Technici4n, line 118: This is already a correct and compact binary search, I don't see why you modified it.

Gabriel-Hiss:
> A 40% improvement is unfortunately not that good. Do you have a profile of the calculation code?

my bad. i was thinking about the multi-branch case where it might fail early but didn't actually profile it to see if it helped. the original code is cleaner anyway.

Gabriel-Hiss:
i will revert this change. The other optimizations (pooled child states + exponential probing in request()) should still be good though.

Gabriel-Hiss:
i'm making the profiles for comparison

Gabriel-Hiss:
you want the profiling files too or i can send a ss?

Gabriel-Hiss:
i will send later is too late rn

Gabriel-Hiss:
image
Before
[before.zip](https://github.com/user-attachments/files/24459568/before.zip)
the jfr is on the zip the craft amount was 100x

Gabriel-Hiss:
the after is with your binary serch

Gabriel-Hiss:
i will capture the before again cuz that is own my survival world and might have some noise

Gabriel-Hiss:
image
[before_2.zip](https://github.com/user-attachments/files/24459958/before_2.zip)

before run 1 had some noise

Gabriel-Hiss:
i'm making a profile on the 1000x items request

PR #8766, opened 2026-01-05
Update ru_ru.json + guidebook (_ru_ru)

deaddiesel:


deaddiesel:
I ask you to merge my localization files in my language (the main file + the complete guide), as I believe that Crowdin does not reveal the potential that mod localizers use, especially since the guide on the Crowdin site has not yet been fully translated into my language and it is not yet clear when it will be ready, it is The translation system is too complicated and too obscure. It's faster to translate manually using notepad. **As they say - No need for this dancing with a tambourine.**

Issue #8675, opened 2025-09-28
Spatial I/O fails to use pylons cross-dimensionally correctly.


ShadowzOfTheDark:
### Describe the bug

A spatial I/O port in a different dimension than the pylons will still use the pylons but swap the blocks in the dimension the spatial I/O port is in, not the dimension the pylons are in.

https://www.youtube.com/watch?v=kC16_5dX-dc

### How to reproduce the bug

Build pylons in say, the overworld. 

Place a quantum link chamber between the overworld and the nether. Give it the basic AE2 stuff to link it all and chunkload it.

Place a spatial I/O port in the nether.

Activate the spatial I/O port in the nether and it will swap into spatial storage in the nether where the pylons are in the overworld.

### Expected behavior

Spatial I/O ports in a different dimension than the pylons should still swap the contents of the pylons in the dimension the pylons are in.

### Which minecraft version are you using?

1.20

### On which mod loaders does it happen?

Forge

### Crash log

https://pastebin.com/raw/uirMAir0

Mari023:
that is a really well done video explaining the issue very well!
also quite a funny bug :D

ProgramCrafter:
Singleplayer test succeeded. I do not have a dedicated server at hand at the moment, unfortunately.

PR #8770, opened 2026-01-09
fix: teleport area from pylons' world and not IO port's (#8675)

ProgramCrafter:


tehgreatdoge:
Have you tested this on a dedicated server? That diff looks sus

ProgramCrafter:
I have not (yet, though I would be grateful if someone who already has a prepared environment tested it).

The intent is to load ServerLevel from the spatial service, as returned by its implementation https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/fd8b717a405672ce4f65ba540f1db8c91317daa4/src/main/java/appeng/me/service/SpatialPylonService.java#L199-L202; it will correspond to the pylons rather than the I/O port. Is there any reason why it would not work?

ProgramCrafter:
Singleplayer test with my fix succeeded. I do not have a dedicated server at hand at the moment, unfortunately.

HaphLife:
Yay, I'm glad this bug is getting fixed. That looks like the correct approach to me.

The problem was that the teleportation occurred in the dimension of the spatial IO port, rather than the dimension of the pylons, so using the SPS as a proxy of the pylons to replace the spatial IO port looks correct.

SPS determines the value you get by the first pylon on it's internal list, and skips any pylons in other dimensions when determining it's volume, so the dimension should always correspond to the volume it found, and thus be appropriate to use here.

PR #8789, opened 2026-02-07
Logic error fix for CraftingCpuLogic.java

### LINKED ISSUE #8781: CraftingCPULogic::executeCrafting() logic error gives infinite co-processor behavior when two exit conditions collide.
### Describe the bug

[CraftingCPULogic's lines 236 - 243](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/fd8b717a405672ce4f65ba540f1db8c91317daa4/src/main/java/appeng/crafting/execution/CraftingCpuLogic.java#L236) has two conditions to check exit conditions to skip the rest of that iteration of the loop:

```
if (task.getValue().value <= 0) {
    it.remove();
    continue taskLoop;
}

if (pushedPatterns == maxPatterns) {
    break taskLoop;
}
```

However, if the first condition (that the last copy of a particular recipe is executed) is coincident with the second condition (that we hit the limit of patterns which can be pushed this tick), the second condition is skipped.

Skipping the second condition causes the loop to both not break this iteration (incorrect behavior) and not break on any subsequent iteration (unless it overflows and loops all the way back around through every integer value), behaving as if the CPU cluster had "infinite" co-processors for a tick.

### How to reproduce the bug

This behavior can occur in-game whenever those two conditions coincide, which is particularly easily to cause if executing a crafting plan with many different recipes each executed once.

### Expected behavior

I would expect that when both conditions are true, the task gets removed *and* the loop breaks, rather than continuing.

### Additional details

A simple fix for this bug would be to just add a copy of the pushedPatterns==maxPatterns check between the remove() and the continue.

1.21

Forge

### Crash log

n/a

HaphLife:
Fixes #8781 by adding an additional check to break before continuing.

Probably not the most elegant solution, as it simply repeats the second check within the first one (thus avoiding skipping the second check) but it is the simplest one I could think of.

After building and running AE2 with the change, this branch appears to fix the incorrect behavior.

### DIFF
--- a/src/main/java/appeng/crafting/execution/CraftingCpuLogic.java
+++ b/src/main/java/appeng/crafting/execution/CraftingCpuLogic.java
@@ -235,6 +235,11 @@ public class CraftingCpuLogic {
                     task.getValue().value--;
                     if (task.getValue().value <= 0) {
                         it.remove();
+
+                        if (pushedPatterns == maxPatterns) {
+                            break taskLoop;
+                        }
+
                         continue taskLoop;
                     }
 


src/main/java/appeng/crafting/execution/CraftingCpuLogic.java
```diff
@@ -235,6 +235,11 @@ public int executeCrafting(int maxPatterns, CraftingService craftingService, IEn
                     task.getValue().value--;
                     if (task.getValue().value <= 0) {
                         it.remove();
+
```
andrewsf, line 238: Alternative that scopes the `continue` more specifically & keeps the `break` logic to one spot

```
                    if (task.getValue().value <= 0) {
                        it.remove();

                        if (pushedPatterns != maxPatterns) {
                            continue taskLoop;
                        }
                    }

                    if (pushedPatterns == maxPatterns) {
```

PR #8799, opened 2026-02-26
Fix jade compat of remove FE and Fluid tooltip

Frostbite-time:
The calls to `tooltip.remove(JadeIds.UNIVERSAL_ENERGY_STORAGE_DETAILED)` and `tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE_DETAILED)` currently have no effect. As a result, many block entities that extend `AEBaseBlockEntity` still display both FE and AE energy at the same time. This is especially noticeable for addon authors like myself.

In addition, even AE2’s own blocks can end up with inconsistent tooltips. For example, the Inscriber requires a direction to be provided when querying its energy capability; otherwise the query returns `null`. Because of that, Jade can’t obtain the information and ends up showing only AE energy. Meanwhile, blocks like the ME Controller will show both FE and AE energy.

I also adjusted the fluid-tooltip removal so that it actually takes effect. However, I’m not sure whether you intentionally wanted to remove the fluid tooltip in the first place, since fluid information doesn’t conflict with other tooltip lines. If that wasn’t the intention, you can simply remove that line.

before:
bc3be7bfdc460126341c8167f82b5b74
b4238b5be771714107ca1774f7c90eeb

after:
d45d17de22dc1db395ff72a51f679ace
83ba99272b70c0619d749bb30faf711d

### DIFF
--- a/src/main/java/appeng/integration/modules/jade/BodyProviderAdapter.java
+++ b/src/main/java/appeng/integration/modules/jade/BodyProviderAdapter.java
@@ -29,7 +29,6 @@ class BodyProviderAdapter<T extends BlockEntity> extends BaseProvider implements
     @Override
     public void appendTooltip(ITooltip tooltip, BlockAccessor accessor, IPluginConfig config) {
         tooltip.remove(JadeIds.UNIVERSAL_ENERGY_STORAGE);
-        tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE);
 
         var context = ContextHelper.getContext(accessor);
         var tooltipBuilder = new JadeTooltipBuilder(tooltip);
--- a/src/main/java/appeng/integration/modules/jade/BodyProviderAdapter.java
+++ b/src/main/java/appeng/integration/modules/jade/BodyProviderAdapter.java
@@ -28,8 +28,8 @@ class BodyProviderAdapter<T extends BlockEntity> extends BaseProvider implements
 
     @Override
     public void appendTooltip(ITooltip tooltip, BlockAccessor accessor, IPluginConfig config) {
-        tooltip.remove(JadeIds.UNIVERSAL_ENERGY_STORAGE_DETAILED);
-        tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE_DETAILED);
+        tooltip.remove(JadeIds.UNIVERSAL_ENERGY_STORAGE);
+        tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE);
 
         var context = ContextHelper.getContext(accessor);
         var tooltipBuilder = new JadeTooltipBuilder(tooltip);


Mari023:
the idea behind the fluid storage removal is to show (partially) filled tanks, but not empty ones, since that would clutter stuff a lot (the interface has 9 tanks otherwise).
which from your picture it seems like it works as intended?

For the FE storage, it only really makes sense for standalone devices that aren't connected to the network (e.g. inscriber, charger), but that sounds a lot more complicated than just not showing the FE tooltip.

Frostbite-time:
> the idea behind the fluid storage removal is to show (partially) filled tanks, but not empty ones, since that would clutter stuff a lot (the interface has 9 tanks otherwise). which from your picture it seems like it works as intended?
> 
> For the FE storage, it only really makes sense for standalone devices that aren't connected to the network (e.g. inscriber, charger), but that sounds a lot more complicated than just not showing the FE tooltip.

For fluid storage tooltip, I’m very certain that the original `tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE_DETAILED)` was ineffective—it doesn’t actually do anything. Even if you remove that line entirely, the ME Interface will still behave exactly the same as shown in my “before” screenshot, which I have verified. So if you want to preserve the behavior you described, only need to remove the fluid-related line before merging.

Mithi83:
Regarding the tooltips for fluids I'd think empty tanks should not appear in the list, but you "after" variant does not show any fluids, not even the partially filled water tanks, assuming the interface is in the same internal state as in the "before" variant. To assess if the first screenshot is actually correct or not I'd have to see the interface GUI. If two slots in the interface actually have one bucket of water each then the "before" screenshot would be correct in my opinion. If however only one slot in the interface has one bucket of water and all others are empty the "before" would indicate that it is wrong.

Frostbite-time:
> Regarding the tooltips for fluids I'd think empty tanks should not appear in the list, but you "after" variant does not show any fluids, not even the partially filled water tanks, assuming the interface is in the same internal state as in the "before" variant. To assess if the first screenshot is actually correct or not I'd have to see the interface GUI. If two slots in the interface actually have one bucket of water each then the "before" screenshot would be correct in my opinion. If however only one slot in the interface has one bucket of water and all others are empty the "before" would indicate that it is wrong.

Uh, I think I need to explain the fluid tooltip situation in more detail:
	1.	The interface indeed has exactly two slots containing 1 bucket of water, and this is true in both the “before” and “after” screenshots. So, as you said, the “before” state is the expected one.
	2.	I removed the fluid tooltip in the “after” because I noticed the original code contains a line tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE_DETAILED), and that line is wrong—it doesn’t actually do anything. Based on that line alone, I assumed the original intention might have been to remove the fluid tooltip, so I replaced it with tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE). This change is what makes the fluid tooltip disappear in the “after” screenshot.
	3.	Even if the “before” state is correct, the original tooltip.remove(JadeIds.UNIVERSAL_FLUID_STORAGE_DETAILED) is still incorrect because it does nothing. Keeping that line or deleting it results in exactly the same fluid tooltip behavior: the interface has two slots each containing one bucket of water, the tooltip shows exactly two buckets of water, and empty slots are not shown.
	4.	If you want to keep the fluid tooltip behavior as in the “before” screenshot, you only need to delete the fluid-related line I changed (and not replace it with the original). I can push a commit tomorrow to do that.

If you have any further questions, I’ll answer after I wake up tomorrow.

Frostbite-time:
Now the fluid tooltip state matches what it was before. Here are screenshots after this change:

before:
f25d1fafd54c3afde7e7b8e5f2b3f4aa
f83faf4ba50c080a2e1c63f8a00f1222

after:
ce6080e0e02c312f03776d9c7cef9f0e
927d38856b59ed77467cad4bfcb4111c

state inside the ME Interface:
27e5fab8bf5d1d9f3ec83736c191a73d

PR #8808, opened 2026-03-11
feat: Add manual pinning to storage screen

llpaull:


src/main/java/appeng/client/gui/me/common/Repo.java
```diff
@@ -210,13 +218,18 @@ private void addEntriesToView(Collection<GridInventoryEntry> entries) {
...
-            if (hasPinnedRow && pinnedRow.size() < rowSize && PinnedKeys.isPinned(entry.getWhat())) {
-                pinnedRow.add(entry);
-                continue;
+            if (PinnedKeys.isPinned(entry.getWhat())) {
+                PinnedKeys.PinInfo info = PinnedKeys.getPinInfo(entry.getWhat());
+
+                if (info.reason == PinnedKeys.PinReason.MANUAL && pinnedRowManual.size() < rowSize) {
+                    pinnedRowManual.add(entry);
+                    continue;
+                } else if (pinnedRowOther.size() < rowSize) {
```
Mithi83, line 229: I think you have to explicitly check for `info.reason == PinnedKeys.PinReason.CRAFTING` here as well. Otherwise you end up in the `else if` with `PinReason.MANUAL` if the `if (info.reason == PinnedKeys.PinReason.MANUAL && pinnedRowManual.size() < rowSize)` in line 226 failed due to the `rowSize` check.

Issue #8450, opened 2025-04-02
Crafting CPUs frequently lock up completely and need manually resetting


Vectrobe:
### Describe the bug

When performing relatively simple craft trees but with many ingredients, say crafting a new stack (64) of blank patterns, the crafting CPU that gets chosen has a high likelyhood of completely locking up, including the timer, and with an arbitrary count of ingredients in the CPU store

### How to reproduce the bug

Unclear, but in my grid that I have attached I currently use 16 standard pattern providers, 3 CPU towers of which are x1 16k store and x3 co-processor, and then just the usual all-in-one inscriber tower and the common mekanism machines for smelting, crushing, etc.

One catch with some of the mekanism machines, crusher and smelter, they have pattern providers on two faces to fit more patterns, so in the machine's sides config the top is set to input-output with auto eject, and the sides are just input, patterns on the sides do complete crafting jobs perfectly fine so I dont expect this specifically to be an issue, nor do items go missing, items get left in the main storage, or the CPU complain of missing items.

A possibility that does come to mind is maybe the CPU gets stuck in a self-dependency-loop because quartz glass has the glass tag that is used to make itself, this could explain why the timer also locks up

![Image](https://github.com/user-attachments/assets/829c8a9e-37ca-4fa4-b5ec-cefc8c4b2e46)

[ae2_grid_269_20250402_2215.zip](https://github.com/user-attachments/files/19567062/ae2_grid_269_20250402_2215.zip)

### Expected behavior

-

### Additional details

[crash-2025-04-02_22.33.56-client.txt](https://github.com/user-attachments/files/19567166/crash-2025-04-02_22.33.56-client.txt)

### Which minecraft version are you using?

1.21

### On which mod loaders does it happen?

NeoForge

### Crash log

https://github.com/user-attachments/files/19567166/crash-2025-04-02_22.33.56-client.txt

Vectrobe:
just to further note; performing the manual crash (while still within the save) and then restarting and jumping back to the save; the crafting CPU is still stuck exactly the same, the only difference is the timer has reset to 00:00:00.

timer wise as well, it has undefined behavior, sometimes it does actually tick the seconds up but only when the UI isnt open, and extremely rarely at that, seems as if it only does so while the game is paused, rather counterintuitively

Vectrobe:
[...]

Issue #1944, opened 2015-10-11
Crafting pylons seems to loose inventory on server reboot. 


melekaiah:
Has anyone experienced this. 
I got to autocraft a whole set of disks, with all the parts, having to be made, a task that can take well over 20 minutes depend on power availability etc. 

it gets closed to finsihed and the server reboots for its daily reboot, 
when you log back in, half the items are gone from the pylon, its progress has stopped, and just sits there. 

this has happend twice to me now, loosing tons of stuff in the process, (not a big deal since im admin)
just wanted to know if its something that you guys can reproduce in the build 8 of the latest round of alphas. or if my mod pack is to blame.

mindforger:
which exact version are you using? i noticed this behaviour in rv2_8 or rv2_9 

but the key factor was using chickenbone's chunkloaders! They break something when force loading the chunk, so we kicked chickenchunks from our server as those chunkloaders are way too OP anyways :P

can you can add some more infos on what modpack you are using and in which way you do chunkloading? is the whole energy setup chunkloaded? and what the heck is a crafting-"pylon"?

melekaiah:
rv-3 alpha 8 is what im using at the moment

and yes i am indeed using chickenchunks, we have a chunk area with 10 chunk radius loaded as our city core, in which we have a central ME-system with a crafting setup. 

below is a picture of my crafting core, the craftying "pylon" is my name for the crafting unit, 2 coprocessors, 1 64k storage 2x4k storage, and 1 crafting dispaly, 
![2015-10-12_13 10 11](https://cloud.githubusercontent.com/assets/2993253/10426669/9dbc133a-70e2-11e5-9343-2b558803b94c.png)

The mod Pack is one ive been building for a while, thats still in testing, called abandoned worlds. as soon as i can get all the obvious bugs fixed im gonna start asking for permissions for all the mods in the pack so i can make it semi public at least. 

i prefer not to remove mods since that seems to screw up the id's left right and center, (is that still the case? i know it was a while ago)

XFactHD:
@mindforger I think he means the crafting CPU. The problem here is that at any time the player leaves the chunk without placing a chunkloader or reloads the world, the chunks unload meaning that the multiblock is invalidated and spills the stored items into the world.

melekaiah:
i havent acutally noticed any SPILLING of items, since ive been online within a minute of server reboot, 
[...]

Issue #8391, opened 2025-02-22
Items get stuck in molecular assembler if server reboots when crafting


Krutontar:
### Describe the bug

If you are doing a long craft and the server crashes or even just does a scheduled reboot, when you get back in, the job is gone and the items that were in the molecular assemblers at the moment of shutdown will be in the assemblers and can block future crafts.

### How to reproduce the bug

Start a big crafting job with a lot of "crafting table" tasks and then reboot it before the job finishes.  Afterwards, go inspect the molecular assemblers attached to the pattern provider that has the recipes for the large crafting job.

### Expected behavior

For them to be empty on a server boot?  I think I'd rather lose what was in them rather than have to go track down which assembler was stuck especially when you have a LOT of crafting automation setup up.  Maybe a process that runs on sever boot that pulls anything back into the network but that could get complicated if there is no space for more items.  I respect that this could be a hard design choice.

### Additional details

I am playing ATM10.
I did the F3+C as instructed below but the problem is not a crash.
I put 1.21 because that's what I'm playing now but I have seen this happen for several past versions.  I don't remember how long ago.

### Which minecraft version are you using?

1.21

### On which mod loaders does it happen?

NeoForge

### Crash log

https://gnomebot.dev/paste/mclogs/q2IhSTh

KrisanaponMeksamoot:
I see [theShroo](https://github.com/theShroo) working on [PR#8819 - Fix crafting CPU inventory loss on chunk unload / server shutdown](https://github.com/AppliedEnergistics/Applied-Energistics-2/pull/8819) with similar idea

PR #8819, opened 2026-04-06
Fix crafting CPU inventory loss on chunk unload / server shutdown

theShroo:
## Summary

Fixes in-progress crafting CPU items being permanently voided when the server shuts down or the chunk containing the crafting CPU is unloaded.

**Root cause**: During chunk unload, `onChunkUnloaded()` destroys the grid node, which triggers `disconnect()` → `cluster.destroy()` → `updateStatus(null)`. This nulls the `cluster` reference on `CraftingBlockEntity` **before** `saveAdditional()` is called by the chunk serializer. With `cluster == null`, `saveAdditional()` skips writing the crafting inventory and job state entirely — all items extracted from the ME network for the in-progress job are permanently lost.

**The fix** (two changes in `CraftingBlockEntity.java`):

1. **`updateStatus(null)`**: Before releasing the cluster reference, snapshot the cluster's state (inventory + job) into `previousState` via the existing `setPreviousState()` mechanism.

2. **`saveAdditional()`**: When `cluster` is null but `previousState` is non-null, merge the snapshot into the save data. On reload, the existing `loadTag()` → `setPreviousState(data.copy())` path restores it, and when the cluster reforms the job resumes with all items intact.

This leverages AE2's existing `previousState` field, which was already designed for exactly this purpose (holding cluster data when the cluster isn't formed yet).

## Related issues

- #1944 — "Crafting pylons seems to loose inventory on server reboot" (closed without fix, 2020)
- #8450 — "Crafting CPUs frequently lock up completely and need manually resetting" (open, related symptom)

## Testing

1. Start a large crafting job (e.g. 64x of a multi-step recipe)
2. While the job is in progress, `/stop` the server
3. Restart the server
4. **Before fix**: crafting CPU is empty, job gone, items voided
5. **After fix**: crafting CPU retains its inventory and the job resumes

Also tested with chunk unload (walking far away from the crafting CPU and returning).

Tested on NeoForge 21.1.219 / MC 1.21.1 with a 200+ mod modpack (FTB Evolution).

### DIFF
--- a/src/main/java/appeng/blockentity/crafting/CraftingBlockEntity.java
+++ b/src/main/java/appeng/blockentity/crafting/CraftingBlockEntity.java
@@ -118,6 +118,17 @@ public class CraftingBlockEntity extends AENetworkedBlockEntity
 
     public void updateStatus(CraftingCPUCluster c) {
         if (this.cluster != null && this.cluster != c) {
+            // Persist the cluster's crafting state before releasing the reference.
+            // During chunk unload, onChunkUnloaded() → destroy() → updateStatus(null)
+            // runs BEFORE the chunk is saved to disk. If we null the cluster here,
+            // saveAdditional() will skip writing the inventory and in-progress crafting
+            // jobs are permanently lost. Saving to previousState ensures the data
+            // survives for the subsequent saveAdditional() call.
+            if (c == null && this.isCoreBlock()) {
+                var data = new net.minecraft.nbt.CompoundTag();
+                this.cluster.writeToNBT(data, this.level.registryAccess());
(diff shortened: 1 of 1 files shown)


KrisanaponMeksamoot:
Is this related to #8391
please take note that [appeng.blockentity.crafting.MolecularAssemblerBlockEntity.progress](https://github.com/theShroo/Applied-Energistics-2/blob/13fa5e39f1d13086a7cc9baf57ecd5b4dda45b50/src/main/java/appeng/blockentity/crafting/MolecularAssemblerBlockEntity.java) can also be saved using the same logic

PR #8827, opened 2026-04-18
[26.1] fix GUI - needs to be double-checked

Mithi83:


Mari023:
seems to work from my testing

PR #8828, opened 2026-04-19
[26.1] fix padding of scrollbars and other GUI textures on atlases

Mari023:
fix scrollbar, fluid rendering in GUI and the crafting marker (when pinned) being all offset by the one pixel of padding

based on / contains the changes from #8827 (without, they aren't visible at all)


Mari023:
Long term it might make more sense to use the uv provided by the sprite, but right now `Blitter` doesn't have a way to deal with that.

Mithi83:
I could imagine a rather blunt approach along the lines of:

- use accesstransformer.cfg to make `padding` in `TextureAtlasSprite` public
- use that padding in the constructor `Blitter sprite(TextureAtlasSprite sprite)`

shartte:
Yeah I made padding visible and used that inside of Blitter (and added a new overload to get a sub-region of the sprite).
7fc06db03

PR #8834, opened 2026-04-23
Fix: Low-localizable text of crafting status tooltip (#8833)

### LINKED ISSUE #8833: Low-localizable text of crafting status tooltip
The output item name should be added based on localization text, not on code.

Image

https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/d6efcbd569e7d368a12bd806a0bdeb7663677fe7/src/main/java/appeng/core/localization/ButtonToolTips.java#L42
https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/d6efcbd569e7d368a12bd806a0bdeb7663677fe7/src/main/java/appeng/client/gui/widgets/CPUSelectionList.java#L113-L114

ookkoouu:
close #8833


andrewsf:
Given the mechanical nature of this change, this seems like a case wherein all the locale-specific JSONs should be updated from `%s` to `%s %s` to keep backward compatibility with the existing strings.

PR #8840, opened 2026-04-29
reactivate pattern access terminal searching

Mithi83:


Mithi83:
The code compiles and seems to work. I don't know why it was commented out in the first place but it should be good to go once again.

shartte:
> The code compiles and seems to work. I don't know why it was commented out in the first place but it should be good to go once again.

Pretty sure because of no recipe availability on the client-side in 1.21.4

PR #8848, opened 2026-05-05
[26.1] Fix NeoForge API breaking changes

MPThLee:
In https://github.com/neoforged/NeoForge/pull/3064 (which is NF's `26.1.2.21-beta`)
> ▸ This PR moves BlockEvent.BreakEvent to a standalone top-level class BreakBlockEvent, and updates the firing logic to fire BreakBlockEvent on the client, instead of only firing it on the server. It still preserves the ability to send a packet when cancelled only on the server, but this is an opt-in behavior now.

There was breaking change in neoforge, so I updated neoforge version in gradle.properties and fix the event import. (which is only one)

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -9,8 +9,8 @@ java_version=25
 #########################################################
 minecraft_version=26.1.2
 # https://projects.neoforged.net/neoforged/neoforge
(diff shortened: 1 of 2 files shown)


shartte:
Thank you!

PR #8850, opened 2026-05-05
Include client source set in sources and javadoc jars

Lapis256:
Fixed missing contents from the client source set in the sources and javadoc jars.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -260,7 +260,8 @@ static void publicApiIncludePatterns(PatternFilterable spec) {
 
 tasks.named('javadoc', Javadoc) {
     source = sourceSets.main.allJava
-    classpath = sourceSets.main.compileClasspath + sourceSets.main.output
+    source sourceSets.client.allJava
+    classpath = sourceSets.client.compileClasspath + sourceSets.client.output
 
     options.addStringOption('Xdoclint:none', '-quiet')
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -259,20 +259,30 @@ static void publicApiIncludePatterns(PatternFilterable spec) {
...
 
 tasks.named('javadoc', Javadoc) {
-    source = sourceSets.main.allJava
-    classpath = sourceSets.main.compileClasspath + sourceSets.main.output
+    source sourceSets.main.allJava
+    source sourceSets.client.allJava
+    classpath = (sourceSets.main.compileClasspath
+            + sourceSets.main.output
+            + sourceSets.client.compileClasspath
+            + sourceSets.client.output)
```
shartte: Shouldn't sourceSets.client.compileClasspath already include main compileClasspath+output? 🤔

Lapis256: Thanks, I had misunderstood this. I removed the redundant `sourceSets.main.compileClasspath + sourceSets.main.output`.

Lapis256:
Also cleaned up the Javadoc source configuration to keep the existing `main` assignment, and switched the sources jar addition to `sourceSets.client.allSource` so client resources are included too.

PR #8854, opened 2026-05-08
Fix Curios integration breaking when Curios is not present

Mari023:
The `Open Wireless Terminal` keybinding breaks when curios is not present, due to the curios api 

Revert "remove duplicate curios capability implementation"
This reverts commit d47025097f9e3a84814eec8c27389ce7954e69d3, part of #8838 
I also added a comment explaining why it was done this way.

alternatively checking if curios is present before calling the api would work too, I implemented that in f01ec3a723a40b99347843223b1255bed4871089

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -110,7 +110,6 @@ dependencies {
             break
     }
 
-    compileOnly "top.theillusivec4.curios:curios-neoforge:${project.curios_version}:api"
     if (project.runtime_curio == "true") {
         localRuntimeOnly("top.theillusivec4.curios:curios-neoforge:${project.curios_version}")
     }
--- a/src/main/java/appeng/hotkeys/CuriosHotkeyAction.java
+++ b/src/main/java/appeng/hotkeys/CuriosHotkeyAction.java
@@ -6,9 +6,8 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.ItemLike;
 
-import top.theillusivec4.curios.api.CuriosCapability;
-
 import appeng.api.features.HotkeyAction;
+import appeng.integration.modules.curios.CuriosIntegration;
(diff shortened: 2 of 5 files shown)


Mithi83:
I actually like the alternative better. It preserves the removal of the code duplication. Registering events only when needed sounds like a generally advisable strategy.

However I am not sure the alternative is complete. You make two registrations conditional, but there are three files that would need fixing. What about CuriosItemLocator.java? Or rather its usage in MenuLocators.java?

Thanks for analyzing the mess I produced.

Mari023:
> However I am not sure the alternative is complete.

I think the CuriosItemLocator is only ever created from the CuriosHotkeyAction, so it should only get created when curios is present.
that would leave addons, but this isn't API anyway so imo it's on the addons to check.

on the topic of addons:
it is convenient for ae2wtlib to have the `CuriosIntegration`, tho I guess I can just do that in ae2wtlib itself.

There is also the 3rd option of keeping that class, but having it call the curios API (if it is installed), tho at that point using the curios api is basically a compile time check whether the type is correct (which is a nice to have, tho it kind of falls apart on having to manually update the curios version anyway)

The main reason I originally implemented curios support this way was basically "we have a cool new api in neoforge that allows us to use other mods Capabilities without caring whether the mod exists at all"

Mithi83:
I just suggested that change because I ran into exactly what you describe during the 26.1 port: The signature of the capability changed and we had to manually find and also implement that change.

Funny side note: I'm quite sure that this bug here (#8855) has its root in the changed capability.

Mari023:
> Funny side note: I'm quite sure that this bug here (https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/8855) has its root in the changed capability.

ah, of course, previously we could modify the underlying stack directly, now we just get a copy where we need to set the item on modification.

shartte (approved):

PR #8864, opened 2026-05-11
fix keybinding category

Mithi83:


Mithi83:
image

Fixed in AE2 vs not fixed in GuideME.
I'm not sure what to do about the translations via Crowdin though.

shartte (approved):

PR #8873, opened 2026-05-21
Automatic pattern restock

Satherov:
Automatically tries to restock a blank pattern from the system once the pattern terminal runs out. Will also automatically pull a pattern when a new recipe is set while no patterns are available.

https://github.com/user-attachments/assets/d5c2773b-2f86-418d-b125-4f59555a4bb0


HowXu:
As I known, This is a feat of ExtendedAE

PR #8885, opened 2026-06-04
correct world position and block flags in attemptRecovery

quasar8842:
Plane-local x/y/z were used for BlockPos without x/y/z offsets, unlike addBlockEntity. sendBlockUpdated also passed z as update flags instead of 3. Only affects recovery when completeMove fails during spatial transfer.

### DIFF
--- a/src/main/java/appeng/spatial/CachedPlane.java
+++ b/src/main/java/appeng/spatial/CachedPlane.java
@@ -351,7 +351,7 @@ public class CachedPlane {
     }
 
     private void attemptRecovery(int x, int y, int z, BlockEntityMoveRecord moveRecord, Column c) {
-        var pos = new BlockPos(x, y, z);
+        var pos = new BlockPos(x + this.x_offset, y + this.y_offset, z + this.z_offset);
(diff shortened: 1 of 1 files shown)


shartte:
Have you found a reproducible way of forcing recovery?

quasar8842:
> Have you found a reproducible way of forcing recovery?

No. I tried spatial transfers with block entities from other mods (including spatial-blacklisted blocks), but I couldn’t get `completeMove` to fail, so I never saw `attemptRecovery` run in-game.

The coordinate mismatch is still clear when compared with `addBlockEntity`, which applies the plane offsets:

```java
var newPosition = new BlockPos(x + this.x_offset, y + this.y_offset, z + this.z_offset);
```

`attemptRecovery` uses the same `x`/`y`/`z` arguments without offsets:

```java
var pos = new BlockPos(x, y, z);
```

There’s a second issue in the same method: `sendBlockUpdated` passes `z` as the update-flags argument instead of `3` (matching the `setBlock` call above it).

I’m happy to add a small test or a dev-only hook to force the recovery path if that would help review.

PR #8902, opened 2026-06-17
Fix fluid rendering in recipes in JEI

Mithi83:


Mithi83:
Inspired by https://github.com/AppliedEnergistics/GuideME/blob/main/src/main/java/guideme/scene/level/GuidebookLevel.java

Needs a proper review though.

Before:
image
After:
image

PR #8944, opened 2026-08-01
Fix curio integration

### LINKED ISSUE #8855: AE2 26.1.7 alpha  Wireless Crafting Terminal voids items and doesn't work.
### Describe the bug

ATM11-AE2 26.1.7 alpha
The wireless crafting terminal does not craft. The whole 3x3 crafting grid doesn't work. Not through jei or manually. Anything added to the grid will be voided.

### How to reproduce the bug

Open wireless crafting terminal, try putting any items in the crafting grid.

### Expected behavior

Should craft items and not void things.

### Additional details

_No response_

1.21

NeoForge

### Crash log

There was no crash

Mari023:
fix #8855
based on the wip-upgrade-rework branch (rebased on main)

this has some soundness holes, specifically it will probably break if modifying the content of the ItemAccess is not successful.
I believe properly fixing that would require making InternalInventory transaction aware.

I didn't actually test with curio, since I couldn't get the slots to show up,
I did however test that it works if the given ItemStack is a copy (and thus effectively read only).

A lot of write operations on ItemStacks now open a transaction in order to modify the ItemStack.
This would cause problems if it were to happen during another transaction.

This still needs proper testing (both with and without curios), and probably some polish (the names of some new stuff aren't great), (and according to the ci, some javadoc).
I will however be unavailable until September starting in a few days, so I won't be able to do that myself.


Mithi83:
Not getting curio slots is strange. The original wip-upgrade-rework branch added a temporary (?) JSON file (`src/main/resources/data/ae2/curios/player.json`) that enabled curio slots. Did you enable the curio integration in your runtime environment `gradle.properties` -> `runtime_curio=false`?

Mari023:
I set that to true, curios is loading, and I can open the curios menu, but there isn't anything there
I looked in the curio wiki, and it seems like that json file is supposed to be in `data/ae2/curios/entities/entities.json`
placing it there makes the slots show up.

a quick test confirms that wireless terminals now work properly in curio slots,
and that the main inventory of portable storage cells is still broken. oops.

Mari023:
I fixed the portable cells, everything should be working now

Mithi83:
I've not done a lot with your branch, but it seems to work now (with the renaming of the JSON file and the enabling of curios in `gradle.properties` as two manual steps). Superficially tested, Wireless Crafting Terminals and Portable Item Cells are usable again in curio slots. 

I'm kind of confused because of the JSON file though. I'm 100% positive that what Gimpansor committed initially was the location I told him I used to reproduce the problem in the first place documented in here: https://github.com/AppliedEnergistics/Applied-Energistics-2/issues/8855#issuecomment-4406093149

Just for clarification: This PR does not include making `InternalInventory` transaction aware and thus does not help with the other bugs related to that problem, right?

Thanks for your time and fixing what I broke during the 26.1 port.

Mari023:
> Just for clarification: This PR does not include making InternalInventory transaction aware and thus does not help with the other bugs related to that problem, right?

yes.
I decided against doing that for now, since that would have been a lot more work, and I could get the "replace item on save" method to work.

I still think longterm making it transaction aware is the better way (as it would also fix the soundness hole), but for now I think just getting curios to work is enough

PR #8952, opened 2026-08-14
(#8951) Fix AEKey.hasComponents() returning the inverse in AEItemKey and AEFluidKey

### LINKED ISSUE #8951: AEItemKey.hasComponents returns the inverse and can poison crafting substitution cache
### Describe the bug

`AEKey.hasComponents()` is documented to return `true` when the key has any components attached, but both `AEItemKey` and `AEFluidKey` currently implement it as:

```java
@Override
public boolean hasComponents() {
    return stack.getComponents().isEmpty();
}
```

This returns the inverse of the documented meaning.

For `AEItemKey`, this also affects crafting substitutions because `AECraftingPattern` uses `hasComponents()` to decide whether an `isItemValid` result can be cached by the base `Item`:

```java
private Boolean getTestResult(int slot, AEItemKey what) {
    if (what == null || what.hasComponents()) {
        return null;
    }
    var cache = isValidCache[slot];
    return cache == null ? null : cache.get(what.getItem());
}

private void setTestResult(int slot, AEItemKey what, boolean result) {
    if (what != null && !what.hasComponents()) {
        // cache by Item
    }
}
```

A component-bearing key can therefore be treated as component-less and cached by its base item. If one component variant fails validation first, the cached `false` can then reject another valid component variant of the same item.

The same inverted implementation is present in `AEFluidKey`.

Observed on AE2 19.2.17 for Minecraft 1.21.1 with NeoForge. The same implementation is also present on the current `main` branch.

### How to reproduce the bug

The direct semantic mismatch can be reproduced with a component-bearing stack:

```java
var stack = new ItemStack(Items.PAPER);
stack.set(DataComponents.CUSTOM_NAME, Component.literal("variant"));

var key = AEItemKey.of(stack);

assertTrue(key.hasComponents()); // currently fails
```

The crafting-cache consequence can be reproduced with two stacks that have the same base item but different data-component patches:

1. Create a crafting pattern with substitutions enabled.
2. Make one component variant invalid for a pattern slot and another component variant valid.
3. Validate the invalid variant first.
4. `AECraftingPattern.isValidCache[slot]` stores `false` under the base `Item`.
5. Validate or assemble with the valid variant.
6. The valid variant is rejected by the cached value instead of being checked independently.

The order of fuzzy template enumeration can make this appear intermittent.

### Expected behavior

`hasComponents()` should match its documented contract. Component-bearing item/fluid keys must return `true`, and component variants must not share `AECraftingPattern`'s item-level validation cache.

For item stacks, the implementation may need to check the attached component patch rather than the complete component map, since the latter can contain default item components.

### Additional details

Relevant current sources:

- [AEItemKey.hasComponents()](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/main/src/main/java/appeng/api/stacks/AEItemKey.java)
- [AEFluidKey.hasComponents()](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/main/src/main/java/appeng/api/stacks/AEFluidKey.java)
- [AECraftingPattern component-sensitive validity cache](https://github.com/AppliedEnergistics/Applied-Energistics-2/blob/main/src/main/java/appeng/crafting/pattern/AECraftingPattern.java)

Minecraft version: 1.21.1  
Mod loader: NeoForge  
Crash log: Not applicable; this does not crash.

MakerYuichi:
Fixes #8951

`AEItemKey.hasComponents()` and `AEFluidKey.hasComponents()` were both implemented as `stack.getComponents().isEmpty()`, which returns `true` when the stack has no components and `false` when it does — the inverse of the documented contract.

For `AEItemKey`, this let component-bearing item keys be treated as component-less by `AECraftingPattern`'s `isValidCache`, allowing one component variant's cached validation result to incorrectly apply to a different component variant of the same item.

## Fix

Both now check the attached component **patch** rather than the fully-resolved component map:
- `AEItemKey.hasComponents()` → `!stack.isComponentsPatchEmpty()`
- `AEFluidKey.hasComponents()` → `!stack.getComponentsPatch().isEmpty()`

This matches the pattern already used elsewhere in each class (`toString()`, the `MAP_CODEC`). The fully-resolved component map includes an item's or fluid's default components and is therefore rarely empty, so checking it
directly (as the original code did, just inverted) would have made `hasComponents()` return `true` for almost every stack — the exact trap flagged in the issue.

Also fixes the same inverted check in `AEFluidKey.toString()`, found while fixing the above.

`AECraftingPattern`'s cache gating (`getTestResult`/`setTestResult`) required no change, it already skips the shared cache whenever `hasComponents()` is `true`, which now correctly triggers for component-bearing keys.

## Testing

Added a regression test on `AEItemKey` covering an unmodified item, an item with an explicit component set (`CUSTOM_NAME`), and the same item after that component is removed. Full `AEItemKeyTest` suite passes locally (31/31, 0 failures).

### DIFF
--- a/src/main/java/appeng/api/stacks/AEItemKey.java
+++ b/src/main/java/appeng/api/stacks/AEItemKey.java
@@ -243,7 +243,7 @@ public final class AEItemKey extends AEKey {
 
     @Override
     public boolean hasComponents() {
-        return stack.getComponents().isEmpty();
+        return !stack.isComponentsPatchEmpty();
     }
 
     /**
--- a/src/main/java/appeng/api/stacks/AEFluidKey.java
+++ b/src/main/java/appeng/api/stacks/AEFluidKey.java
@@ -166,7 +166,7 @@ public final class AEFluidKey extends AEKey {
 
     @Override
     public boolean hasComponents() {
-        return stack.getComponents().isEmpty();
+        return !stack.getComponentsPatch().isEmpty();
     }
 
     public FluidResource toResource() {
@@ -200,6 +200,6 @@ public final class AEFluidKey extends AEKey {
         var id = BuiltInRegistries.FLUID.getKey(getFluid());
         String idString = id != BuiltInRegistries.FLUID.getDefaultKey() ? id.toString()
                 : getFluid().getClass().getName() + "(unregistered)";
-        return stack.getComponents().isEmpty() ? idString : idString + " (+components)";
+        return stack.getComponentsPatch().isEmpty() ? idString : idString + " (+components)";
     }
 }
--- a/src/main/java/appeng/api/stacks/AEFluidKey.java
+++ b/src/main/java/appeng/api/stacks/AEFluidKey.java
@@ -166,7 +166,7 @@ public final class AEFluidKey extends AEKey {
 
     @Override
     public boolean hasComponents() {
-        return !stack.getComponentsPatch().isEmpty();
+        return !stack.isComponentsPatchEmpty();
     }
 
     public FluidResource toResource() {
--- a/src/test/java/appeng/api/stacks/AEItemKeyTest.java
+++ b/src/test/java/appeng/api/stacks/AEItemKeyTest.java
@@ -17,7 +17,9 @@ import org.junit.jupiter.params.provider.CsvSource;
 import org.junit.jupiter.params.provider.EnumSource;
 
 import net.minecraft.core.RegistryAccess;
+import net.minecraft.core.component.DataComponents;
 import net.minecraft.nbt.CompoundTag;
+import net.minecraft.network.chat.Component;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.util.ProblemReporter;
 import net.minecraft.world.item.ItemStack;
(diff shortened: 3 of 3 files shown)


src/main/java/appeng/api/stacks/AEItemKey.java
```diff
@@ -243,7 +243,7 @@ public boolean isTagged(TagKey<?> tag) {
 
     @Override
     public boolean hasComponents() {
-        return stack.getComponents().isEmpty();
+        return !stack.isComponentsPatchEmpty();
```
shartte, line 246: Is this ItemStack specific? As in: We (NeoForge) didn't copy that over to FluidStack as well?

MakerYuichi: Turns out isComponentsPatchEmpty() does exist on FluidStack too. So AEFluidKey.hasComponents() could use !stack.isComponentsPatchEmpty() instead of .getComponentsPatch().isEmpty(), which'd match AEItemKey more closely.
Same result either way though, so just a nice-to-have, not a blocker.

MakerYuichi:
Just following up on this PR. 

I decided to keep the code exactly as is since both fluid-checking approaches yield the identical functional output nd all local regression tests are passing (31/31). 

The build environment checks look completely healthy. Whenever you have a moment, could you give this a final review and approval so we can get this bugfix merged? Thanks!

PR #8962, opened 2026-09-01
Fix JEI integration for merequester

Mithi83:
Adapt the fix for this problem for 26.1: https://github.com/Tamaized/AE2-JEI-Integration/issues/3


rlnt:
Some more context: This is not only a fix for ME Requester. This problem was discovered via the add-on mod, though.

The problem addressed in this PR is that the JEI integration's `GhostIngredientHandler` checks the `FakeSlot#canSetFilterTo` method to ensure the stack is valid for the slot, but does not use the slot's logic for actually placing the stack. Instead, it sends the `InventoryActionPacket`, which duplicates the logic inside `FakeSlot#setFilterTo`.

While this works for base AE2 slots, this breaks for add-ons defining custom behavior in their own `FakeSlot`s. With this PR implemented, each `FakeSlot` can override the filter setting logic on its own, which should be the preferred way. At the same time, this also removes the duplicate logic for sending the packet.

PR #8970, opened 2026-09-10
Fix internal inventory resource handlers

shartte:


src/main/java/appeng/api/inventories/SubInventoryProxy.java
```diff
@@ -91,4 +94,9 @@ public ItemStack extractItem(int slot, int amount, boolean simulate) {
     public void sendChangeNotification(int slot) {
         delegate.sendChangeNotification(translateSlot(slot));
     }
+
+    @Override
+    protected ResourceHandler<ItemResource> createResourceHandler() {
+        return RangedResourceHandler.of(toResourceHandler(), fromSlot, toSlot);
```
andrewsf: Infinite recursion, this calls the superclass's `toResourceHandler()` which then calls this class's `createResourceHandler()`

src/main/java/appeng/util/inv/AppEngInternalInventory.java
```diff
@@ -200,4 +210,179 @@ protected final void setHost(InternalInventoryHost host) {
...
+                this.counts = new int[size()];
+            }
+        }
+
+        @Override
+        public void onRootCommit(Snapshot original) {
+            for (int i = 0; i < original.items.length; i++) {
+                var current = stacks.get(i);
+                if (current != original.items[i] || current.getCount() != original.counts[i]) {
+                    sendChangeNotification(i);
```
andrewsf: FYI this is a no-op (`default` impl in `InternalInventory` interface)

although there used to be an override in this class until 1.20.2 port
https://github.com/AppliedEnergistics/Applied-Energistics-2/commit/6e0962ada#diff-22ea0e45356a063d99c9a2475dbb0cce84e80a613e373d4e2de0eef3747f8017

src/main/java/appeng/util/inv/FilteredInternalInventory.java
```diff
@@ -85,4 +90,71 @@ public boolean isItemValid(int slot, ItemStack stack) {
...
+            if (!filter.allowInsert(FilteredInternalInventory.this.delegate, index, resource.toStack())) {
+                return 0;
+            }
+            return super.insert(index, resource, amount, transaction);
+        }
+
+        @Override
+        public int insert(ItemResource resource, int amount, TransactionContext transaction) {
+            // This duplicates the default implementation from ResourceHandler, which is inaccessible here
+            // We need to check the filter for each index we access, which is impossible if we call the delegates
```
Mithi83: That should be "deletage's" I think.

src/main/java/appeng/util/inv/FilteredInternalInventory.java
```diff
@@ -85,4 +90,71 @@ public boolean isItemValid(int slot, ItemStack stack) {
...
+            if (!filter.allowExtract(FilteredInternalInventory.this.delegate, index, amount)) {
+                return 0;
+            }
+            return super.extract(index, resource, amount, transaction);
+        }
+
+        @Override
+        public int extract(ItemResource resource, int amount, TransactionContext transaction) {
+            // This duplicates the default implementation from ResourceHandler, which is inaccessible here
+            // We need to check the filter for each index we access, which is impossible if we call the delegates
```
Mithi83, line 135: That should be "deletage's" I think.

src/main/java/appeng/util/inv/AppEngInternalInventory.java
```diff
@@ -200,4 +210,179 @@ protected final void setHost(InternalInventoryHost host) {
...
+            // Do not allow extraction of wrapped fluid stacks because they're an internal detail
+            if (resource.getItem() == AEItems.WRAPPED_GENERIC_STACK.asItem()) {
+                return 0;
+            }
+
+            updateSnapshots(transaction);
+
+            inTransactionalCode = true;
+            try {
+                return extractItem(index, maxAmount, false).getCount();
```
andrewsf, line 299: this slot may not actually hold the requested resource, need to check `resource` against `stacks.get(index)` before calling?

src/main/java/appeng/util/inv/AppEngInternalInventory.java
Mithi83: You don't implement `sendChangeNotification`. I've checked with a breakpoint here that this goes straight to the empty default implementation from `InternalInventory`.

Maybe something like this:

```
    @Override
    public void sendChangeNotification(int slot) {
        onContentsChanged(slot);
    }
```

I'm surprised that it does not manifest in a bug ingame at first glance though, so maybe I'm mistaken.

src/main/java/appeng/util/inv/AppEngInternalInventory.java
```diff
@@ -200,4 +210,179 @@ protected final void setHost(InternalInventoryHost host) {
...
+    protected ResourceHandler<ItemResource> createResourceHandler() {
+        return new AppEngInternalInventoryResourceHandler();
+    }
+
+    private class AppEngInternalInventoryResourceHandler
+            extends SnapshotJournal<AppEngInternalInventoryResourceHandler.Snapshot>
+            implements ResourceHandler<ItemResource>, IndexModifier<ItemResource> {
+        @Nullable
+        private Snapshot lastReleasedSnapshot;
+        private final BitSet changedSlots = new BitSet(size());
```
andrewsf: dead

src/main/java/appeng/util/inv/AppEngInternalInventory.java
```diff
@@ -41,6 +49,7 @@ public class AppEngInternalInventory extends BaseInternalInventory {
     private final int[] maxStack;
     private IAEItemFilter filter;
     private boolean notifyingChanges = false;
+    private boolean inTransactionalCode;
```
Mithi83, line 51: This is not initialized. If there is a code path that bypasses the `AppEngInternalInventoryResourceHandler` I don't know what would happen.

src/main/java/appeng/util/inv/AppEngInternalInventory.java
```diff
@@ -200,4 +210,179 @@ protected final void setHost(InternalInventoryHost host) {
...
+        }
+
+        @Override
+        public long getAmountAsLong(int index) {
+            return AppEngInternalInventory.this.getStackInSlot(index).getCount();
+        }
+
+        @Override
+        public long getCapacityAsLong(int index, ItemResource resource) {
+            return AppEngInternalInventory.this.getSlotLimit(index);
```
andrewsf, line 330: maybe should keep parity with `isValid` by returning 0 for invalid resources

src/main/java/appeng/util/inv/AppEngInternalInventory.java
```diff
@@ -105,7 +116,6 @@ public ItemStack extractItem(int slot, int amount, boolean simulate) {
         if (stack.getCount() <= toExtract) {
             if (!simulate) {
                 setItemDirect(slot, ItemStack.EMPTY);
-                notifyContentsChanged(slot);
```
Mithi83, line 108: Good catch of the redundant call here.

src/main/java/appeng/api/inventories/BaseInternalInventory.java
```diff
@@ -42,7 +42,5 @@ public final ResourceHandler<ItemResource> toResourceHandler() {
         return platformWrapper;
     }
 
-    protected ResourceHandler<ItemResource> createResourceHandler() {
-        return new InternalInventoryResourceHandler(this);
```
andrewsf, line 46: Note `InternalInventoryResourceHandler` class is dead code now

src/main/java/appeng/blockentity/misc/CondenserBlockEntity.java
```diff
@@ -231,6 +232,11 @@ public ItemStack insertItem(int slot, ItemStack stack, boolean simulate) {
         public ItemStack extractItem(int slot, int amount, boolean simulate) {
             return ItemStack.EMPTY;
         }
+
+        @Override
+        protected ResourceHandler<ItemResource> createResourceHandler() {
+            return new CondenseResourceHandler<>(ItemResource.EMPTY, 1, Integer.MAX_VALUE);
```
Mithi83, line 238: I know it was called that before, but shouldn't if be `CondenserResourceHandler` with an additional "r"?

src/main/java/appeng/util/inv/AppEngInternalInventory.java
```diff
@@ -200,4 +210,179 @@ protected final void setHost(InternalInventoryHost host) {
...
+
+        @Override
+        public int insert(ItemResource resource, int maxAmount, TransactionContext transaction) {
+            TransferPreconditions.checkNonEmptyNonNegative(resource, maxAmount);
+
+            var stack = resource.toStack(maxAmount);
+
+            updateSnapshots(transaction);
+
+            inTransactionalCode = true;
```
andrewsf, line 238: typically this try/finally pattern stores existing value & restores in the `finally` for thread safety (not really) and reentrancy

src/main/java/appeng/api/inventories/SubInventoryProxy.java
Mithi83: Maybe that should rather be `delegate.toResourceHandler()`? Just an educated guess motivated by looking at the other functions in this class.

src/main/java/appeng/blockentity/misc/CondenserBlockEntity.java
```diff
@@ -231,6 +232,11 @@ public ItemStack insertItem(int slot, ItemStack stack, boolean simulate) {
...
+        }
```
andrewsf, line 239: One thing to note is that the ItemHandler this replaces had some checks (`canAddOutput` method) to avoid consuming items if the condenser's output slot was full. This new handler will void inputs without producing anything.

Seems it already works that way for fluids

src/main/java/appeng/util/inv/AppEngInternalInventory.java
shartte: that might be an additional bug with the previous impl this is copied from then

src/main/java/appeng/util/inv/AppEngInternalInventory.java
andrewsf: [implicitly initialized to false](https://docs.oracle.com/javase/specs/jls/se25/html/jls-4.html#jls-4.12.5)

Mithi83: Ok, I was looking one line before, which does exactly that explicitly. Shouldn't matter then.

src/main/java/appeng/util/inv/FilteredInternalInventory.java
```diff
@@ -85,4 +90,71 @@ public boolean isItemValid(int slot, ItemStack stack) {
...
+            // implementation
+            TransferPreconditions.checkNonEmptyNonNegative(resource, amount);
+
+            int inserted = 0;
+            int size = size();
+            for (int index = 0; index < size; index++) {
+                inserted += insert(index, resource, amount - inserted, transaction);
+                if (inserted == amount)
+                    break;
+            }
```
andrewsf: I think this is going to change the way items fill inventories in a bad way. Previously was going through `InternalInventory` which had `addItemSlow` do two passes over slots to try to fill any existing non-full slots before going into empty slots, but this does 1 pass, so empty slots are more likely to get taken up

Well... I guess the impact is pretty low because of where `FilteredInternalInventory` is used today (inscriber inputs are most affected). Might become a more visible issue if new things use it

src/main/java/appeng/api/inventories/BaseInternalInventory.java
shartte: it was in the API package hence removal = breaking change

src/main/java/appeng/api/inventories/BaseInternalInventory.java
```diff
@@ -42,7 +42,5 @@ public final ResourceHandler<ItemResource> toResourceHandler() {
...
-    }
+    protected abstract ResourceHandler<ItemResource> createResourceHandler();
```
andrewsf: Moving the protected method to abstract is an API breaking change

shartte: Yes I am planning to revert that, actually and add a warning here.

src/main/java/appeng/blockentity/misc/CondenserBlockEntity.java
shartte: I might've thought it condenses the items that enter it :-D

src/main/java/appeng/blockentity/misc/CondenserBlockEntity.java
shartte: Yeah I suppose it does, I think I'll port the old "blocking" mechanism over to the generic handler.

src/main/java/appeng/util/inv/AppEngInternalInventory.java
shartte: Typically yes, but the subclasses of this are known (and internal). However, I'll actually change it as future proofing.

src/main/java/appeng/util/inv/FilteredInternalInventory.java
shartte: I can indeed just use ResourceHandlerUtil.insertStacking here

PR #8971, opened 2026-09-11
Fix JEI drag&drop highlights

Mithi83:


Mithi83:
Before:
image

After:
image

This happens rarely, but seems to be linked to the GUI top offset becoming negative. On my dev PC this is the default window size and GUI scale 3.

shartte (approved):

PR #8974, opened 2026-09-14
Allow only storage cells as input to ME IO Port

### LINKED ISSUE #8879: IO port any item is inputed with hopper (or more) in the storage cell stot (left side)
### Describe the bug

IO port any item is inputed with hopper (or more) in the storage cell stot (left side)

Image

### How to reproduce the bug

hopper + IO port

### Expected behavior

same as manual input (storage cell only)

### Additional details

_No response_

26.1

NeoForge

### Crash log

no crash

Mithi83:
Fixes #8879


Mithi83:
`IOPortMenu` uses a `RestrictedInputSlot`, so manually placing arbitrary items via GUI was not possible, but the block entity itself did not restrict the type of item. Thus a ME IO Port with a hopper on top would accept any item from the hopper. Heavily inspired by `CellValidInventoryFilter` from `DriveBlockEntity`.

PR #8986, opened 2026-09-21
(#8985) Use the command key for the craft-missing modifier on macOS

### LINKED ISSUE #8985: Ctrl+click to craft missing items never triggers on macOS in the JEI crafting terminal
### Describe the bug

On macOS the JEI recipe transfer button's "CTRL + click to craft unavailable
items" never fires. Holding Control does update the tooltip to "Will craft
unavailable items", so AE2 sees the key, but the click never reaches the
handler and no crafting job is requested.

Minecraft rewrites Control + left click into a right click on macOS.
`MouseHandler#simulateRightClick` does this whenever
`InputQuirks.SIMULATE_RIGHT_CLICK_WITH_LONG_LEFT_CLICK` is set, and
`InputQuirks` sets that field to `ON_OSX`:

```java
if (InputQuirks.SIMULATE_RIGHT_CLICK_WITH_LONG_LEFT_CLICK && info.button() == 0) {
    if (pressed && (info.modifiers() & 2) == 2) {   // 2 = GLFW_MOD_CONTROL
        this.fakeRightMouse++;
        return new MouseButtonInfo(1, info.modifiers());
    }
}
```

JEI wraps the transfer button in its `IconButton`, which gates on vanilla
`AbstractWidget#isValidClickButton`. That accepts button 0 only, so the
rewritten click is dropped before `UseCraftingRecipeTransfer#transferRecipe`
runs.

The tooltip still updates because that handler reads the modifier through
`Minecraft#hasControlDown()`, which polls GLFW keys 341 and 345 directly and is
unaffected by the mouse rewrite. That method has no macOS mapping, so Command
and Option cannot satisfy it either. Control is therefore the only key that sets
the flag, and Control is also the key whose click gets rewritten. No combination
both sets the flag and delivers a left click.

### How to reproduce the bug

1. Run the client on macOS, with AE2 and JEI installed.
2. Open a crafting terminal on a network that has a pattern for some item, while
   lacking the ingredients for it.
3. View that item's recipe in JEI and hover the transfer (+) button. The tooltip
   reads "CTRL + click to craft unavailable items".
4. Hold Control. The tooltip changes to "Will craft unavailable items", which
   confirms AE2 sees the key.
5. Control + click the button. Nothing happens and no crafting job starts.
6. Command + click and Option + click also do nothing, because the flag stays
   false and the plain transfer then fails on the missing ingredients.

The same steps work on Windows and Linux.

### Expected behavior

Clicking the transfer button with the craft-missing modifier held queues the
missing ingredients for autocrafting, as it does on Windows and Linux. On macOS
that modifier should be the command key, which Minecraft does not rewrite, and
the tooltip should name whichever key the platform actually listens for.

### Additional details

Reproduced on AE2 26.1.12-beta, which is also current `main` (b7cf5822d).

`UseCraftingRecipeTransfer` is the only handler in the repository that reads
this modifier, so only the JEI crafting terminal path is affected.
`EncodePatternTransferHandler` never reads it.

JEI already handles this distinction for its own keybinds:
`JeiKeyModifier.CONTROL_OR_COMMAND` checks
`InputQuirks.REPLACE_CTRL_KEY_WITH_CMD_KEY` and falls back to
`GLFW_KEY_LEFT_SUPER` / `GLFW_KEY_RIGHT_SUPER`.

I have a fix ready and will open a pull request against `main`.

26.1

NeoForge

### Crash log

Not a crash, so there is no log. The mechanism is traced in full above.

julian-perge:
Fixes #8985

On macOS the "CTRL + click to craft unavailable items" action on the JEI recipe
transfer button never fires. Holding Control updates the tooltip, so the key is
seen, but the click itself never reaches the handler.

Minecraft rewrites Control + left click into a right click on that platform.
`MouseHandler#simulateRightClick` does so whenever
`InputQuirks.SIMULATE_RIGHT_CLICK_WITH_LONG_LEFT_CLICK` is set, and `InputQuirks`
sets it to `ON_OSX`. JEI wraps the transfer button in its `IconButton`, which
gates on vanilla `AbstractWidget#isValidClickButton`, and that accepts button 0
only. The rewritten click is therefore dropped before
`UseCraftingRecipeTransfer#transferRecipe` runs.

The tooltip still updates because the handler reads the modifier through
`Minecraft#hasControlDown()`, which polls GLFW keys 341 and 345 and is
unaffected by the mouse rewrite. That method has no macOS mapping, so Command
and Option cannot satisfy it either. Control is therefore the only key that sets
the flag, and Control is also the key whose click gets rewritten, so no
combination both sets the flag and delivers a left click.

## Fix

`UseCraftingRecipeTransfer` now reads the modifier through
`isCraftMissingModifierDown()`, which checks `GLFW_KEY_LEFT_SUPER` /
`GLFW_KEY_RIGHT_SUPER` when `InputQuirks.REPLACE_CTRL_KEY_WITH_CMD_KEY` is set
and falls back to `Minecraft#hasControlDown()` everywhere else. Command is not
rewritten by `simulateRightClick`, so the click survives.

`TransferHelper#createCraftingTooltip` now names whichever key the platform
listens for, via a new `ItemModText.CMD_CLICK_TO_CRAFT`. Otherwise the hint on
macOS would name Control, the one key that cannot work there.

Both places branch on `InputQuirks.REPLACE_CTRL_KEY_WITH_CMD_KEY` rather than
re-deriving the platform test, so there is a single source of truth. This is the
same field JEI uses for `JeiKeyModifier.CONTROL_OR_COMMAND`, which makes AE2
consistent with JEI's own handling of this distinction.

Behavior on Windows and Linux is unchanged: `REPLACE_CTRL_KEY_WITH_CMD_KEY` is
false there, so both paths still use Control and the tooltip still reads "CTRL".

## Testing

I built this branch and installed it into a 265 mod pack on macOS 26.5.2,
NeoForge 26.1.2.109, JEI 29.37.0.100, on a network with a pattern provider set
up for an item whose ingredients were missing.

Before the change, on that setup: hovering the transfer button showed "CTRL +
click to craft unavailable items", and holding Control switched it to "Will
craft unavailable items". That difference is what showed the key was being read
while the click was not arriving. Control + click, Command + click and Option +
click all did nothing.

After the change, on that same setup: Command + click on the transfer button
queues the missing ingredients for autocrafting. The new hint string is present
in the built jar's `en_us.json`.

I ran `gradlew spotlessApply`, and `gradlew jar` succeeds. I have not checked
the Windows and Linux paths on hardware. They are unchanged by construction,
since `REPLACE_CTRL_KEY_WITH_CMD_KEY` is false there and the fallback is the
original `hasControlDown()` call.


src/main/java/appeng/core/localization/ItemModText.java
```diff
@@ -16,6 +16,7 @@ public enum ItemModText implements LocalizationEnum {
     NO_ITEMS("Found no compatible items"),
     WILL_CRAFT("Will craft unavailable items"),
     CTRL_CLICK_TO_CRAFT("CTRL + click to craft unavailable items"),
+    CMD_CLICK_TO_CRAFT("CMD + click to craft unavailable items"),
```
andrewsf: Apple style is to either write
* Command-click
* ⌘-click
* Click while holding Command   (obviously this is bad, but it's the most correct)

Never "Cmd" / "CMD", and no "+"

julian-perge: Thanks for the heads up. I copied what I saw in JEI, but also good to know there's an official Apple doc about the styling.

For anyone else curious, here's the Apple docs about it: https://support.apple.com/guide/applestyleguide/c-apsgb744e4a3/web#apd6dd5430cd3f64

